PR #1, opened 2021-10-29
Changed description into textarea, added placeholders

MaggieCabrera:
I added a few changes to make it clearer what the info needed looks like:
Screenshot 2021-10-29 at 14 56 09

Lowering the opacity of the placeholders and the width of the inputs would be nice, I'll have a look how's the best way to do that.


scruffian (approved):
LGTM

PR #3, opened 2021-11-05
Generate slug from theme name

MaggieCabrera:
This removes the slug input and generates it directly from the theme name. This is what the underscores generator did too and I did this because it reduces the amount of input needed from the user and ensures that theme name and theme slug are consistent with each other.


scruffian (approved):
LGTM

PR #4, opened 2021-11-05
Only get templates from the child theme not from Blockbase

scruffian:
Previously, when creating a child Blockbase theme, all of the templates from Blockbase would be exported. Since the child theme inherits these already we only need to export any changes.

To test this, you will need to use the empty blockbase child theme in this PR: https://github.com/Automattic/themes/pull/4999

To test
- Switch to the Empty Blockbase Child
- Make changes to the templates
- Export a new theme using the Create Blockbase Child form
- Your theme should only contain the modified templates.

### DIFF
--- a/index.php
+++ b/index.php
@@ -138,6 +138,11 @@ GNU General Public License for more details.
 ";
 }
 
+// Switches the template directory to the stylesheet directory.
+function create_blockbase_theme_reset_template_directory() {
+	return get_stylesheet_directory();
+}
+
 /**
  * Creates an export of the current templates and
  * template parts from the site editor at the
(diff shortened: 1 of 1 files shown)


MaggieCabrera:
I think this is missing some code? I'm getting errors saying that the `create_blockbase_theme_reset_template_directory` callback doesn't exist

MaggieCabrera:
I think I understand what you want but this means that the user needs to have an extra theme installed for this to work that in turn we also need to maintain. I wonder if there's a way to do the same without needing an extra theme.

pbking:
Yea, looks like you're missing the meat in this sandwich.

Betcha this fixes the shortcomings mentioned in https://github.com/Automattic/create-blockbase-theme/pull/10

scruffian:
Oooh sorry about that. It's there now.

scruffian:
Solved by https://github.com/Automattic/create-blockbase-theme/pull/10 instead

PR #5, opened 2021-11-09
added meta info for the plugin and translated strings

MaggieCabrera:
This adds extra meta info to the plugin and a text domain. I went ahead and prepared strings for translation too.

Screenshot 2021-11-09 at 11 12 49


mikachan (approved):
This looks great to me!

PR #6, opened 2021-11-09
add all current theme's css to the generated one

MaggieCabrera:
This PR gets the parent theme's CSS dynamically rather than assuming that it lives under `assets/theme.css` and adds any other .css files to it that are present on the theme we are building from.

To test this, try building a child of Quadrat. The generated CSS should contain both Blockbase's and Quadrat's CSS. I've added some comments so that the user can discern where everything is coming from.

Test this with Blockbase active and check that the theme.css file is empty (Blockbase's CSS is already being inherited)

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ A WordPress plugin to create Blockbase child themes
 Find out more about Blockbase at [blockbasetheme.com](https://blockbasetheme.com)
 
 ### Step 1 – Setup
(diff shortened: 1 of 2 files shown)


mikachan (approved):
This is really cool, and seems to work well for me! I tried with Quadrat and Skatepark, and both times I saw the child themes CSS included. I also tried with Blockbase and I didn't see any extra CSS added.

Here's an example from my Skatepark child test:

image

Should we update the readme to mention that you can run this plugin with either Blockbase or any of its children active?

MaggieCabrera:
@mikachan just updated the README, do you think that'll do?

mikachan:
Yep, that looks good to me!

PR #7, opened 2021-11-10
Cleanup and pre-commit hook

MaggieCabrera:
I moved the plugin to its own folder and added the same pre-commit hook we are using in the themes repo.

To test this run `npm install` and `composer install` (note that I had issues with v2 of composer and I had to downgrade to v1.10). Make a change to the PHP file like removing spaces between parentheses and try to commit the change. The pre commit hook should format it correctly.


pbking:
I had issues installing (maybe a PHP version mismatch as I have 8 currently installed...).  I could probably work around that, but this feels like overkill for the plugin.  We could probably remove (or simplify) most of this relating to .js or .scss/.css and trigger the pre-commit from bash rather than node eliminating the need for a package.json and those dependencies.

PR #9, opened 2021-11-12
Add classes to form elements

mikachan:
This PR just adds a couple of classes to some of the form elements to slightly improve the styling, mainly making the inputs the same width and making the button a primary button.

image


MaggieCabrera (approved):
Looking good!

PR #10, opened 2021-11-12
Filter templates and JSON from result when using Blockbase as base theme

pbking:
When the base theme is BLOCKBASE:

only export theme.json settings modified by the user
do not export any block templates/parts except those modified by the user

NOTE: When exporting with a CHILD theme as the base then UNMODIFIED templates/parts that originated in Blockbase are still exported.  This is a bug.

### DIFF
--- a/index.php
+++ b/index.php
@@ -18,41 +18,38 @@
  * @package gutenberg
  */
 
-/*
- if theme is blockbase don't export any templates/parts
- if theme is a blockbase child export all templates/parts
- no matter the theme, always export MODIFIED templates/parts
-*/
 
-/*
- if theme is blockbase only export theme.json settings that have been modified.
- if theme is blockbase child export all theme.json settings as well as any settings that have been modified.
-*/
+// I feel like there should be a function to do this in Gutenberg but I couldn't find it
+function remove_theme_key( $data ) {
+	if ( is_array( $data ) ) {
+		if ( array_key_exists( 'theme', $data ) ) {
+			if ( array_key_exists( 'user', $data ) ) {
+				return $data['user'];
+			}
+
+			return $data['theme'];
+		}
+		foreach( $data as $node_name => $node_value  ) {
+			$data[ $node_name ] = remove_theme_key( $node_value );
+		}
+	}
+
+	return $data;
+}
 
 function gutenberg_edit_site_get_theme_json_for_export() {
+
+	$base_theme = wp_get_theme()->get('TextDomain');
 	$child_theme_json = json_decode( file_get_contents( get_stylesheet_directory() . '/theme.json' ), true );
 	$child_theme_json_class_instance = new WP_Theme_JSON_Gutenberg( $child_theme_json );
 	$user_theme_json = WP_Theme_JSON_Resolver_Gutenberg::get_user_data();
+
+	if ( $base_theme === 'blockbase' ) {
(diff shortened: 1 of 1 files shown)


MaggieCabrera:
Ok, now I know why Ben used that function `remove_theme_key`, this is what a modified blockbase looks like:
Screenshot 2021-11-12 at 17 01 18

That "user" key needs to go, otherwise the json file won't work

MaggieCabrera:
This is an alternative to https://github.com/Automattic/create-blockbase-theme/pull/4

index.php
```diff
@@ -183,8 +191,15 @@ function gutenberg_edit_site_export_theme_create_zip( $filename, $theme ) {
 	// Load templates into the zip file.
 	$templates = gutenberg_get_block_templates();
 	foreach ( $templates as $template ) {
-		$template->content = _remove_theme_attribute_from_content( $template->content );
 
+		//Currently, when building against CHILD themes of Blockbase, block templates provided by Blockbase, not modified by the child theme or the user are included in the page. This is a bug.
+		//if the theme is blockbase and the source is "theme" we don't want it
+		if ($template->source === 'theme' && $template->theme === 'blockbase') {
```
MaggieCabrera: this will fail if the theme is in a subdirectory. I'm getting this when I dump `$template->theme`:

` public 'theme' => string 'a8c-themes/blockbase'`

pbking:
> Ok, now I know why Ben used that function remove_theme_key, this is what a modified blockbase looks like:
Good catch, thanks.

I renamed that function to `flatten_theme_json` now that what it does is a bit more clear, called it when just returning user JSON and made sure it worked in a few scenarios.

index.php
pbking: Another good catch.  Refactored the check to `... && strpos($template_part->theme, 'blockbase') !== false` to account for that.

pbking:
Thanks for checking out this work @MaggieCabrera .  I believe it's ready for another review.

MaggieCabrera (approved):
it works now, thank you!!

scruffian:
This does deal with the issue of duplicating templates, but it doesn't really help with the other aspect of duplicated code, which is the theme.json part, so I think the solution in #4 coupled with https://github.com/Automattic/themes/pull/4999 is preferable.

MaggieCabrera:
> This does deal with the issue of duplicating templates, but it doesn't really help with the other aspect of duplicated code, which is the theme.json part, so I think the solution in #4 coupled with [Automattic/themes#4999](https://github.com/Automattic/themes/pull/4999) is preferable.

Mmh, which duplicated code? This PR only includes, in the case of a direct BB child, the changes in theme.json made by the user, and lets the rest be inherited by the parent.

scruffian:
The problem is that if you're on Blockbase itself it copies everything that's in theme.json - see https://github.com/Automattic/themes/pull/5030 for an example...

MaggieCabrera:
> The problem is that if you're on Blockbase itself it copies everything that's in theme.json - see [Automattic/themes#5030](https://github.com/Automattic/themes/pull/5030) for an example...

If you are seeing that then it's a bug. This is a child based from BB that only has the header template changed and colors changed via customizer using this PR:

Screenshot 2021-11-15 at 12 31 30

scruffian:
Oh yes, you're right! I think this looks good to go!

PR #13, opened 2021-11-17
Add a screenshot and remove the functions.php file

scruffian:
This adds a blank screenshot to the child theme and removes the functions.php file.


pbking (approved):
How cool is that?
Loving how super-small these child themes are able to get!

PR #19, opened 2021-11-19
Rename function to align with Gutenberg naming

scruffian:
This function was renamed in Gutenberg, so we need to rename it here too. This shouldn't be merged until the change is in the latest version of Gutenberg.


MaggieCabrera:
I'm bringing this in, since it already happens with the latest release version

PR #22, opened 2021-11-25
Fix gutenberg function fatal

MaggieCabrera:
After https://github.com/WordPress/gutenberg/pull/36559 the function `_remove_theme_attribute_from_content` has been renamed to `_remove_theme_attribute_in_block_template_content` and I think it's going to stay that way when backported to core. This PR resolves fatals related to this rename happening on GB 12.0.0


MaggieCabrera:
Ugh, it's the same as #19 XD

PR #23, opened 2021-11-25
Remove the code to replace - with _

scruffian:
When we were using the theme name in function calls we needed it to have _s not -s. Now we don't do that we can use - according to themes convention.


mikachan (approved):
LGTM!

PR #24, opened 2021-11-26
Determine whether functions exist before we call them

scruffian:
Since we're not sure which version of GB you might be on, lets try both!


MaggieCabrera (approved):
Works for me both on trunk and 12.0.0

PR #25, opened 2021-12-01
Only allow child themes of Blockbase

scruffian:
This prevents people using the plugin to create children of other themes.

### DIFF
--- a/index.php
+++ b/index.php
@@ -342,11 +342,22 @@ function blockbase_save_theme() {
 			return add_action( 'admin_notices', 'create_blockbase_child_admin_notice_error' );
 		}
 
+		if ( wp_get_theme()->get( 'Name' ) !== 'Blockbase' ) {
+			return add_action( 'admin_notices', 'create_blockbase_child_admin_notice_error_wrong_theme' );
+		}
+
 		add_action( 'admin_notices', 'create_blockbase_child_admin_notice_success' );
 		gutenberg_edit_site_export_theme( $_GET['theme'] );
 	}
 }
 add_action( 'admin_init', 'blockbase_save_theme');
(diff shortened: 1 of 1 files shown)


pbking:
There's a fair bit of "If it's Blockbase do this, if it's a child do that" code that we could remove to accompany this change.

https://github.com/Automattic/create-blockbase-theme/blob/trunk/index.php#L57
https://github.com/Automattic/create-blockbase-theme/blob/trunk/index.php#L101
https://github.com/Automattic/create-blockbase-theme/blob/trunk/index.php#L189
https://github.com/Automattic/create-blockbase-theme/blob/trunk/index.php#L211

pbking:
Somewhat relatedly I think it might be good to use a child theme... but as a means to update or "re-export" that child theme.

Meaning...

* Use Blockbase to build "Turnup"
* Activate "Turnup"
* Make further changes
* Export a new version of "Turnup"
(NOT a new theme called "Carrot")
which may incorporate changes made in blockbase since Turnup was first exported and may include changes and improvements made by the user in the FSE, etc.

scruffian:
https://github.com/Automattic/create-blockbase-theme/blob/trunk/index.php#L57
https://github.com/Automattic/create-blockbase-theme/blob/trunk/index.php#L101

These two I have removed, as we no longer need them.

https://github.com/Automattic/create-blockbase-theme/blob/trunk/index.php#L189
https://github.com/Automattic/create-blockbase-theme/blob/trunk/index.php#L211

These two I think we still need. The aim is to detect whether the template is part of Blockbase rather than a modified template, which is still relevant I think.

pbking (approved):

PR #26, opened 2021-12-10
Add support for array keys called custom as well as user

scruffian:
This adds support for the custom array key, which is the new name in Gutenberg 12.1.


pbking (approved):
👍 🙏

PR #28, opened 2021-12-15
Removes an errant var_dump call

ChaosExAnima:
Ran into issues with this plugin and realized that there's a `var_dump` in the code that was outputting before the headers were sent. This PR removes that line.


scruffian (approved):
Thanks

PR #31, opened 2022-01-12
First pass at theme generation

scruffian:
This is just a proof of concept aimed at Videomaker, there's is obviously a lot more work to do if we want to pursue the idea.

The premise is that every theme can be represented by a JSON object which defines everything that goes into the theme. The themes.json file contains a list of all our themes and some definitions for each of them. Ideally the different parts of the theme would have more descriptive names than simply `videomaker`, so that they were more reusable. Only Videomaker has had much input so far.

Ideally we'd note dependencies between different aspects of the theme - so assets used in patterns would need to be connected, for example.

If you run `npm run build:children`, the node script will run and generate a copy of videomaker in your themes directory (assuming that this repo is checked out at the same level as themes).  If you do a git diff inside your themes repo you'll see that most parts of the theme are unchanged.

For videomaker most of the work is done the outstanding pieces are:
- the php file to include the patterns
- readme.txt
- screenshot.png
- functions.php

These would all be possible to script as well, but I thought I'd share this work before it goes further to get early feedback.

Some initial thoughts:
There is a lot of promise here - the script turns many of the decisions we have made into repeatable decisions that can be made for future themes. This will change theme maintenance from a simple but very laborious job to a more complex but also much faster one. It should mean out themes are more consistent and don't have the kind of weird idiosyncrasies that arise from copy/pasted code.

I do have concerns though. This is redefining the way we use theme.json works, and creating an abstraction that will make our work harder to understand


pbking:
I think we've moved passed this direction.  Closing this PR.

PR #32, opened 2022-02-17
Generate themes from any block theme

MaggieCabrera:
This PR refactors the plugin so that it can be used with any block theme, not just Blockbase. These are the cases that it covers:

- If the active theme is a classic theme: it will show an error.
- If the active theme is a block theme child of another theme: the generated theme will be a child theme of the parent of the active theme, copying the template files and settings of the active theme plus the user changes on templates and Global Styles. I've tested this scenario using Geologist.
- If the active theme is a standalone block theme: the form will show two options for the new theme. I've tested this one with Remote and Blockbase.
  - We can generate a child theme of the active theme. The new theme will only contain the user changes in templates and Global Styles, the rest is inherited from the parent.
  - We can generate a standalone based on the active theme. The new theme will be a copy of the active theme plus the customizations made by the user.

### DIFF
--- a/index.php
+++ b/index.php
@@ -145,8 +145,8 @@ function gutenberg_edit_site_export_theme_create_zip( $filename, $theme ) {
 	$zip = new ZipArchive();
 	$zip->open( $filename, ZipArchive::OVERWRITE );
 	$zip->addEmptyDir( $theme['slug'] );
-	$zip->addEmptyDir( $theme['slug'] . '/block-templates' );
-	$zip->addEmptyDir( $theme['slug'] . '/block-template-parts' );
+	$zip->addEmptyDir( $theme['slug'] . '/templates' );
+	$zip->addEmptyDir( $theme['slug'] . '/parts' );
 
 	// Load templates into the zip file.
 	$templates = gutenberg_get_block_templates();
@@ -166,7 +166,7 @@ function gutenberg_edit_site_export_theme_create_zip( $filename, $theme ) {
 			$template->content = _remove_theme_attribute_from_content( $template->content );
 		}
 		$zip->addFromString(
-			$theme['slug'] . '/block-templates/' . $template->slug . '.html',
+			$theme['slug'] . '/templates/' . $template->slug . '.html',
 			$template->content
 		);
 	}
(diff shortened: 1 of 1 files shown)


pbking:
@MaggieCabrera  I made some good progress on what we were pairing on today.  

The utility functions are patched into WP_Theme_JSON_Resolver_Gutenberg from the plugin.  I think once we get that settled we can consider getting that into Gutenberg proper but like you mentioned in the meantime we’ve got a solution that does what we need now.

The three scenarios seem to be working OK (at least for the JSON being exported): standalone (export it all), child (export user only) and grandchild (export current theme and user but not parent theme).

Color palettes (either changed via Customizer or Global Styles panel) is working too.  This is due to some filtering in the utility that is filtering out duplicate slugs once "custom" and "theme" have been merged.  Blockbase's color values still need to be transitioned to "theme" I think, but in the meantime how they are stored today isn't getting in the way of this plugin working as we intend.

MaggieCabrera:
@alaczek when you have a moment please test this PR to see if it fits your workflow, it should let you build Blockbase children the same way you've done so far and also let you create standalone block themes.

index.php
```diff
@@ -70,23 +141,24 @@ function blockbase_get_style_css( $theme ) {
...
+	$template = !create_blockbase_get_new_parent( $theme ) ? "" : "Template: ". create_blockbase_get_new_parent( $theme ) ."\n";
+
+	return "/*\n" .
+	"Theme Name: {$name}\n" .
+	"Theme URI: {$uri}\n" .
+	"Author: {$author}\n" .
+	"Author URI: {$author_uri}\n" .
+	"Description: {$description}\n" .
+	"Requires at least: 5.8\n" .
+	"Tested up to: 5.8\n" .
```
scruffian:
```suggestion
	"Tested up to: 5.9\n" .
```

pbking:
It looks like this PR is getting rather messy...
I think I lost some work to a force-push and there are merge instructions in here now.  Not sure when exactly...

I think I will branch my work into another branch to avoid toe-stepping.  I don't think we should merge this one anymore, there's too much noise and too many changes.

scruffian:
Sorry, that was my fault. It didn't help that Slack was down so I couldn't tell if anyone else was working on it.

pbking:
> Sorry, that was my fault.

Not your fault man.  I also rebased and was tinkering with things without a heads up.  I think we're just trying to do too many things in this PR.  Breaking the bits into smaller chunks will help.

MaggieCabrera:
I'm closing this in favor of the other PRs you all have been working on

PR #34, opened 2022-02-22
Remove confusing references to blockbase

scruffian:
This removes some of the references to Blockbase

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,18 +1,18 @@
-# Create Blockbase Theme
-A WordPress plugin to create Blockbase child themes
+# Create Block Theme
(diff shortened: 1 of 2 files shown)


MaggieCabrera:
some of this was done by https://github.com/Automattic/create-block-theme/pull/32 but I didn't go to far cause I didn't know if we could rename the plugin

pbking:
I think we can ultimately take the generic-ness further as it should be able to make child themes of any Block theme.  But I think this is a good start toward that.

pbking (approved):

PR #35, opened 2022-02-22
Augment Gutenberg to export theme.json according to given parameters

pbking:
Change how the theme.json is constructed by augmenting Gutenberg with how we THINK we MIGHT want that to work.  Allows for the export of ALL settings, USER + THEME settings or just USER settings.

To test with a Blockbase CHILD theme activated (such as geologist) create a theme with the utility.  Notice that all of the JSON from Geologist is in there, as well as any user changes made with FSE/Customizer, but none of the JSON from Blockbase.  (This is building a 'grandchild' theme)

Now with BLOCKBASE activated create a theme.  Notice the theme.json is ONLY the values customized by the user.

### DIFF
--- a/gutenberg_additions.php
+++ b/gutenberg_additions.php
@@ -1,0 +1,104 @@
+<?php
+
+function augment_gutenberg_with_utilities() {
+
+	//Ultimately it is desireable for Gutenberg to have this functionality natively.
+	// In the meantime we are patching the functionality we are expecting into the Theme JSON Resolver here
+
+	class MY_Theme_JSON_Resolver extends WP_Theme_JSON_Resolver_Gutenberg {
+
+		/**
+		 * 'Flatten' theme data that expresses both theme and user data.
+		 * change property.[custom|theme].value to property.value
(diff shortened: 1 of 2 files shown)


gutenberg_additions.php
```diff
@@ -0,0 +1,104 @@
...
+	        		$theme_json_data = static::translate( $theme_json_data, wp_get_theme()->get( 'TextDomain' ) );
+				$theme_theme     = new WP_Theme_JSON_Gutenberg( $theme_json_data );
+ 				$theme->merge( $theme_theme );
+			}
+
+	        	$theme->merge( static::get_user_data() );
+
+			$data = MY_Theme_JSON_Resolver::flatten_theme_json($theme->get_raw_data(), null);
+
+			// die(json_encode($data));
```
scruffian: Don't forget to remove this :)

pbking: Thanks. :)

scruffian (approved):
LGTM

PR #36, opened 2022-02-22
Add function to gather CSS to be exported for themes

pbking:
This change will export CSS for the new theme.

When building a BLOCKBASE CHILD theme no CSS should be exported.
When building a BLOCKBASE GRANDCHILD (for instance using Geologist as the source theme) then the SOURCE THEME's CSS should be exported.

Additionally this will export any CSS in the styles.css file into the child's /assets/theme.css class.  

Evaluated with Blockbase and Geologist.

NOTE: This code will also export STANDALONE themes (from Blockbase) which would include blockbase's ponyfill.css.  However the plugin doesn't yet export standalone themes so this portion isn't testable yet.

### DIFF
--- a/index.php
+++ b/index.php
@@ -109,25 +109,32 @@ GNU General Public License for more details.
  * When building a CHILD theme no extra CSS is included.
  */
 function create_block_theme_get_theme_css( $theme ) {
-	// NOTE: Themes that keep their CSS in a structure other than Blockbase's will need something different...
+
+	// if we are building a CHILD theme we don't need any CSS
+	if ( $theme['type'] === 'child' ) {
+		return '';
+	}
+
+	$css_string = '';
+
+	// For those themes that keep styles in styles.css include anything that isn't metadata.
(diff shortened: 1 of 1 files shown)


MaggieCabrera:
I remember making a script that gets all the CSS from the theme, I thought it was built into the plugin, I'll look for it

MaggieCabrera:
This [PR](https://github.com/Automattic/create-block-theme/pull/6) gets the css dynamically regardless of where it is stored, I don't know what happened to the code

EDIT: ok, we removed that here https://github.com/Automattic/create-block-theme/pull/25

scruffian:
I agree with Maggie, we should try to get the CSS that is enqueued rather than just assuming where it is.

scruffian:
I added a commit to reinstate that code.

pbking:
The only issue I see at the moment is that it includes the metadata from every theme's style.css file and I think it would be cleaner to exclude the metadata portion.

PR #37, opened 2022-02-23
Update the name of the folders templates and template parts are exported to.

pbking:
This change updates the folders that templates and template parts are exported to.  This is the location expected in the current version of Gutenberg.

To test create a "Grandchild" theme (using a Blockbase Child as a source) such as Geologist and note the folders the templates are stored in.  Load the created theme to ensure that the templates in those folders are loaded as expected.


scruffian (approved):
LGTM

PR #39, opened 2022-02-23
Add/standalone themes

pbking:
This work borrows very heavily from @MaggieCabrera 's work in #32.

This works exporting any parent theme as a standalone theme.  

To test attempt to export ALL THREE kinds of themes:

* GRANDCHILD : Activate a Blockbase CHILD theme (such as Geologist).  After making changes export the theme.  It should be a BLOCKBASE child theme with all of the theme.json values from the activated theme as well as any changes you have made.  The child theme's /assets/theme.css resource should be included. Templates and template parts from the active theme should also be included.
* CHILD : Activate Blockbase.  After making changes export the theme (selecting 'Child them of the current active theme' from the Create Block Theme page).  The exported theme.json should ONLY include changes made by the user.  The /assets/theme.css file should be empty and only changes made to templates and template-parts should be included.  Repeat the process with a non-Blockbase Block theme (such as TwentyTwentyTwo).
* STANDALONE : Activate Blockbase.  After making changes export the theme (selecting 'Standalone theme' from the Create Block Theme page).  The exported theme.json should include ALL of the settings from Blockbase as well as any changes you made.  The /assets/theme.css file should include the contents of Blockbase's /assets/ponyfill.css file.  That /assets/theme.css file should be automatically loaded (in both the view and editor).  Likewise any fonts selected (as defined in the theme.json) should be loaded (accomplished by the bare-bones functions.php file included in the generated theme.  Repeat the process with a non-Blockbase Block theme (such as TwentyTwentyTwo)

Note: This does NOT YET export any of the theme's PATTERNS.  As such any of the templates that leverage those patterns will not work.

### DIFF
--- a/index.php
+++ b/index.php
@@ -390,8 +390,8 @@ function create_blockbase_theme_page() {
 			<p><?php wp_kses_post( _e( 'The new theme will be based on whatever theme you have active at the moment. <br />If you want a fresh start, try using <a href="https://github.com/WordPress/theme-experiments/tree/master/emptytheme">Empty Theme</a> as a base.', 'create-block-theme' ) ); ?></p>
 			<p><?php _e('The current active theme is:', 'create-block-theme'); ?> <?php echo wp_get_theme()->get('Name'); ?></p>
 			<form method="get">
-				<label><?php _e('Theme name', 'create-block-theme'); ?><br /><input placeholder="<?php _e('Blockbase', 'create-block-theme'); ?>" type="text" name="theme[name]" class="regular-text" required /></label><br /><br />
-				<label><?php _e('Theme description', 'create-block-theme'); ?><br /><textarea placeholder="<?php _e('Blockbase is a simple theme that supports full-site editing. Use it to build something beautiful.', 'create-block-theme'); ?>" rows="4" cols="50" name="theme[description]" class="regular-text"></textarea></label><br /><br />
+				<label><?php _e('Theme Name', 'create-block-theme'); ?><br /><input required placeholder="<?php echo wp_get_theme()->get('Name'); ?>" type="text" name="theme[name]" class="regular-text" required /></label><br /><br />
+				<label><?php _e('Theme Description', 'create-block-theme'); ?><br /><textarea placeholder="<?php echo wp_get_theme()->get('Description'); ?>" rows="4" cols="50" name="theme[description]" class="regular-text"></textarea></label><br /><br />
 				<?php if ( ! is_child_theme() ): ?>
 				<label><input checked value="child" type="radio" name="theme[type]" class="regular-text code" /><?php _e('Child theme of the current active theme', 'create-block-theme'); ?></label><br /><br />
 				<label><input value="block" type="radio" name="theme[type]" class="regular-text code" /><?php _e('Standalone theme', 'create-block-theme'); ?></label><br /><br />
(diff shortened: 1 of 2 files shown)


jffng:
✅  blockbase grandchild
✅  blockbase child
✅  tt2 child
✅  blockbase standalone
✅  tt2 standalone (the heading font didn't load, but I think that's expected at this stage in the plugin's development?)

scruffian:
I added a commit which exports the whole theme if you select standalone and then makes any necessary modifications.

pbking:
I really like the direction, this seems like the right move.

One aspect to consider is all of the "namespaced" things that are in the exported "standalone" theme.  Rather than considering this a NEW theme perhaps we should consider this an option an "export" (of a type that we eventually expect the FSE to do).  

In the scenario where a user selects the (what we call "standalone") I suggest we disable the form, the exported theme would have the metadata of the current theme and could be a possible drop-in replacement (with updated templates/theme.json settings).  

Perhaps the only change to the metadata might be a version change?

Should a user decide to refactor said exported theme as a "new" theme that seems work better suited to a person (or a separate tool).

index.php
```diff
@@ -151,6 +150,114 @@ function create_block_theme_get_theme_css( $theme ) {
 	return $css_string;
 }
 
+/**
+ * Standalone themes need a little extra logic to:
+ *  - Load the theme.css
+ *  - Load fonts from theme.json
+ * This is that logic.
+ */
+function create_block_theme_get_functions( $theme ) {
```
pbking: This won't be necessary anymore.

index.php
```diff
@@ -159,27 +266,52 @@ function create_block_theme_get_theme_css( $theme ) {
...
+				$zip->addFile( $file_path, $relative_path );
+			}
+		}
+	}
 
 	// Load templates into the zip file.
 	$templates = gutenberg_get_block_templates();
+	if ( $templates ) {
+		$zip->addEmptyDir( 'templates' );
+	}
```
pbking, line 255: Exporting a "standalone" theme where the current them uses the out-of-date templates directory (such as blockbase) results in the exported theme having templates in BOTH locations.  

I think better would be to have this exported location exclusively (no matter where the original has them).  It's best to export them (to pick up the user changes).

Alternately we could just choose to export them where the current theme has them declared.

MaggieCabrera:
With remote active and standalone selected, I'm getting the block patterns and font files a bit mangled:
Screenshot 2022-02-25 at 14 52 43

index.php
pbking: Removed.

index.php
pbking: I've added code to exclude the templates during "copy all the things".  I think this addresses things here.

pbking:
> With remote active and standalone selected, I'm getting the block patterns and font files a bit mangled...

Interesting.  Remote exported as expected for me. 
#WorksOnMyMachine :P

MaggieCabrera:
> > With remote active and standalone selected, I'm getting the block patterns and font files a bit mangled...
> 
> Interesting. Remote exported as expected for me. #WorksOnMyMachine :P

I did have some template changes and global styles tweaks, but maybe it's my environment!

pbking:
The idea of "exporting" the theme (into a useable state) allowing the modification of themes based on changes in the FSE seems like a great boon.  I'm excited to see that shake out if we agree on that move.

In the same vein, while I also see great value in CREATING a "grandchild" theme, based on a child theme I also see value in EXPORTING said child theme in the same way that a parent theme could be exported.  

For example the same tool could be used to both CREATE a "Rockhunter" GRANDCHILD (from Geologist) but also allow minor updates to Geologist to happen exclusively in the FSE.

Here's how I see that form possibly looking:

For a non-child theme:
image
and
image

And for a child theme:
image
and
image

pbking:
So after a chat @MaggieCabrera and I agreed that when exporting PARENT themes it will be important to:
* Export the theme (with the existing metadata) allowing a user to update a theme using the FSE resulting in a .zip that represents the theme with those changes
* Create a STANDALONE theme based based on the current theme with the current changes, and with a well documented and potentially limited amount of resources changed to support that. (i.e. changing the namespace of functions, localizations, etc) The work to change the namespacing can't possibly handle all of the scenarios but we can make it an option documenting those places where that work will happen.
* Create a CHILD theme of the activated parent theme with the changes made via the FSE.
Additionally, relating to activated CHILD themes it’s also important to be able to:
* Export the currently activated CHILD theme with local changes allowing a user to update the theme using the FSE resulting in a .zip that represents the theme with those changes
* Create a NEW child theme, based on the currently activated child theme.

(@Maggie please LMK if I misstated common understanding)

Most of the remaining work then is to ensure that what is exported for the Standalone scenario above has had the namespacing changed and to allow the "export" options for both parent and child activated themes.

MaggieCabrera:
> (@Maggie please LMK if I misstated common understanding)
> 
> Most of the remaining work then is to ensure that what is exported for the Standalone scenario above has had the namespacing changed and to allow the "export" options for both parent and child activated themes.


Yeah, that sounds good to me, let me know if you want help with this. You can check how we did the namespace stuff for the [empty theme generator](https://github.com/WordPress/theme-experiments/blob/master/new-empty-theme.php). Just be aware that the [function to sanitize exists already in core](https://developer.wordpress.org/reference/functions/sanitize_title_with_dashes/), I needed a copy cause the script was to be run on a terminal, without loading any WP code.

pbking:
Closing in deference to #41

PR #41, opened 2022-03-01
Refactor w/ boilerplate & encapsulate logic

pbking:
This PR continues the work in #39 and refactors that work in a number of ways:

The plugin is refactored to use the structure from [here](https://github.com/DevinVinson/WordPress-Plugin-Boilerplate).  (I may not have done that correctly, but I hope it's a good start...)

Logic has been broken into smaller functions where possible.

* EXPORTING a PARENT theme: Activate a parent theme (Blockbase and TTT are two examples I've experimented with).  Make changes using FSE to templates and global styles.  Select the "Export" option.  Note that the zip file contains everything in the original theme but theme.json, templates and parts have user changes integrated.
* CREATING CHILD: Activate a parent theme (Blockbase/TTT) and make changes.  Select the "Create child..." option.  Note that the zip file contains ONLY the theme.json file with the changes you made as well as templates/parts folders containing modifications you made
* CLONING a PARENT: (This used to be called the "Standalone" option). Activate a parent theme.  Make changes.  Select the "Clone" option.  Note that the zip contains everything available from an EXPORT option but with new metadata in style.css  **NOTE: The resulting theme will NOT have namespaced values changed.  This is a limitation noted in the instructions.**  
* EXPORTING a CHILD theme: Activate a child theme (Geologist is an example I experimented with).  make changes and select the "Export" option.  You will see the theme exported with the changes to theme.json and templates integrated.  
* CREATING A SIBLING theme: This one is still a work-in-progress.  GOAL: Activate a child theme and make changes.  Choose the "Create Sibling..." option.  Note that the zip contains everything that the currently activated theme provided as well as the user changes, but with a different style.css metadata creating a new child theme with the same parent as the currently activated theme, with all of the customizations from the currently activated theme as well as from the user. **NOTE: The resulting theme will NOT have namespaced values changed.  This is a limitation noted in the instructions.**


admin/class-create-block-theme-admin.php
```diff
@@ -0,0 +1,428 @@
...
+
+	/**
+	 * Create a sibling theme of the activated theme
+	 */
+	function create_sibling_theme( $theme ) {
+		// Sanitize inputs.
+		$theme['name'] = sanitize_text_field( $theme['name'] );
+		$theme['description'] = sanitize_text_field( $theme['description'] );
+		$theme['uri'] = sanitize_text_field( $theme['uri'] );
+		$theme['author'] = sanitize_text_field( $theme['author'] );
```
MaggieCabrera, line 63: I wonder if $theme shouldn't be a property of the class

admin/class-create-block-theme-admin.php
```diff
@@ -0,0 +1,428 @@
...
+		header( 'Content-Length: ' . filesize( $filename ) );
+		flush();
+		echo readfile( $filename );
+		die();
+	}
+
+	/**
+	 * Create a child theme of the activated theme
+	 */
+	function create_child_theme( $theme ) {
```
MaggieCabrera, line 169: this function is very similar to `create_sibling_theme`, maybe there's a way we can share the common parts of the code?

pbking: create_sibling_theme isn't quite done yet, and I suspect as it resolves it's going to take on aspects of both create_child_theme and export_theme.  I think there's common logic we can extract from all four "theme creation" functions though so maybe we can "piggyback" that from one into another somehow.

admin/class-create-block-theme-admin.php
pbking: Generally I prefer functions that accept all the parameters it needs and returns the thing it does work on.  I'm not sure I see a gain doing that other than eliminating the need to define $theme as a function parameter.  However building up that theme object could likely happen elsewhere to enforce DRY (as you mention below).

admin/class-create-block-theme-admin.php
pbking: Definitely have some optimizations we can make to this to make it more DRY.  Going to resolve this for now as we can make those changes later (and since it's all working as expected at the moment. :) )

pbking:
I believe this is ready for review.  The problem with namespacing is the only major issue yet to be resolved (I think).  If these changes check out then I hope to tackle that in a later change. 

There's also some DRY refactoring that we could to to simplify; there's a fair bit of copy-pasta.  Another thing to tackle in a subsequent change.

MaggieCabrera:
I think we need to show an error message when the active theme is a classic theme.

madhusudhand:
Verified the changes. It works great except for the following observations.

* Like @MaggieCabrera mentioned, `Create block theme` button is always active and generating the themes even if active theme is not a block theme.
* also the button on Theme details modal window is always active. Show only if its a block theme
* New theme version in styles set to `0.0.1` in styles.css but `1.0.0` in readme.txt
* Parent theme image not copied for child themes.
* README.md could be copied.
* readme.txt license year could be current year as it is created now. `Blockbase Child WordPress Theme, (C) 2021 Author`
* Edits done in site-editor are copied to child themes. But only index.html template is getting generated. Style edits to other areas such as footer are ignored.

pbking:
Thank you @madhusudhand  I have adjusted the readme version to match styles.css and also corrected the copyright year.

I have also disabled the plugin when the active theme is not a block theme.  It no longer shows as a menu item nor will the page load.

I'm not sure what you refer to by:
> the button on Theme details modal window is always active. Show only if its a block theme

Can you clarify please? I'm not sure which button on which modal you mean.

Regarding the readme.md:  If I export the theme a theme that has one that file is copied like any other.  Do you suggest we generate that file as well?  I'm not of the opinion that a readme.md should always be needed for a theme.  Blockbase is an exception I think but not many would I think.  My 'druthers is to leave that out from themes generated; it can be added later should it be warranted.

~~I am currently looking into your comments about some changes not being exported as expected.~~

I have not been able to reproduce this.  If I activate Blockbase, then make changes to the header, footer and 404 page template then when I observe the `/parts/header.html`, `/parts/footer.html` and `/templates/404.html` in the exported theme I see the changed I made.  This is true if I create a child theme, clone the theme or export it.  

~~I am looking into the screenshot issues.  There are a few problems with this at the moment I think...~~

There was a path issue adding the screenshot file.  This has been addressed.

MaggieCabrera:
I think what he means is that folders are being created even if no templates are inside them (like a child theme of archeo that only changes global styles, will produce the templates and parts folders with nothing in them)

MaggieCabrera (approved):
I think we can bring this in and keep iterating

madhusudhand:
> the button on Theme details modal window is always active. Show only if its a block theme

I meant the following button in theme preview. Seems like the buttons are derived from the menu and it is fixed now. :)

image

Copying readme.md isn't a good as for the new theme the contents may not be relevant. Though its optional, we may still consider optionally generate with given form information such as name, desc etc..
Optionally I mean, adding a checkboxes in the form and let user decide if a readme is required, just like how github asks while creating a repo. We could continue discussing this idea over a separate iteration.

For parts and templates, I will add the steps in a separate issue.

Verified other changes. LGTM.

madhusudhand (approved):

PR #51, opened 2022-03-11
Added a 'save' option 

### LINKED ISSUE #45: Add "Save" option
Perhaps we could consider a "save" option for the theme.  If the current option exports a .zip an additional option could be to "save" the theme where the changes get stored in the theme folder rather than exported as a .zip file.

This would allow the development experience of themes to happen in the FSE and be updated in what is my development environment allowing me to skip the export/unpack/copy/delete process that happens whenever I export the theme.

### LINKED ISSUE #50: Consider Clearing User Changes
When EXPORTING a theme if the theme with changes is installed again (or if something like saving a theme happens) then USER changes are still shown.   We may want to consider clearing the user changes when a theme is SAVED (essentially moving that changes from 'user space' to 'theme space' rather than COPYING from 'user space' to 'theme space'.

pbking:
Added a 'Save Theme' option to save the theme.json and template changes to local theme folder and clear user changes.

Addresses: #45 and enables a development workflow.

image

To Test:

* Start with a theme activated.  (I have done most testing with Archeo but ANY theme [parent or child, Blockbase or not] should work fine.)
* Make sure you begin with NO user changes to the theme (such as color, font or template changes)
* Make changes to the theme using the FSE (or the Customizer).  A good test for this is to change the color values of the theme colors.
* Make changes to a template part (for instance add a simple paragraph block)
* Make a change to a template (another simple change you can easily see)
* Create a brand-new template and template part.
* Use the Plugin to 'Overwrite'.
* Load the site (view).  Note that it still looks the same (global style changes and template changes are still in tact)
* Attempt to "Reset to Defaults" the Global Styles using the FSE (nothing changes)

* Attempt to "Clear Customizations" from the template part you previously edited.  Note that it is no longer an option.

* Observe the changes made to theme.json file in the theme in the filesystem.  Note that the Global Styles changes made in the FSE are now represented in the theme.json file in `/wp-content/themes/SLUG/theme.json`
* Note the changes made to templates in the theme filesystem, as well as additional templates/parts added based on the templates created with the FSE

This change also clears user settings when a theme is 'saved' addressing #50

### DIFF
--- a/admin/class-create-block-theme-admin.php
+++ b/admin/class-create-block-theme-admin.php
@@ -423,8 +423,8 @@ class Create_Block_Theme_Admin {
 
 			$template->content = _remove_theme_attribute_in_block_template_content( $template->content );
 
-			// NOTE: Dashes are replaced with \u002d in the content that we get (noteably in things like css variables used in templates)
-			// This replaces that with dashes again.
+			// NOTE: Dashes are encoded as \u002d in the content that we get (noteably in things like css variables used in templates)
+			// This replaces that with dashes again. We should consider decoding the entire string but that is proving difficult.
 			$template->content = str_replace( '\u002d', '-', $template->content );
 
 			if ( $new_slug ) {
@@ -452,8 +452,8 @@ class Create_Block_Theme_Admin {
 
 			$template_part->content = _remove_theme_attribute_in_block_template_content( $template_part->content );
 
-			// NOTE: Dashes are replaced with \u002d in the content that we get (noteably in things like css variables used in templates)
-			// This replaces that with dashes again.
+			// NOTE: Dashes are encoded as \u002d in the content that we get (noteably in things like css variables used in templates)
+			// This replaces that with dashes again. We should consider decoding the entire string but that is proving difficult.
 			$template_part->content = str_replace( '\u002d', '-', $template_part->content );
 
 			if ( $new_slug ) {
@@ -496,8 +496,8 @@ class Create_Block_Theme_Admin {
 
 			$template->content = _remove_theme_attribute_in_block_template_content( $template->content );
 
-			//NOTE: Dashes are replaced with \u002d in the content that we get (noteably in things like css variables used in templates)
-			// This replaces that with dashes again.
+			// NOTE: Dashes are encoded as \u002d in the content that we get (noteably in things like css variables used in templates)
+			// This replaces that with dashes again. We should consider decoding the entire string but that is proving difficult.
 			$template->content = str_replace( '\u002d', '-', $template->content );
 
 			file_put_contents(
@@ -522,8 +522,8 @@ class Create_Block_Theme_Admin {
 
 			$template_part->content = _remove_theme_attribute_in_block_template_content( $template_part->content );
 
-			//NOTE: Dashes are replaced with \u002d in the content that we get (noteably in things like css variables used in templates)
-			// This replaces that with dashes again.
+			// NOTE: Dashes are encoded as \u002d in the content that we get (noteably in things like css variables used in templates)
+			// This replaces that with dashes again. We should consider decoding the entire string but that is proving difficult.
 			$template_part->content = str_replace( '\u002d', '-', $template_part->content );
 
 			file_put_contents(
--- a/admin/class-create-block-theme-admin.php
+++ b/admin/class-create-block-theme-admin.php
@@ -273,18 +273,22 @@ class Create_Block_Theme_Admin {
 
 	function add_theme_json_to_zip ( $zip, $export_type ) {
 		$theme_json = MY_Theme_JSON_Resolver::export_theme_data( $export_type );
+		$theme_json = wp_json_encode( $theme_json, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE );
+		$theme_json = preg_replace( '~(?:^|\G)\h{4}~m', "\t", $theme_json );
 		$zip->addFromString(
 			'theme.json',
-			wp_json_encode( $theme_json, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES  )
+			$theme_json
 		);
 		return $zip;
 	}
 
 	function add_theme_json_to_local ( $export_type ) {
 		$theme_json = MY_Theme_JSON_Resolver::export_theme_data( $export_type );
+		$theme_json = wp_json_encode( $theme_json, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE );
+		$theme_json = preg_replace( '~(?:^|\G)\h{4}~m', "\t", $theme_json );
 		file_put_contents(
 			get_template_directory() . '/theme.json',
-			wp_json_encode( $theme_json, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES )
+			$theme_json
 		);
 	}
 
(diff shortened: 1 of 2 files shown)


pbking:
Thus far I've gotten this to SAVE the theme.json and templates.
It also REMOVES user Global Styles.
Currently trying to figure out how to delete user templates.

I have noticed that the json being exported is missing a few settings (such as `appearanceTools:true` and `lineHeight:true`.  These things will be missing from any created theme.

MaggieCabrera:
Does that happen too with https://github.com/WordPress/gutenberg/pull/39048 ? We should be using that function if we can and if it's not working correctly, patch upstream?

pbking:
> Does that happen too with https://github.com/WordPress/gutenberg/pull/39048 ? We should be using that function if we can and if it's not working correctly, patch upstream?

The problem is also in Gutenberg's export.  I'll be reporting that in detail directly.

This plugin does not yet use the logic to build theme.json from Gutenberg. Unfortunately that logic (added in #39048) isn't exposed in a way that we can take advantage of.  But it is done in the same way as this plugin so the result is the same.

MaggieCabrera:
@scruffian to keep you in the loop

scruffian:
cc @draganescu

admin/class-create-block-theme-admin.php
```diff
@@ -327,6 +374,11 @@ function add_templates_to_zip( $zip, $export_type ) {
 			}
 
 			$template->content = _remove_theme_attribute_in_block_template_content( $template->content );
+
+			//NOTE: Dashes are replaced with \u002d in the content that we get (noteably in things like css variables used in templates)
+			// This replaces that with dashes again.
+			$template->content = str_replace( '\u002d', '-', $template->content );
```
MaggieCabrera: this looks like the original content was encoded, is it maybe [this](https://github.com/WordPress/gutenberg/blob/7aa042605ff42bb437e650c39132c0aa8eb4ef95/lib/global-styles.php#L28) what's happening? Maybe we can use the decode function instead in case there's more than dashes that we should consider?

pbking: Ah, very good call.  Thank you.  That felt pretty shabby when I wrote it. :P

pbking: I tried very hard to do this. :(
I can't figure out how. 

AFAICT this is something that should be addressed in Gutenberg and until it is perhaps this "good enough" solution will be ok?

MaggieCabrera: I mentioned it to @scruffian on his latest PR to the export functionality in GB, hopefully there will be a fix for it.

draganescu:
Just arrived here. Thanks for the ping @scruffian As I'll start working on https://github.com/WordPress/gutenberg/issues/39194 this is a great inspiration (a.k.a. place to "steal" from haha).

My thinking on the core ticket was more to create one single REST endpoint that hanles this and to send from the editor only one command: `update_theme_files`, only when the theme was in "development" mode. I think if a theme is in development clicking update in the site editor should update the files themselves.

pbking:
> As I'll start working on https://github.com/WordPress/gutenberg/issues/39194 this is a great inspiration (a.k.a. place to "steal" from haha).

That's the idea, please steal away.  I just consider this a playground to determine what may be best to surface in Gutenberg.  I'm testing out this functionality as much as I can for the development of [Pendant](https://github.com/Automattic/themes/milestone/186).

I would love to lend a hand on #39194 @draganescu and will keep an eye on it.  I didn't even actually know that issue existed until now. :)

admin/class-create-block-theme-admin.php
scruffian: I think this actually happens when the template is modified using the site editor...

admin/class-create-block-theme-admin.php
```diff
@@ -414,6 +469,70 @@ function add_templates_to_zip( $zip, $export_type, $new_slug ) {
 		return $zip;
 	}
 
+	function add_templates_to_local( $export_type ) {
+		//NOTE: This was heavily copied from add_templates_to_zip
```
scruffian: Yeah I agree this needs a refactor ideally.

admin/class-create-block-theme-admin.php
```diff
@@ -236,11 +275,19 @@ function add_theme_json_to_zip ( $zip, $export_type ) {
...
 		$zip->addFromString(
 			'theme.json',
-			wp_json_encode( $theme_json, JSON_PRETTY_PRINT )
+			wp_json_encode( $theme_json, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES  )
 		);
 		return $zip;
 	}
 
+	function add_theme_json_to_local ( $export_type ) {
+		$theme_json = MY_Theme_JSON_Resolver::export_theme_data( $export_type );
```
scruffian: Does this depend on another plugin? That's not ideal...

pbking: It's not another plugin, just [this](https://github.com/Automattic/create-block-theme/blob/trunk/admin/gutenberg_additions.php) that we had already added previously.  All stuff that we had hoped would ultimately eventually be done by Gutenberg... maybe.

scruffian:
When I do a save I see this notice: 

"New block theme created!"

We should update this to be more descriptive...

scruffian:
This converts tabs to spaces in theme.json. I think we should keep tabs for now.

admin/class-create-block-theme-admin.php
scruffian: It looks like this suffers from the same issues as the Gutenberg exporter, in that it removes $schema and appearanceTools keys. Feel free to use the code from that PR until it is resolved...

pbking:
> This converts tabs to spaces in theme.json. I think we should keep tabs for now.

I agree, but how do you suggest we do that?  I'm not sure that I'm keen for a code formatting to hold up the merge, but I would ultimately like to provide tabs, not spaces.  Just not sure how to do that with PHP.

scruffian:
> I agree, but how do you suggest we do that? I'm not sure that I'm keen for a code formatting to hold up the merge, but I would ultimately like to provide tabs, not spaces. Just not sure how to do that with PHP.

https://github.com/WordPress/gutenberg/pull/39751!

pbking:
Thanks for feedback, I have made some changes:

Message when saving:
image

theme.json is now saved with tabs, not spaces 

The export code is refactored to be a lot more dry.

There is still an issue with $schema and appearanceTools, as does the Gutenberg exporting logic and I'm keen to let that settle before doing anything special here.  I don't believe it's a blocker for now.

The unencoding is still hacked in, also looking for a better solution in Gutenberg and in the meantime the limited "decoding" that it's doing now I think is enough for what we're doing.

scruffian:
> The unencoding is still hacked in, also looking for a better solution in Gutenberg and in the meantime the limited "decoding" that it's doing now I think is enough for what we're doing.

The resolution for this in Gutenberg was to do nothing  - see https://github.com/WordPress/gutenberg/issues/39576

scruffian (approved):
LGTM

I saw a couple of notices but I think they are unrelated to this PR:

[28-Mar-2022 11:26:13 UTC] PHP Notice:  Undefined variable: old_slug in /var/www/html/wp-content/plugins/create-block-theme/admin/class-create-block-theme-admin.php on line 426
[28-Mar-2022 11:26:13 UTC] PHP Notice:  Undefined variable: old_slug in /var/www/html/wp-content/plugins/create-block-theme/admin/class-create-block-theme-admin.php on line 439

pbking:
> I saw a couple of notices but I think they are unrelated to this PR:

They are caused by a recent change I did cleaning up the code.  Thanks for catching.  Patching that up.

PR #52, opened 2022-03-15
Refactor all of the namespaced items during cloning

### LINKED ISSUE #43: Refactor Namespaced Resources
When CLONING a PARENT or creating a SIBLING theme (cloning a child theme) the resulting resources could be namespaced.

For example functions like this:
```
function twentytwentytwo_support() {
```
should be refactored to:
```
function my_new_theme_support() {
```
and 
```
wp_register_style('twentytwentytwo-style',...
```
to
```
wp_register_style('my-new-theme-style',...
```

We should consider the following resources when refactoring the namespace:
```
functions.php
??? what else ???
```

pbking:
This addresses #43

When a theme is cloned (either a CHILD theme to make a sibling or a PARENT theme to make a completely new theme) the resources that are copied from the source theme are modified to have the namespaced resources changed to the new name.

The only (current) stipulation is that if the SOURCE theme has a single-word name (such as 'apples') and the DESTINATION theme has a multi-word name (such as 'red-fruit') the export is prevented.

This is because of the naming differences where some resources should have underscores (such as function names in PHP files) and some should have hyphens (such as text-domain values... also in PHP files).  Determining which instance should be refactored to which is much more complicated than the existing logic (which covers most use-cases).

TODO: Currently the failing situation noted above doesn't show an error on the form page but instead is just dumping the error message to the page.

To Test:

* Activate a Block Theme.  A theme with patterns is the best test.  (I used Archeo)
* Make changes to templates and styles.
* CLONE the theme using the Create Block Theme plugin
* Install and activate the created theme, note that the style and templates mirror the source theme, but with the changes made.
* Observe the resources.  Note that any resources referencing the original theme slug now reference the new theme slug.

### DIFF
--- a/admin/class-create-block-theme-admin.php
+++ b/admin/class-create-block-theme-admin.php
@@ -257,11 +257,11 @@ class Create_Block_Theme_Admin {
 		);
 
 		// Add all the files (except for templates)
-		foreach ( $files as $name => $file )
-		{
+		foreach ( $files as $name => $file ) {
+
 			// Skip directories (they would be added automatically)
-			if ( ! $file->isDir() )
-			{
+			if ( ! $file->isDir() ) {
+
 				// Get real and relative path for current file
 				$file_path = $file->getRealPath();
 
@@ -276,15 +276,26 @@ class Create_Block_Theme_Admin {
 				}
 
 				$relative_path = substr( $file_path, strlen( $theme_path ) + 1 );
-				$contents = file_get_contents( $file_path );
 
-				// Replace namespace values if provided
-				if ( $new_slug ) {
-					$contents = $this->replace_namespace( $contents, $new_slug );
+				// Replace only text files, skip png's and other stuff.
+				$valid_extensions = array( 'php', 'css', 'scss', 'js', 'txt', 'html' );
+				$valid_extensions_regex = implode( '|', $valid_extensions );
+				if ( ! preg_match( "/\.({$valid_extensions_regex})$/", $relative_path ) ) {
+					$zip->addFile( $file_path, $relative_path );
+				}
+				
+				else {
+					$contents = file_get_contents( $file_path );
+
+					// Replace namespace values if provided
+					if ( $new_slug ) {
+						$contents = $this->replace_namespace( $contents, $new_slug );
+					}
+
+					// Add current file to archive
+					$zip->addFromString( $relative_path, $contents );
 				}
 
-				// Add current file to archive
-				$zip->addFromString( $relative_path, $contents );
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


MaggieCabrera:
I don't see why we can't replace namespaces without spaces, in my testing of course the first replace overwritten the first, but having dashes instead of underscores for function names is not that terrible compared to having to rename everything manually. And in this cases dashes is the way to go since the only thing that's going to be sensitive to that are translatable string, that need to match the text domain of the theme.

admin/class-create-block-theme-admin.php
```diff
@@ -275,15 +272,47 @@ function copy_theme_to_zip( $zip ) {
...
 				}
 
-				// Add current file to archive
 				$relative_path = substr( $file_path, strlen( $theme_path ) + 1 );
-				$zip->addFile( $file_path, $relative_path );
+				$contents = file_get_contents( $file_path );
+
+				// Replace namespace values if provided
+				if ( $new_slug ) {
+					$contents = $this->replace_namespace( $contents, $new_slug );
```
MaggieCabrera: we should only do these for certain file extensions. Right now it's trying to do it to images too. Let's go with `array( 'php', 'css', 'scss', 'js', 'txt', 'html' );` like the emptytheme script does

pbking: Ahh, yes.  Good call.

admin/class-create-block-theme-admin.php
```diff
@@ -39,7 +39,7 @@ function export_child_theme( $theme ) {
 		$zip = $this->create_zip( $filename );
 
 		$zip = $this->copy_theme_to_zip( $zip );
-		$zip = $this->add_templates_to_zip( $zip, 'current' );
```
pbking, line 42: These were missing on purpose.  Perhaps we should pass null so that's clear?

The namespace is only refactored if the slug is passed and in these cases there is no refactoring needed.

MaggieCabrera: I was getting errors when I chose the other options, so yeah, maybe pass null

pbking: It probably would have been wise of me to have actually tested those other options when I changed that instead of assuming that things worked like I thought they would. :P

admin/class-create-block-theme-admin.php
pbking: Done

admin/class-create-block-theme-admin.php
pbking: done.  passing null.  Tested other methods and seem to work as expected.

pbking:
Fixed a small bug where references to the theme NAME weren't being refactored.

Tested this with 'Block Canvas' and observed `functions.php` to confirm that they are being converted.

MaggieCabrera (approved):
Thanks for your hard work with this, it's working great on my env now!

PR #55, opened 2022-04-29
Remove flatten_theme_json logic

pbking:
Remove flatten_theme_json logic and replace `get_raw_data()` with a call to `get_data()` which returns the values already flattened, organized and better cleaned up in general.


scruffian (approved):
Yes!

PR #60, opened 2022-05-27
Remove .DS_Store file

### LINKED ISSUE #54: Remove .DS_Store from repo
.DS_Store should not be in the repo.

See this S.O. topic https://stackoverflow.com/questions/107701/how-can-i-remove-ds-store-files-from-a-git-repository

mikachan:
Small PR to remove the `.DS_Store` file. We're already ignoring it in [.gitignore](https://github.com/Automattic/create-block-theme/blob/trunk/.gitignore#L26), it looks like it just needs to be deleted as well.

Closes https://github.com/Automattic/create-block-theme/issues/54


matiasbenedetto (approved):
:shipit:

PR #67, opened 2022-06-14
fix issue where child theme overrides parent theme templates

### LINKED ISSUE #64: Override option on child theme overrides child theme changes and user changes on parent theme
#### Steps:

* Activate a child theme such as `Quadrat` and use override option and export theme.
* Observe that it overrides parent theme `blockbase`

image

madhusudhand:
### Fix summary

Used [get_stylesheet_directory](https://developer.wordpress.org/reference/functions/get_stylesheet_directory/) to fetch the current theme path and override the theme templates. It works for both child and parent themes.

Also, updated git repo references.

### Testing instructions

* Activate a child theme
* add customisations to templates in FSE
* select "Appearance > Create Block Theme > Override <theme>
* click export

Observe that, parent theme is not disturbed and the customisations are changed only in the child theme.

### Related issues

Fixes: #64

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Creates a new theme by cloning the activated theme. The resulting theme will hav
 Saves user's changes to the theme files and deletes the user's changes.
 
 ### Step 1 – Setup
-Install and activate the [Create Block Theme](https://github.com/Automattic/create-block-theme) plugin. You will also need to be running the latest version of the Gutenberg plugin.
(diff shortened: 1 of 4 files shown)


pbking:
Ahh, good catch and good fix.  TY @madhusudhand.

Fix looks good and tests out.
🚢

pbking (approved):

PR #70, opened 2022-06-17
Add option to create blank theme

### LINKED ISSUE #68: Create boilerplate theme
Sorry if this question is naive!

What do you think about adding a fifth option to the plugin — create a new boilerplate theme from scratch, maybe based on block-canvas or something like it? 

When I meet folks that are new to block theming, the first thing they always ask is for a reference point — a starter, boilerplate block theme. Also the name of this plugin implies it could create a "new" theme, to me, maybe disconnected from the currently active one. (This is the perspective of someone who has used "create-react-app" before coming to WordPress, and using that utility results in a blank application, granted there's more tooling and overhead involved there because of webpack...) 

This might make more sense as a CLI or maybe even in Gutenberg itself, but wanted to ask in any case.

jffng:
This PR is an idea for addressing #68 — an option for generating an "empty" block theme inside the current site's themes directory.

The boilerplate is pretty much [block-canvas](https://github.com/Automattic/themes/tree/trunk/block-canvas). I thought it would be best to put the boilerplate code inside the plugin so it could all be maintained in one place.

**To test**
- Check out the PR
- Select the last option, give the theme a name
- Verify you can activate the theme and edit it via the site editor


assets/boilerplate/theme.json
```diff
@@ -0,0 +1,303 @@
+{
+    "$schema": "https://schemas.wp.org/trunk/theme.json",
+    "settings": {
```
jffng, line 3: I removed the `customTemplates` but could be convinced to add them back...

admin/class-create-block-theme-admin.php
```diff
@@ -598,6 +654,69 @@ function build_child_style_css( $theme ) {
...
+		);
+
+		// Enqueue theme stylesheet.
+		wp_enqueue_style( '{$slug}-style' );
+
+	}
+
+endif;
+
+add_action( 'wp_enqueue_scripts', '{$slug}_styles' );";
```
carolinan: Maybe we shouldn't include this and instead encourage theme developers to use the block styles as much as possible?

assets/boilerplate/parts/header.html
```diff
@@ -0,0 +1,30 @@
...
+		<!-- wp:navigation {"__unstableLocation":"primary","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right","orientation":"horizontal"},"style":{"spacing":{"margin":{"top":"0"}}}} /-->
+	
+	</div>
+	<!-- /wp:group -->
+
+</div>
+<!-- /wp:group -->
+
+<!-- wp:spacer {"height":50} -->
+<div style="height:50px" aria-hidden="true" class="wp-block-spacer"></div>
```
carolinan, line 28: I recommend using the new markup for the spacer block.

assets/boilerplate/parts/post-meta.html
```diff
@@ -0,0 +1,8 @@
+<!-- wp:group {"layout":{"type":"flex"}} -->
+<div class="wp-block-group">
+	<!-- wp:post-author {"showAvatar":false,"showBio":false,"style":{"typography":{"fontSize":"14px"}}} /-->
+	<!-- wp:post-date {"isLink":true,"style":{"typography":{"fontSize":"14px"}}} /-->
+	<!-- wp:post-terms {"term":"category","style":{"typography":{"fontSize":"14px"}}} /-->
+	<!-- wp:post-terms {"term": "post_tag","style":{"typography":{"fontSize":"14px"}}} /-->
+</div>
+<!-- /wp:group -->
```
carolinan: We can reduce this by:
- Setting font size in theme.json
- Or setting the font-size on the group.

assets/boilerplate/style.css
```diff
@@ -0,0 +1,103 @@
+/*
```
carolinan: This is is a lot to remove for any users who wants to create their own theme.
May I recommend limiting this to the alignments that prevents content from running up to the edges?

assets/boilerplate/theme.json
```diff
@@ -0,0 +1,303 @@
...
+                {
+                    "name": "Extra Large",
+                    "size": "2rem",
+                    "slug": "x-large"
+                }
+            ],
+            "lineHeight": true
+        }
+    },
+    "styles": {
```
carolinan: This is a lot of styles for someone to remove.
Especially things like spacing below the gallery and heading font weight.

assets/boilerplate/theme.json
```diff
@@ -0,0 +1,303 @@
...
+                    "size": "1.75rem",
+                    "slug": "large"
+                },
+                {
+                    "name": "Extra Large",
+                    "size": "2rem",
+                    "slug": "x-large"
+                }
+            ],
+            "lineHeight": true
```
carolinan: This lineHeight can be removed because appearance tools is already enabled :)
https://developer.wordpress.org/block-editor/reference-guides/theme-json-reference/theme-json-living/#appearancetools

assets/boilerplate/parts/header.html
```diff
@@ -0,0 +1,30 @@
+<!-- wp:group {"className":"gapless-group","layout":{"inherit":"true"}} -->
+<div class="wp-block-group gapless-group">
```
carolinan: What is `gapless-group`?

assets/boilerplate/theme.json
```diff
@@ -0,0 +1,303 @@
...
+        },
+        "layout": {
+            "contentSize": "620px",
+            "wideSize": "1000px"
+        },
+        "spacing": {
+            "blockGap": true,
+            "margin": true,
+            "padding": true,
+            "units": [
```
carolinan, line 10: blockGap, margin, padding are already enabled through appearance tools.

assets/boilerplate/theme.json
```diff
@@ -0,0 +1,303 @@
+{
+    "$schema": "https://schemas.wp.org/trunk/theme.json",
+    "settings": {
+        "appearanceTools": true,
+        "border": {
+            "color": true,
+            "radius": true,
+            "style": true,
+            "width": true
```
carolinan: These border options are already enabled through appearancetools.

assets/boilerplate/theme.json
```diff
@@ -0,0 +1,303 @@
...
+        },
+        "color": {
+            "custom": true,
+            "customGradient": true,
+            "link": true,
```
carolinan: This is becoming repetitive :p Link color is also enabled by appearancetools.

assets/boilerplate/theme.json
```diff
@@ -0,0 +1,303 @@
...
+        "appearanceTools": true,
+        "border": {
+            "color": true,
+            "radius": true,
+            "style": true,
+            "width": true
+        },
+        "color": {
+            "custom": true,
+            "customGradient": true,
```
carolinan: Custom color and custom gradients are enabled by default, so these two lines can be removed.
https://developer.wordpress.org/block-editor/reference-guides/theme-json-reference/theme-json-living/#color

assets/boilerplate/theme.json
```diff
@@ -0,0 +1,303 @@
...
+                "%",
+                "px",
+                "em",
+                "rem",
+                "vh",
+                "vw"
+            ]
+        },
+        "typography": {
+            "customFontSize": true,
```
carolinan: customFontSize is enabled by default, this can be removed.
https://developer.wordpress.org/block-editor/reference-guides/theme-json-reference/theme-json-living/#typography

assets/boilerplate/parts/header.html
jffng: What is the new markup?

assets/boilerplate/style.css
scruffian: I wonder if it would be better to ship this with no alignment code at all?

admin/class-create-block-theme-admin.php
```diff
@@ -271,6 +271,62 @@ function export_theme( $theme ) {
 		die();
 	}
 
+	function create_blank_theme( $theme ) {
```
scruffian, line 274: A lot of this code looks similar to the theme export process in core, Could we reuse some of that code?

admin/class-create-block-theme-admin.php
scruffian: yeah, can we remove functions.php entirely?

scruffian (commented):
I think this looks like a great start. I'd like to see if we can remove all the CSS and functions.php so we can make things simpler.

assets/boilerplate/style.css
carolinan: It might be, because everyone would need to remove that once that spacing is solved upstream.

jffng:
Thank you for the reviews @carolinan @scruffian! 

I removed all the CSS, as well as functions.php. Do you think there are any other simplifications to make to the boilerplate? 

I will look into re-using some of the theme export code in core.

assets/boilerplate/templates/index.html
```diff
@@ -0,0 +1,39 @@
+<!-- wp:template-part {"slug":"header","tagName":"header"} /-->
+
+<!-- wp:query {"tagName":"main","layout":{"inherit":true}} -->
+<main class="wp-block-query">
+	<!-- wp:post-template -->
+	<!-- wp:group -->
+	<div class="wp-block-group">
+		<!-- wp:post-title {"isLink":true} /-->
+		<!-- wp:post-featured-image {"isLink":true} /-->
+		<!-- wp:post-excerpt /-->
```
scruffian: I think post-content would be better than post-excerpt.

scruffian:
> I will look into re-using some of the theme export code in core.

This could be done in a followup

scruffian (approved):
LGTM

assets/boilerplate/parts/header.html
carolinan: It includes the unit, like this:
```
<!-- wp:spacer {"height":"50px"} -->
<div style="height:50px" aria-hidden="true" class="wp-block-spacer"></div>
<!-- /wp:spacer -->
```

PR #76, opened 2022-07-04
Improve form instructions

### LINKED ISSUE #44: Improve Instructions
The instructions and descriptions given on the form are not very clear and designed poorly.

Update the form to give clear instructions and descriptions of the options. (What should this be?  How should it be designed?)

Update the form to better handle required fields (Name should be required when prompted).  Javascript might be necessary to handle this.

jonathanbossenger:
Fixes #44 

- Adds context for each field
- Indicates which fields are required for a theme to load in the WordPress dashboard

### DIFF
--- a/admin/class-create-block-theme-admin.php
+++ b/admin/class-create-block-theme-admin.php
@@ -677,11 +677,12 @@ Tags: one-column, custom-colors, custom-menu, custom-logo, editor-style, feature
 				<?php _e('[Generates a boilerplate "empty" theme inside of this site\'s themes directory.]', 'create-block-theme'); ?></label><br /><br />
 
 				<div hidden id="new_theme_metadata_form">
-					<label><?php _e('Theme Name', 'create-block-theme'); ?><br /><input type="text" name="theme[name]" class="regular-text" /></label><br /><br />
-					<label><?php _e('Theme Description', 'create-block-theme'); ?><br /><textarea rows="4" cols="50" name="theme[description]" class="regular-text"></textarea></label><br /><br />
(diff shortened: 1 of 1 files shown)


admin/class-create-block-theme-admin.php
```diff
@@ -677,11 +677,12 @@ function create_admin_form_page() {
...
 
 				<div hidden id="new_theme_metadata_form">
-					<label><?php _e('Theme Name', 'create-block-theme'); ?><br /><input type="text" name="theme[name]" class="regular-text" /></label><br /><br />
-					<label><?php _e('Theme Description', 'create-block-theme'); ?><br /><textarea rows="4" cols="50" name="theme[description]" class="regular-text"></textarea></label><br /><br />
-					<label><?php _e('Theme URI', 'create-block-theme'); ?><br /><input type="text" name="theme[uri]" class="regular-text code" /></label><br /><br />
-					<label><?php _e('Author', 'create-block-theme'); ?><br /><input type="text" name="theme[author]" class="regular-text" /></label><br /><br />
-					<label><?php _e('Author URI', 'create-block-theme'); ?><br /><input type="text" name="theme[author_uri]" class="regular-text code" /></label><br /><br />
+                    <p><?php _e('Items indicated with (*) are required.', 'create-block-theme'); ?></p>
+					<label><?php _e('Theme Name (*): The name of the theme.', 'create-block-theme'); ?><br />  <input type="text" name="theme[name]" class="regular-text" /></label><br /><br />
```
MaggieCabrera, line 681: I thought we had the required attr on this input, maybe it got lost in one of the reactors that we've done to the plugin

MaggieCabrera (approved):
Thanks for this! I think we could follow up on this with other improvements too, like proper placeholders for example, and the required attr I mentioned on the comment

PR #77, opened 2022-07-04
Form cleanup and Theme name check

MaggieCabrera:
Looks like some of the changes we made to the form got lost among the refactors, so I've made a few changes:

- Added placeholder text to the inputs
- Used a two-column layout for the form (I wanted to make the form fields bigger without adding extra CSS and this is what I found that uses core CSS without needing to add some in the plugin).
- Made the theme name a requirement for the cases that show the form to the user.

I'm open to improvements, I don't think it looks the best. Maybe we should audit the dev experience in general of the plugin and have another pass at the design and copy that we are using in general. Maybe we don't want to do that just yet if we want to integrate some of this functionality with the site editor instead of this standalone plugin page. Feel free to discuss this!

Before:

Screenshot 2022-07-04 at 17 51 58

After:

Screen Capture on 2022-07-04 at 17-45-07

### DIFF
--- a/admin/class-create-block-theme-admin.php
+++ b/admin/class-create-block-theme-admin.php
@@ -655,11 +655,11 @@ Tags: one-column, custom-colors, custom-menu, custom-logo, editor-style, feature
 		}
 		?>
 		<div class="wrap">
+			<h2><?php _e('Create Block Theme', 'create-block-theme'); ?></h2>
 			<form method="get">
 				<div id="col-container">
 					<div id="col-left">
 						<div class="col-wrap">
-							<h2><?php _e('Create Block Theme', 'create-block-theme'); ?></h2>
 							<p><?php _e('Export your current block theme with changes you made to Templates, Template Parts and Global Styles.', 'create-block-theme'); ?></p>
 
 							<label>
--- a/admin/class-create-block-theme-admin.php
+++ b/admin/class-create-block-theme-admin.php
@@ -843,9 +843,11 @@ Tags: one-column, custom-colors, custom-menu, custom-logo, editor-style, feature
 	}
 
 	function admin_notice_blank_success() {
+		$theme_name = $_GET['theme']['name'];
+
 		?>
 			<div class="notice notice-success is-dismissible">
-				<p><?php _e( 'Blank theme created, head over to Appearance > Themes to activate it!', 'create-block-theme' ); ?></p>
+				<p><?php printf( esc_html__( 'Blank theme created, head over to Appearance > Themes to activate %1$s', 'create-block-theme' ), esc_html( $theme_name ) ); ?></p>
 			</div>
 		<?php
 	}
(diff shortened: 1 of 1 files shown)


madhusudhand:
The form is much organised now. 👍 
I have the following observations/suggestions over all.

1. Theme name requirement validation works for `create child of` and `clone` options. But not for `create blank theme`.
It gets submitted without validations and does nothing.

Also, the message could be improved I guess. It doesn't explain the new theme name.

image

Something like: Blank theme created, head over to Appearance > Themes to activate **theme-name**!

2. While the button label is "Export theme" not all options export the theme. Only first three export a theme and other two create or override. Either the name of the button could be toggled or keep a general name like "create block theme".

admin/class-create-block-theme-admin.php
```diff
@@ -656,37 +656,96 @@ function create_admin_form_page() {
...
-					<label><?php _e('Theme Description: A short description of the theme.', 'create-block-theme'); ?><br /><textarea rows="4" cols="50" name="theme[description]" class="regular-text"></textarea></label><br /><br />
-					<label><?php _e('Theme URI: The URL of a public web page where users can find more information about the theme.', 'create-block-theme'); ?><br /><input type="text" name="theme[uri]" class="regular-text code" /></label><br /><br />
-					<label><?php _e('Author: The name of the individual or organization who developed the theme.', 'create-block-theme'); ?><br /><input type="text" name="theme[author]" class="regular-text" /></label><br /><br />
-					<label><?php _e('Author URI: The URL of the authoring individual or organization.', 'create-block-theme'); ?><br /><input type="text" name="theme[author_uri]" class="regular-text code" /></label><br /><br />
+				<div id="col-container">
+					<div id="col-left">
+						<div class="col-wrap">
+							<p><?php _e('Export your current block theme with changes you made to Templates, Template Parts and Global Styles.', 'create-block-theme'); ?></p>
```
madhusudhand: Let's add the theme name here?

Export your current block theme **[current-theme-name]** with changes you made to Templates, Template Parts and Global Styles.

MaggieCabrera:
> The form is much organised now. 👍 I have the following observations/suggestions over all.
> 
>     1. Theme name requirement validation works for `create child of` and `clone` options. But not for `create blank theme`.
>        It gets submitted without validations and does nothing.
> 
> Also, the message could be improved I guess. It doesn't explain the new theme name.
> Something like: Blank theme created, head over to Appearance > Themes to activate **theme-name**!
> 
>     2. While the button label is "Export theme" not all options export the theme. Only first three export a theme and other two create or override. Either the name of the button could be toggled or keep a general name like "create block theme".

Done! thanks for the review!

pbking:
>Maybe we should audit the dev experience in general of the plugin and have another pass at the design and copy that we are using in general.   Maybe we don't want to do that just yet if we want to integrate some of this functionality with the site editor instead of this standalone plugin page.

I'm very keen to spend more time transitioning this tool into the FSE as opposed to spending much more time on this form.  I think that what we've got here is a great improvement and works well.  I think we should take this as it stands!

🚢

pbking (approved):

PR #87, opened 2022-07-18
Automate/deployment

### LINKED ISSUE #83: Automate WordPress.org deployment
As we bring changes in it would be fantastic to ship those changes automatically to [WordPress.org](https://wordpress.org/plugins/create-block-theme/)

[This](https://github.com/marketplace/actions/wordpress-plugin-deploy) is a good github action to automate this.

pbking:
This change adds a deployment action triggered when a tag is created.

To deploy a build create a tag from /trunk labeled the same as the version to be shipped.

For this to work SVN_USERNAME and SVN_PASSWORD need to be stored as secrets on this account.  (I'm not sure who has access to that for this repository.  I do not.)

Fixes: #83

### DIFF
--- a/.github/workflows/deploy-to-dotorg.yml
+++ b/.github/workflows/deploy-to-dotorg.yml
@@ -1,0 +1,16 @@
+name: Deploy to WordPress.org
+on:
+  push:
+    tags:
+    - "*"
+jobs:
+  tag:
+    name: New tag
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


dd32:
> For this to work SVN_USERNAME and SVN_PASSWORD need to be stored as secrets on this account.

Secrets have been enabled and confirmed it has access to the destination repo.

madhusudhand (approved):
Now that secrets are in place, this can be merged!

PR #88, opened 2022-07-20
Get the correct merged theme.json data

### LINKED ISSUE #82: Exporting with only Styles applied does not seem to work.
Apologies in advance if this is expected, or if I am doing something wrong, but I discovered that by creating a blank block theme, applying only some styles via the Styles panel in the Site Editor, when I export or clone the theme using Create Block Theme, the theme.json doesn't seem to be updated with the new styles.

However, when I export the theme using the Site Editor Export option, the theme.json is updated as expected.

Steps to reproduce:

1. Create a block theme with only an empty `style.css` (except for the required header), `theme.json` with basic content (below), and a blank `templates/index.html` page.
2. Activate the theme, and only apply some styles like background image, text color, etc via Site Editor, and save.
3. Use the plugin page to either Export or Clone the theme

Original theme.json
```
{
  "version": 2,
  "settings": {
    "appearanceTools": true
  }
}
```
Exported/cloned theme.json

```
{
	"settings": {
		"appearanceTools": true
	},
	"version": 2
}
```

Expected theme.json (which I get via the Site Editor Export option).

```
{
	"$schema": "https://schemas.wp.org/wp/6.0/theme.json",
	"settings": {
		"appearanceTools": true
	},
	"styles": {
		"color": {
			"background": "var:preset|color|pale-cyan-blue",
			"text": "var:preset|color|vivid-cyan-blue"
		},
		"elements": {
			"link": {
				"color": {
					"text": "var:preset|color|vivid-purple"
				}
			}
		},
		"spacing": {
			"padding": {
				"bottom": "20px",
				"left": "20px",
				"right": "20px",
				"top": "20px"
			}
		}
	},
	"version": 2
}
```

jonathanbossenger:
Fixes #82 

This seems to have been a preexisting bug, as just calling get_data on the merged $theme appears to work across all block themes. However, it would be ideal if this was tested and confirmed.

Tested using the empty theme.json from the original issue, as well as Blockbase.

### DIFF
--- a/admin/resolver_additions.php
+++ b/admin/resolver_additions.php
@@ -38,7 +38,7 @@ function augment_resolver_with_utilities() {
 
 			$theme->merge( static::get_user_data() );
 
-			$data = MY_Theme_JSON_Resolver::$theme->get_data();
+			$data = $theme->get_data();
 
 			$theme_json = wp_json_encode( $data, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE );
 			return preg_replace ( '~(?:^|\G)\h{4}~m', "\t", $theme_json );


madhusudhand (approved):
Great catch. Thanks you for the fix.
Tested various scenarios and It works as expected.

jonathanbossenger:
Thank you @madhusudhand

PR #90, opened 2022-07-26
Save a theme variation

### LINKED ISSUE #89: New option: save style variation
With the incoming [new default theme](https://make.wordpress.org/design/2022/07/19/proposal-a-new-kind-of-default-theme/) being style variation focused, this plugin could have an option to generate Global style changes into a variation instead of creating a new theme. This will help users unfamiliar with theme.json syntax to create a new style variation for a theme.

MaggieCabrera:
This new option saves the users global styles changes into a variation instead of a child theme or a new clone of the theme. I opted in for the easiest way to do this, saving the file locally with a generic file name and downloading the zip file, but it's up for discussion if we want to do it this way:

- Should we ask the user for a variation name?
- Should the exported ZIP file only include the variation or the whole theme included like in this case?
- Is the wording on the new option clear for what the feature does?

Screenshot 2022-07-26 at 10 17 22

Closes #89

### DIFF
--- a/admin/class-create-block-theme-admin.php
+++ b/admin/class-create-block-theme-admin.php
@@ -33,8 +33,9 @@ class Create_Block_Theme_Admin {
 		$this->add_theme_json_to_local( $export_type );
 	}
 
-	function save_variation_locally ( $export_type ) {
+	function save_variation ( $export_type, $theme ) {
 		$this->add_theme_json_variation_to_local( $export_type );
+		$this->export_theme( $theme );
 	}
 
 	function clear_user_customizations() {
@@ -801,12 +802,13 @@ Tags: one-column, custom-colors, custom-menu, custom-logo, editor-style, feature
 
 			else if ( $_GET['theme']['type'] === 'variation' ) {
 				if ( is_child_theme() ) {
-					$this->save_variation_locally( 'current' );
+					$this->save_variation( 'current', $_GET['theme'] );
 				}
 				else {
-					$this->save_variation_locally( 'all' );
+					$this->save_variation( 'all', $_GET['theme'] );
 				}
-				die('eeeee');
+
+				add_action( 'admin_notices', [ $this, 'admin_notice_variation_success' ] );
 			}
 
 			else if ( $_GET['theme']['type'] === 'blank' ) {
@@ -880,6 +882,16 @@ Tags: one-column, custom-colors, custom-menu, custom-logo, editor-style, feature
 		?>
 			<div class="notice notice-success is-dismissible">
 				<p><?php printf( esc_html__( 'Blank theme created, head over to Appearance > Themes to activate %1$s', 'create-block-theme' ), esc_html( $theme_name ) ); ?></p>
+			</div>
+		<?php
+	}
+
+	function admin_notice_variation_success() {
+		$theme_name = $_GET['theme']['name'];
+
+		?>
+			<div class="notice notice-success is-dismissible">
+				<p><?php printf( esc_html__( 'Your variation of %1$s has been created succesfully', 'create-block-theme' ), esc_html( $theme_name ) ); ?></p>
 			</div>
 		<?php
 	}
--- a/admin/class-create-block-theme-admin.php
+++ b/admin/class-create-block-theme-admin.php
@@ -339,10 +339,22 @@ class Create_Block_Theme_Admin {
 	function add_theme_json_variation_to_local ( $export_type, $theme ) {
 		$variation_slug = sanitize_title( $theme['variation'] );
 		$variation_path = get_stylesheet_directory() . DIRECTORY_SEPARATOR . 'styles' . DIRECTORY_SEPARATOR;
+		$file_counter = 0;
+
 		if ( ! file_exists( $variation_path ) ) {
 			mkdir( $variation_path, 0755, true );
 		}
+		
+		if ( file_exists( $variation_path . $variation_slug . '.json' ) ) {
+			$file_counter++;
+			while ( file_exists( $variation_path . $variation_slug . '_' . $file_counter . '.json' ) ) {
+				$file_counter++;
+		   	}
+			$variation_slug = $variation_slug . '_' . $file_counter;
+		}
 
+		$_GET['theme']['variation_slug'] = $variation_slug;
+		
 		file_put_contents(
 			$variation_path . $variation_slug . '.json',
 			MY_Theme_JSON_Resolver::export_theme_data( $export_type )
@@ -912,7 +924,7 @@ Tags: one-column, custom-colors, custom-menu, custom-logo, editor-style, feature
 
 	function admin_notice_variation_success() {
 		$theme_name = wp_get_theme()->get( 'Name' );
-		$variation_name = get_stylesheet_directory() . DIRECTORY_SEPARATOR . 'styles' . DIRECTORY_SEPARATOR . sanitize_title( $_GET['theme']['variation'] ) .'.json';
+		$variation_name = get_stylesheet_directory() . DIRECTORY_SEPARATOR . 'styles' . DIRECTORY_SEPARATOR . $_GET['theme']['variation_slug'] .'.json';
 
 		?>
 			<div class="notice notice-success is-dismissible">
--- a/admin/class-create-block-theme-admin.php
+++ b/admin/class-create-block-theme-admin.php
@@ -35,8 +35,7 @@ class Create_Block_Theme_Admin {
 	}
 
 	function save_variation ( $export_type, $theme ) {
-		$this->add_theme_json_variation_to_local( $export_type );
-		$this->export_theme( $theme );
+		$this->add_theme_json_variation_to_local( $export_type, $theme );
 	}
 
 	function clear_user_customizations() {
@@ -337,14 +336,14 @@ class Create_Block_Theme_Admin {
 		);
 	}
 
-	function add_theme_json_variation_to_local ( $export_type ) {
+	function add_theme_json_variation_to_local ( $export_type, $theme ) {
 		$variation_path = get_stylesheet_directory() . DIRECTORY_SEPARATOR . 'styles' . DIRECTORY_SEPARATOR;
 		if ( ! file_exists( $variation_path ) ) {
 			mkdir( $variation_path, 0755, true );
 		}
 
 		file_put_contents(
-			$variation_path . '/variation.json',
+			$variation_path . '/' . $theme['variation'] . '.json',
 			MY_Theme_JSON_Resolver::export_theme_data( $export_type )
 		);
 	}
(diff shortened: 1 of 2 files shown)


mikachan (commented):
This is fab, thanks for working on it!

> Should we ask the user for a variation name?

Yeah, I think this would be good so we could populate the `title` key in the json file. Is there currently a way to see which theme a variation is of just from the json file? I'm thinking if we're making it easy for users to create variations, it would be nice to track which theme it was designed for. (I guess if you download the whole ZIP of the theme, it's obvious which theme it's for..)

> Should the exported ZIP file only include the variation or the whole theme included like in this case?

I think it should only include the variation, and maybe just download the variation file without the ZIP. If we did this, it's probably more important to ask the user for a variation name, and maybe the generated file could be named '[themename]-[variationname].json'.

> Is the wording on the new option clear for what the feature does?

Yeah, I think it makes sense. If we chose to only include the variation file or just download the variation file, we could include this in the text too, e.g. "Generates a style variation JSON file for the current theme. Any template changes will be ignored." It would be cool if the button text changed to 'Create variation' when this option is selected too.

MaggieCabrera:
> I think it should only include the variation, and maybe just download the variation file without the ZIP. If we did this, it's probably more important to ask the user for a variation name, and maybe the generated file could be named '[themename]-[variationname].json'.
> 

I wasn't sure about this because the variation file without the context of the rest of the theme files implies that the user knows what to do with it, and where to put it. If we want to give the user just the .json file, I'd rather give them a zip with folders like so: `themename/styles/variationname.json` so at the very least the structure is clear. 

Still, in that case the zip you are downloading is not useful per se, you still need the rest of the theme files to make it work, and are relying on the user to know what to do to make it work.

madhusudhand:
> Should we ask the user for a variation name?

Yeah, I think it's a good idea to ask the name. It can drive the file name.

> Is the wording on the new option clear for what the feature does?

Current behaviour acts as `copy user changes as style variation`. It should move the user changes to new variation instead, otherwise there are two copies with same changes.

So, similar to the option of saving user changes as theme, this option could be `Save USER changes as new style variation changes and delete the USER changes. Any template changes will be ignored.`
And they should be together for the better context to user on how to save the changes.

> Should the exported ZIP file only include the variation or the whole theme included like in this case?

Should it needs to be exported?
I guess, It is fine to just saves a new variation in the theme. Standalone export of the file is of no use anyways. User can later decide to clone the entire theme using the other option.

MaggieCabrera:
> Current behaviour acts as `copy user changes as style variation`. It should move the user changes to new variation instead, otherwise there are two copies with same changes.
> 
> So, similar to the option of saving user changes as theme, this option could be `Save USER changes as new style variation changes and delete the USER changes. Any template changes will be ignored.` And they should be together for the better context to user on how to save the changes.

Good call! and if the theme already has a variation, should it edit the existing one or create a new one then?


> Should it needs to be exported? I guess, It is fine to just saves a new variation in the theme. Standalone export of the file is of no use anyways. User can later decide to clone the entire theme using the other option.

In the context of the default theme we do want the user to get the file so they can submit it for it to be consider for the release

MaggieCabrera:
I do agree that we already have the export option, and the variation will be locally saved so you can get the whole zip from there, maybe we don't need to download anything at all, just make it explicit on the success message for the user to know where the variation files are and what folder they are in.

mikachan:
> Still, in that case the zip you are downloading is not useful per se, you still need the rest of the theme files to make it work, and are relying on the user to know what to do to make it work.

I understand what you mean, I feel in any case the user will need to know what to do with the file unless we supply the entire theme.

> I do agree that we already have the export option, and the variation will be locally saved so you can get the whole zip from there, maybe we don't need to download anything at all, just make it explicit on the success message for the user to know where the variation files are and what folder they are in.

Yeah, I think not downloading anything at all makes more sense. I'm now struggling to see the benefit of exporting a copy of the theme just for variations, i.e. I'm not sure what type of user would be intending to do that, at that point you probably just want to export the whole theme anyway. I think showing the explicit success message with more info sounds good!

MaggieCabrera:
This is ready for another review:

- I changed the wording
- I added the input for the variation name
- I cleared the user changes after saving the variation json file locally
- I removed the zip download and added a message for the user to let them know where the variation file is going to be

admin/class-create-block-theme-admin.php
```diff
@@ -679,38 +696,50 @@ function create_admin_form_page() {
...
 							<input type="submit" value="<?php _e('Create theme', 'create-block-theme'); ?>" class="button button-primary" />
 
 						</div>
 					</div>
 					<div id="col-right">
 						<div class="col-wrap">
+							<div hidden id="new_variation_metadata_form">
+								<label>
+									<?php _e('Variation Name (*):', 'create-block-theme'); ?><br />
+									<input placeholder="<?php _e('Variation Name', 'create-block-theme'); ?>" type="text" name="theme[variation]" class="large-text" />
```
mikachan, line 753: We could use HTML5 validation on the required fields like theme name and variation name, so the feedback is quicker for the user. Something like this:

```suggestion
									<input placeholder="<?php _e('Variation Name', 'create-block-theme'); ?>" type="text" name="theme[variation]" class="large-text" required />
```

We should still keep the server-side validation too, but with the above, it means the page doesn't need to reload. Although we'd need to remove the `required` attribute from the forms that are hidden, so they're not evaluated if they're not being used... Maybe client-side validation could be handled in a separate PR (it could be added to the new `toggleForm` function).

admin/class-create-block-theme-admin.php
```diff
@@ -331,6 +336,18 @@ function add_theme_json_to_local ( $export_type ) {
...
 	}
 
+	function add_theme_json_variation_to_local ( $export_type, $theme ) {
+		$variation_path = get_stylesheet_directory() . DIRECTORY_SEPARATOR . 'styles' . DIRECTORY_SEPARATOR;
+		if ( ! file_exists( $variation_path ) ) {
+			mkdir( $variation_path, 0755, true );
+		}
+
+		file_put_contents(
+			$variation_path . '/' . $theme['variation'] . '.json',
```
mikachan: Do you think we should replace spaces in the variation name with hyphens? e.g. changing Dark Blue.json to dark-blue.json? I think most file systems will handle spaces anyway, but it's probably best practice to remove them.

admin/class-create-block-theme-admin.php
```diff
@@ -331,6 +336,18 @@ function add_theme_json_to_local ( $export_type ) {
...
+			MY_Theme_JSON_Resolver::export_theme_data( $export_type )
```
mikachan, line 360: Should we add the `title` property to the new json file? It doesn't look like we have to, as the name is pulled from the filename in global styles:

image

But then I'm not sure what the `title` property is for! Maybe so you can format the title differently from the filename? So in this case we could set this to the original variation name that's entered.

mikachan (commented):
This is looking great! I've added some comments, but nothing major.

admin/class-create-block-theme-admin.php
MaggieCabrera: yeah we would need to toggle the attributes, that's why I didn't add it. Looks like a good enhancement for a followup PR

admin/class-create-block-theme-admin.php
madhusudhand: `$variation_path` already ending with `DIRECTORY_SEPARATOR`. extra `/` can be removed.

admin/class-create-block-theme-admin.php
```diff
@@ -679,38 +696,50 @@ function create_admin_form_page() {
...
-								<input value="blank" type="radio" name="theme[type]" class="regular-text code" onchange="document.getElementById('new_theme_metadata_form').removeAttribute('hidden', null);" />
+								<input value="blank" type="radio" name="theme[type]" class="regular-text code" onchange="toggleForm( 'new_theme_metadata_form', false );" />
 								<?php _e('Create blank theme ', 'create-block-theme'); ?><br />
 								<?php _e('[Generates a boilerplate "empty" theme inside of this site\'s themes directory.]', 'create-block-theme'); ?>
 							</label>
 							<br /><br />
+							<label>
+								<input value="variation" type="radio" name="theme[type]" class="regular-text code" onchange="toggleForm( 'new_variation_metadata_form', false );" />
+								<?php _e('Create a style variation ', 'create-block-theme'); ?><br />
+								<?php printf( esc_html__('[Saves user changes as a style variation of %1$s.]', 'create-block-theme'),  esc_html( wp_get_theme()->get('Name') ) ); ?>
```
madhusudhand: Just a minor comment on wording.
All the choices start with singular `Export` `create` `save` the last two options are using plural wording `Generates`, `Saves`.
Make them singular to match the other options.

admin/class-create-block-theme-admin.php
```diff
@@ -852,4 +909,15 @@ function admin_notice_blank_success() {
 		<?php
 	}
 
+	function admin_notice_variation_success() {
+		$theme_name = $_GET['theme']['name'];
+		$variation_name = get_stylesheet_directory() . '/styles/' . $_GET['theme']['variation'] .'.json';
```
madhusudhand: use `DIRECTORY_SEPARATOR` for the right path.

admin/class-create-block-theme-admin.php
```diff
@@ -852,4 +909,15 @@ function admin_notice_blank_success() {
...
+
+		?>
+			<div class="notice notice-success is-dismissible">
+				<p><?php printf( esc_html__( 'Your variation of %1$s has been created succesfully. The new variation file is in %2$s', 'create-block-theme' ), esc_html( $theme_name ) , esc_html( $variation_name )  ); ?></p>
```
madhusudhand: Message is missing theme name for some reason. `Your variation of has been created succesfully.`
There is a small typo `successfully`

madhusudhand (commented):
It is working great! 🏆 
Dropped few comments inline.

madhusudhand:
> if the theme already has a variation, should it edit the existing one or create a new one then?

Current behaviour seem to overrides existing variation. 
I think it is best to create a new one (may be suffix it with a number).

admin/class-create-block-theme-admin.php
MaggieCabrera: Thanks for catching this!

admin/class-create-block-theme-admin.php
MaggieCabrera: I think changing the content of the generated json will bring extra complexity, but sanitizing the slug makes a lot of sense. We can revisit the title of it in a follow up maybe?

MaggieCabrera:
Made a few comments and fixes

scruffian (approved):

mikachan (approved):
LGTM!

PR #92, opened 2022-07-28
Make UI string 'Create Block Theme' can be translatable

alexclassroom:
The shared string 'Create Block Theme' is the plugin name and also is a UI string. This PR makes the UI string split from the plugin name and translatable.


scruffian (approved):
Thanks for taking care of this.

PR #96, opened 2022-07-29
Add style variations to the Site Editor

scruffian:
This adds style variation creation to the Site Editor:

https://user-images.githubusercontent.com/275961/181806075-2aa918a8-88aa-4652-b52c-bce0d95b8620.mp4

This adds a new JS package that allows us to integrate this tool more deeply into the Gutenberg Site Editor.

The initial step is to add the creation of style variations, but in the future we should move across other tools.

To test you need to use this Gutenberg PR: https://github.com/WordPress/gutenberg/pull/42824


scruffian:
I also added blank theme creation to the site editor:

Screenshot 2022-08-01 at 12 11 10

src/index.js
```diff
@@ -0,0 +1,111 @@
...
+					? error.message
+					: __( 'An error occurred while creating the site export.' );
+			createErrorNotice( errorMessage, { type: 'snackbar' } );
+		}
+	}
+
+	return (
+		<PanelBody>
+			<MenuGroup label={ __( 'Create style variation' ) }>
+				<TextControl onChange={ ( value ) => setVariationName( value ) } placeholder={ __( 'Variation name' ) }></TextControl>
```
madhusudhand: May be a short description can be added to give the context of the feature.

`Save user changes as a new style variation of the current theme`

madhusudhand (commented):
Plugin is crashing on click of the button.

image

image

src/index.js
MaggieCabrera: I wonder if there's a tooltip component or something like that in GB so we don't add a lot fo text while still conveying all the information for the action

pbking: Crashing for me too.

scruffian:
@madhusudhand @pbking sorry I forgot to mention that you also need this associated Gutenberg PR: https://github.com/WordPress/gutenberg/pull/42824

src/index.js
```diff
@@ -0,0 +1,111 @@
...
+import apiFetch from '@wordpress/api-fetch';
+import { Button, MenuGroup, MenuItem, PanelBody, TextControl } from '@wordpress/components';
+import { store as noticesStore } from '@wordpress/notices';
+import { useDispatch } from '@wordpress/data';
+
+const GlobalStylesProvider = wp.editSite.GlobalStylesProvider;
+
+const StyleVariations = () => {
+	const [ variationName, setVariationName ] = useState( '' );
+	const [ canReset, onReset ] = wp.editSite.useGlobalStylesReset();
```
jffng: It's nice for demo purposes but I'm wondering from a UX perspective, does it make sense to reset the styles when the variation is created?

src/index.js
```diff
@@ -0,0 +1,111 @@
...
+				error.message && error.code !== 'unknown_error'
+					? error.message
+					: __( 'An error occurred while creating the site export.' );
+			createErrorNotice( errorMessage, { type: 'snackbar' } );
+		}
+	}
+
+	return (
+		<PanelBody>
+			<MenuGroup label={ __( 'Create style variation' ) }>
```
jffng: There's a padding applied to this label that I think looks off, I think it should be just left aligned, if possible. Maybe that's a problem with the component though.

Screen Shot 2022-08-09 at 2 25 04 PM

jffng (commented):
This worked well for me in combination with the https://github.com/WordPress/gutenberg/pull/42824.

If Gutenberg is deactivated, the editor crashes when I click on the Create Block Theme panel, so perhaps a check should be added for whether to initialize this panel.

src/index.js
scruffian: That's what we do in the PHP side of the tool.

pbking: Ideally when in this perspective we could activate the variation (which would leave all of the styles as-set instead of resetting to default).

pbking:
> If Gutenberg is deactivated, the editor crashes when I click on the Create Block Theme panel, so perhaps a check should be added for whether to initialize this panel.

Yeah, definitely either need a check for capabilities or wait for that change to land in CORE to release this.

scruffian:
> Yeah, definitely either need a check for capabilities or wait for that change to land in CORE to release this.

Done in dfbc74f.

src/index.js
scruffian: Yeah I'd rather fix that in the component itself.

jffng (approved):
Would like to see this one land since it makes designing a variation in the site editor so much easier, especially if you are switching around between versions / branches a lot.

jffng:
Ah just as I green checked, I noticed one bug — when I try to save a variation with a space in its name, it 404s:

Screen Shot 2022-08-11 at 12 29 58 PM

vcanales:
It's been a while, so I took the liberty to bring this up to date removing some code that was breaking due to changes in the App. I also refactored it to reflect how other parts of the Plugin behave within the Site Editor. If @scruffian has no objections, I'd be happy to dedicate a bit of time to this.

includes/class-create-block-theme-api.php
```diff
@@ -105,7 +126,7 @@ function rest_get_readme_data( $request ) {
 			return new WP_REST_Response(
 				array(
 					'status'  => 'SUCCESS',
-					'message' => __( 'Readme file data retrieved.', 'create-block-theme' ),
+					'message' => __( 'Readme file data retrieved . ', 'create-block-theme' ),
```
vcanales: It seems like an auto-formatter was a bit overzealous with its job. Will be correcting these.

pbking: 🤣

pbking:
I would love to see this wrap up and land.  It was the inspiration and reference for other site editor changes but that never included variations!  Would be keen to review or assist on wrapping this up.

scruffian:
I'm happy for anyone to take this over.

madhusudhand:
closing this in favour of #532

PR #99, opened 2022-08-05
Add $schema and use Gutenberg classes

### LINKED ISSUE #81: Add the `$schema` key to theme.json
When creating a theme, the generated theme.json does not include the `$schema` key, and so doesn't automatically format the file.

We should add the following to the start of the generated theme.json:

```
"$schema": "https://schemas.wp.org/trunk/theme.json"
```

scruffian:
This adds the $schema key when overwriting and exporting themes. It also uses the WP_Theme_JSON_Gutenberg class if it's available, so that the latest features of the plugin are included.

To test, try exporting/overwriting a block theme and check that the $schema key still exists. Also check that button elements aren't wiped. You can see an example here: https://github.com/Automattic/themes/pull/5708

Fixes #81

### DIFF
--- a/admin/resolver_additions.php
+++ b/admin/resolver_additions.php
@@ -29,7 +29,13 @@ function augment_resolver_with_utilities() {
 				// Get parent theme.json.
 				$parent_theme_json_data = static::read_json_file( static::get_file_path_from_theme( 'theme.json', true ) );
 				$parent_theme_json_data = static::translate( $parent_theme_json_data, wp_get_theme()->parent()->get( 'TextDomain' ) );
+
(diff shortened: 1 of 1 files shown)


MaggieCabrera:
This works as expected in every case except when trying to create a child theme. It generates a zip that won't open (it works on trunk, but fails on this branch)

I'm seeing this:

Screenshot 2022-08-05 at 16 56 53

scruffian:
Thanks @MaggieCabrera, I added a commit to fix that...

MaggieCabrera (approved):
Lovely, it works as intended! Thank you for this

PR #105, opened 2022-08-24
Generate $schema URL in the same way as core.

### LINKED ISSUE #104: $schema URI should include WP Version
When the core theme export runs the $schema URI checks for Gutenberg to determine if the trunk version of the schema should be used, if Gutenberg is not present then the URI for the currently installed version of WordPress should be used instead.

DaisyOlsen:
My attempt at matching the way the $schema URI is generated to match what WordPress core does.

Note this does not adjust the URI included in the boiler plate theme.json
Fixes #104

### DIFF
--- a/admin/resolver_additions.php
+++ b/admin/resolver_additions.php
@@ -67,7 +67,12 @@ function augment_resolver_with_utilities() {
 
 			// Add the schema.
 			if ( empty( $schema ) ) {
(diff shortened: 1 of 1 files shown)


pbking (approved):
Checks out. 👍 

We should do the blank theme in another change as well.

PR #107, opened 2022-08-30
Change theme.json schema of blank theme if Gutenberg isn't installed.

pbking:
Follows up https://github.com/WordPress/create-block-theme/pull/105

### DIFF
--- a/admin/class-create-block-theme-admin.php
+++ b/admin/class-create-block-theme-admin.php
@@ -317,6 +317,16 @@ class Create_Block_Theme_Admin {
 					copy($item, $blank_theme_path . DIRECTORY_SEPARATOR . $iterator->getSubPathname());
 				}
 			}
+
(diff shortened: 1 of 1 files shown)


DaisyOlsen (approved):
LGTM. Working as expected with 6.0.x with and without Gutenberg active.

Might be worth noting, however, it all breaks down when a development version of WP is installed because the next version's schema doesn't yet exist at the appropriate URL. I don't see this as a blocker at the moment because it's also a problem in core 😁  Treating an alpha, beta, RC the same as Gutenberg would be my first inclination... And, of course this would also affect #105

pbking:
Hrm, I didn't think of that...
It doesn't render the theme broken, just the schema, and if you're running a pre-release version + this plugin there's a good chance you'll be able to work around that situation.  Agree a fix would be good, though low priority.  Thanks for noting it.

PR #110, opened 2022-09-12
Add check and directory creation for template and parts folders.

### LINKED ISSUE #109: Overwriting local theme fails if folder doesn't exist.
This was discovered during @DaisyOlsen WCUS workshop.  We created 3 template parts via the Site Editor and there was no `parts` directory in the theme.  Running the **Overwrite {Theme Name} Theme** command caused a fatal error.

It looks like adding a check for the folder and creating it before trying to write the files would address the issue - https://github.com/WordPress/create-block-theme/blob/trunk/admin/class-create-block-theme-admin.php#L595

# Steps to Reproduce

1. Create a custom theme
2. Do not include a `parts` directory
3. Create a Header and Footer template part via the Site Editor and save.
4. Run the Override functionality.

ryanwelcher:
Fixes #109 by checking for and creating the template or parts folder if they don't exist locally

### DIFF
--- a/admin/class-create-block-theme-admin.php
+++ b/admin/class-create-block-theme-admin.php
@@ -585,11 +585,21 @@ class Create_Block_Theme_Admin {
 		$theme_templates = $this->get_theme_templates( $export_type, null );
 		$template_folders = get_block_theme_folders();
 
+		// If there is no templates folder, create it.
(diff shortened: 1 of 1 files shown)


DaisyOlsen (approved):
I've given this a go and confirmed that the behavior experienced during the Workshop is resolved with this chance.

LGTM!

PR #112, opened 2022-09-23
Change button text

### LINKED ISSUE #111: Consider changing primary form button text
The plugin does so much more than Create Themes now, so it might be a good idea to change the button text to something like "Create", but that doesn't really work for the "Overwrite" option.

Maybe "Generate" is a better option.

jonathanbossenger:
Change the button text to Generate, to cover all current and possible future plugin actions.

Fixes #111


madhusudhand (approved):
LGTM!

PR #119, opened 2022-10-05
Validating mime type of font file on server side

matiasbenedetto:
## What ?
Validating mime type of font file on server side

### DIFF
--- a/admin/class-embed-fonts-in-theme.php
+++ b/admin/class-embed-fonts-in-theme.php
@@ -347,7 +347,7 @@ class Embed_Fonts_In_Theme_Admin {
 		$theme_name = wp_get_theme()->get( 'Name' );
 		?>
 			<div class="notice notice-error is-dismissible">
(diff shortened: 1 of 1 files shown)


admin/class-embed-fonts-in-theme.php
```diff
@@ -324,6 +342,15 @@ function admin_notice_embed_font_permission_error () {
 			</div>
 		<?php
 	}
+
+    function admin_notice_embed_font_file_error () {
+		$theme_name = wp_get_theme()->get( 'Name' );
+		?>
+			<div class="notice notice-error is-dismissible">
+				<p><?php printf( esc_html__( 'Error adding %1$s font to %2$s theme.The font file submitted is not valid.', 'create-block-theme' ), esc_html( $_POST['font-name'] ), esc_html( $theme_name ) ); ?></p>
```
jffng:
```suggestion
				<p><?php printf( esc_html__( 'Error adding %1$s font to %2$s theme. The uploaded file is not valid.', 'create-block-theme' ), esc_html( $_POST['font-name'] ), esc_html( $theme_name ) ); ?></p>
```

jffng (approved):
Tested that bypassing client side verification by altering the `input` does not work. Also verified uploading a valid font file still works.

PR #120, opened 2022-10-05
Check for nonce index 

ramonjd:
With the plugin activated I'm seeing this PHP warning in admin

```
Notice: Undefined index: nonce in /var/www/html/wp-content/plugins/create-block-theme/admin/class-embed-fonts-in-theme.php on line 198 Notice: Undefined index: nonce in /var/www/html/wp-content/plugins/create-block-theme/admin/class-embed-fonts-in-theme.php on line 161 Warning: Cannot modify header information - headers already sent by (output started at /var/www/html/wp-content/plugins/create-block-theme/admin/class-embed-fonts-in-theme.php:198) in /var/www/html/wp-admin/includes/misc.php on line 1416 Warning: Cannot modify header information - headers already sent by (output started at /var/www/html/wp-content/plugins/create-block-theme/admin/class-embed-fonts-in-theme.php:198) in /var/www/html/wp-includes/functions.php on line 6865 Warning: Cannot modify header information - headers already sent by (output started at /var/www/html/wp-content/plugins/create-block-theme/admin/class-embed-fonts-in-theme.php:198) in /var/www/html/wp-includes/pluggable.php on line 1421 Warning: Cannot modify header information - headers already sent by (output started at /var/www/html/wp-content/plugins/create-block-theme/admin/class-embed-fonts-in-theme.php:198) in /var/www/html/wp-includes/pluggable.php on line 1424
```

This PR checks for the nonce before trying to use it in order to prevent undefined index PHP warning.

```php
! empty( $_POST['nonce'] )
```

## Testing

Install the plugin (WP_DEBUG should be `true`). Check that the warning does not appear.

cc @matiasbenedetto

### DIFF
--- a/admin/class-embed-fonts-in-theme.php
+++ b/admin/class-embed-fonts-in-theme.php
@@ -1,7 +1,7 @@
-<?php 
+<?php
 
 class Embed_Fonts_In_Theme_Admin {
-    
+
 	public function __construct() {
         add_action( 'admin_menu', [ $this, 'create_admin_menu' ] );
         add_action( 'admin_init', [ $this, 'save_google_fonts_to_theme' ] );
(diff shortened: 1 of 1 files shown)


admin/class-embed-fonts-in-theme.php
```diff
@@ -1,7 +1,7 @@
-<?php 
+<?php
```
ramonjd, line 1: This and the other formatting changes below are WordPress PHPCS formatting. No code changes.

matiasbenedetto (approved):
Thanks for the fix! LGTM

PR #124, opened 2022-10-10
Prototype: Add custom templates

scruffian:
This is a prototype of how plugins might want to add custom templates to a theme without needing to modify the theme itself. 

This works by attaching to two filters: `get_block_template` and `get_block_templates`. The new custom template is added programatically to the theme using these filters.

This approach could be used by plugins who require themes to have specific templates in order to function, without having to actually add the files to the theme.

One problem with this approach is that it makes some assumptions about the base template of the theme.

### DIFF
--- a/create-block-theme.php
+++ b/create-block-theme.php
@@ -40,18 +40,30 @@ function run_create_block_theme() {
 }
 run_create_block_theme();
 
+function replace_between($str, $needle_start, $needle_end, $replacement) {
+    $pos = strpos($str, $needle_start);
+    $start = $pos === false ? 0 : $pos;
+
+    $pos = strpos($str, $needle_end, $start);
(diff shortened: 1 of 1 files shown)


create-block-theme.php
```diff
@@ -39,3 +39,58 @@ function run_create_block_theme() {
...
+}
+
+function get_custom_template() {
+	$plugin = 'plugin-name';
+	$slug = 'custom-template';
+	$template_id = $plugin . '//' . $slug;
+	$title = 'template-title';
+	$description = 'description';
+	$custom_content = '<!-- wp:woocommerce/legacy-template {"template":"single-product"} /-->';
+	$index_template = wp_get_theme()->get_file_path( '/templates/index.html' );
```
scruffian, line 60: This assumes a block theme

pbking:
Interesting experiment, but no longer relevant to the project.  Closing as "interesting".

PR #126, opened 2022-10-17
Manage theme fonts

matiasbenedetto:
## What?
This PR consolidates the previous work around handling theme fonts using the plugin. 
The new changes:
- Allows previewing theme font families and font faces.
- Allows removing font families.
- Allows removing font faces.
- Removes the "embed google font" and "embed local font" options from the dashboard menu and replaces it for "Manage fonts".

## Why?
Because currently is possible just add fonts but not remove them.

## Screencast
[Screencast from 18-10-22 00:31:11.webm](https://user-images.githubusercontent.com/1310626/196302664-2a55ad6c-f9f7-43c1-8db9-6baf73e0d572.webm)


admin/class-manage-fonts.php
```diff
@@ -53,6 +58,51 @@ function can_read_and_write_font_assets_directory () {
...
+        $fonts_json = wp_json_encode( $theme_font_families, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE );
+        $fonts_json_string = preg_replace ( '~(?:^|\G)\h{4}~m', "\t", $fonts_json );
+        
+    ?>
+    <div class="wrap">
+        <h1 class="wp-heading-inline"><?php _e('Manage Theme Fonts', 'create-block-theme'); ?></h1>
+        <a href="<?php echo admin_url( 'themes.php?page=add-google-font-to-theme-json' ); ?>" class="page-title-action"><?php _e('Add Google Font', 'create-block-theme'); ?></a>
+        <a href="<?php echo admin_url( 'themes.php?page=add-local-font-to-theme-json' ); ?>" class="page-title-action"><?php _e('Add Local Font', 'create-block-theme'); ?></a>
+        <hr class="wp-header-end" />
+        <p><?php printf( esc_html__( 'These are the fonts currently available in your theme (%1$s).', 'create-block-theme' ), esc_html( $theme_name ) ); ?></p>
```
jffng: If the theme itself bundles a custom font, should it display them? For example, I tested with Pixl which bundles DMMono, but I do not see it: 

Screen Shot 2022-10-18 at 12 32 57 PM

jffng (commented):
This is great! :clap:

I loaded a custom google font, and applied it in the site editor > global styles > typography, but it is not rendering: 

Screen Shot 2022-10-18 at 12 34 58 PM

The front-end and post editor are working.

admin/class-manage-fonts.php
matiasbenedetto: Yep, all the fonts present in theme.json should be listed. This was a glitch happening with some themes. It should be solved with the latest changes.

matiasbenedetto:
> I loaded a custom google font, and applied it in the site editor > global styles > typography, but it is not rendering:

I think this was caused by a glitch that was making the fonts not being saved and because of that not loaded in the editor. Please, try again with the latest changes.

admin/class-manage-fonts.php
madhusudhand: I see it working now!

package.json
```diff
@@ -0,0 +1,25 @@
+{
+    "name": "05-recipe-card-esnext",
```
madhusudhand: update the fields to the appropriate values.

madhusudhand (commented):
Great work @matiasbenedetto 🎉  It is much simplified and working great.

madhusudhand:
Can you pick `name` and show it. And if it doesn't exists then fallback to `familyName`

image

package.json
matiasbenedetto: updated!

matiasbenedetto:
> Can you pick name and show it. And if it doesn't exists then fallback to familyName

Done!

src/font-family.js
```diff
@@ -0,0 +1,81 @@
...
+                        </div>
+                    </div>
+                </tbody>
+            )}
+        </table>
+    )
+}
+
+FontFamily.defaultProps = {
+    demoText: __("The quick brown fox jumps over the lazy dog."),
```
jffng:
```suggestion
    demoText: __("The quick brown fox jumps over the lazy dog.", "create-block-theme"),
```

src/font-family.js
```diff
@@ -0,0 +1,81 @@
...
+            {hasFontFaces && (
+                <tbody className="font-family-contents">
+                    <div className="container">
+                        <div className={` slide ${isOpen ? "open" : "close"}`}>
+                            <table className="wp-list-table widefat striped table-view-list">
+                                <thead>
+                                    <td>{__('Style')}</td>
+                                    <td>{__('Weight')}</td>
+                                    <td>{__('Preview')}</td>
+                                    {/* <td>{__('Edit')}</td> */}
```
jffng:
```suggestion
```

src/font-face.js
```diff
@@ -0,0 +1,27 @@
...
+        fontStyle: fontFace.fontStyle,
+        fontWeight: fontWeight,
+    };
+
+    return (
+        <tr className="font-face">
+            <td>{fontFace.fontStyle}</td>
+            <td>{fontFace.fontWeight}</td>
+            <td className="demo-cell"><p style={ demoStyles }>{demoText}</p></td>
+            {/* <td><Button variant="secondary">Edit</Button></td> */}
```
jffng:
```suggestion
```

build/index.asset.php
```diff
@@ -0,0 +1 @@
+<?php return array('dependencies' => array('react', 'wp-components', 'wp-element', 'wp-polyfill'), 'version' => 'adf49c3b45f9ee2c8230');
```
jffng: Now that there is a compiled code for the plugin, is it common to include the build inside the repo, or should that be done another way?

jffng (commented):
Looking good and working well for me, a few nit picks.

build/index.asset.php
matiasbenedetto: Good catch, thanks! I will remove the build files from this PR. We will need to add extra steps in the release process to build the app. I will follow up on that task in another PR.

matiasbenedetto: The follow up PR: https://github.com/WordPress/create-block-theme/pull/131

jffng (approved):
Looking and working great, pre-approving with the message that we should add something to the readme now that the plugin requires a build step for development.

matiasbenedetto:
> Looking and working great, pre-approving with the message that we should add something to the readme now that the plugin requires a build step for development.

Ok, I think we could add a section around contributing/developing of the plugin. I'm adding an issue as a follow-up: https://github.com/WordPress/create-block-theme/issues/132

PR #127, opened 2022-10-21
Force https to load Google fonts preview

matiasbenedetto:
## What?
Force https to load Google fonts preview

## Why?

For some mysterious reason the Google Fonts API gives us the URL address of the fonts assets using http instead of https.

It seems like a Google Fonts API bug, because I don't see why they still use http instead of https when they mark all sites using http as insecure. But, anyway, we can add a quick workaround in the plugin code.

Why this problem didn't appear before? Because I think we were testing it in the local dev env using just http. When it is used in a prod site as now, the pages are loaded with https so the browser avoids mixing contents loaded using https (as the site) and http as the Google font asset files.


jffng (approved):
Strange indeed. The solution makes sense to me.

PR #128, opened 2022-10-21
Add the ability to select/unselect all google font variants

matiasbenedetto:
## What?
Add the ability to select/unselect all google font variants

## Why?
To speed up the selection of font faces variants.

### Note:
I'm submitting this PR with `try/manage-fonts` as the branch base because on that branch we made several changes that would make it easier to merge both branches in this way.

## Screenshot

image


madhusudhand (approved):
This improves the experience. LGTM!

PR #131, opened 2022-10-25
Automate release: build, version bump, changelog, deploy to wp.org

matiasbenedetto:
## What?

Currently to make a release of the plugin we need:
1. Build the Javascript app
2. Check the latest version number released
3. Think which is the next version number
4. Update the version number in readme.txt file
5. Write the changelog in the readme.txt file
6. Update the version number in package.json file
7. Manually create a release in the repo indicating the right version number.

With the new approach we need to:
1. Click the button "Run workflow" to select the type of release you are doing (patch, minor, major). 

That's it. The action will take care of building the JS app, bumping to the right semantic version, writing the changelog and submitting it to wporg.

## How?
This PR adds functionality to the release GitHub action. This action now will do all the steps that are currently a manual process.

## Demo
I tried all this code in [another repo](https://github.com/matiasbenedetto/testing-actions) to avoid polluting the history of this one. 

image

Here you can see it in action:

[Screencast from 25-10-22 08:53:26.webm](https://user-images.githubusercontent.com/1310626/197774048-95f6e52c-6239-4815-93cf-791536c9d7bc.webm)

### DIFF
--- a/update-version-and-changelog.js
+++ b/update-version-and-changelog.js
@@ -30,8 +30,10 @@ async function updateVersion () {
     }
 
     const package = require('./package.json');
-    const currentTag = package.version;
-    const newTag = semver.inc(currentTag, releaseType);
+    const currentVersion = package.version;
+    const newVersion = semver.inc(currentTag, releaseType);
+    const currentTag = `v${currentVersion}`;
+    const newTag = `v${newVersion}`;
 
     if (!semver.valid(currentTag)) {
         console.error(`❎  Error: current tag ( ${ currentTag } ) is not a valid semver version."`);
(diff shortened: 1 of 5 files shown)


update-version-and-changelog.js
```diff
@@ -0,0 +1,84 @@
...
+
+    // update package.json version
+    package.version = newVersion;
+    fs.writeFileSync('./package.json', JSON.stringify(package, null, 2));
+    console.info('✅ Version updated', currentTag, '=>', newTag);
+    
+    // update readme.txt version with the new changelog
+    const readme = fs.readFileSync('./readme.txt', 'utf8');
+    const changelogChanges = changes.all.map(change => `* ${ change.body || change.message }`).join('\n');
+    const newChangelog = `== Changelog ==\n\n= ${ newVersion } =\n\n${ changelogChanges }`;
```
jffng: Should it be two line breaks or one, between the `newVersion` and the changes? 😄 
```suggestion
    const newChangelog = `== Changelog ==\n\n= ${ newVersion } =\n${ changelogChanges }`;
```

update-version-and-changelog.js
```diff
@@ -0,0 +1,84 @@
+const semver = require('semver');
```
jffng, line 1: More of a tidiness comment, but would it make sense to put this file, along with the update-google-fonts script in a `utils` folder or something along those lines?

jffng (approved):
Two minor comments, pending those LGTM!

update-version-and-changelog.js
matiasbenedetto: In the current version, we have only one :)

update-version-and-changelog.js
matiasbenedetto: Yep, good idea, we can follow up that in another PR.

PR #145, opened 2022-11-01
I18N: Some new UI strings are not translatable

### LINKED ISSUE #139: I18N: Some new UI strings are not translatable
The uploaded screenshot shows some UI strings cannot be translated, please make the following UI strings translatable.
- Info
- This is a list of your font families listed in the theme.json file of your theme.
- If your theme.json makes reference to fonts providers other than local they may not be displayed correctly.
- These are the fonts currently embedded in your theme
- Weight
- Preview
- Remove Font Family
- Remove
- This action will delete the font definition and the font file assets from your theme.
- Are you sure you want to delete "style-name" variant of "font-name" from your theme?
- Are you sure you want to delete "font-name" from your theme?

gutenberg l - 20221026 - 132257
gutenberg l - 20221026 - 133423

matiasbenedetto:
## Changes:
- Adding dependency when enqueuing script
- Addding text domain for some translatable strings

Fixes: #139

### DIFF
--- a/admin/class-manage-fonts.php
+++ b/admin/class-manage-fonts.php
@@ -71,6 +71,7 @@ class Manage_Fonts_Admin {
         }
      
         // Load our app.js.
+        array_push( $asset_file['dependencies'], 'wp-i18n' );
         wp_enqueue_script( 'create-block-theme-app', plugins_url( 'build/index.js', __DIR__ ), $asset_file['dependencies'], $asset_file['version'] );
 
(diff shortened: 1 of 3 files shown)


mikachan (commented):
I've just had a quick search to see if there are any other missing text domains.

It looks like there's one missing in `font-face.js` for the "Remove" string [here](https://github.com/WordPress/create-block-theme/blob/trunk/src/font-face.js#L21).

I can't see any others that are missing, so this is looking good to me. It would be great if @alexclassroom could take a look at this too, as I'm not sure how to test the `wp-i18n` change.

pbking (approved):
This seems to cover what's missing, we should bring this in.

PR #148, opened 2022-11-10
Check permission before running functions that need file write permissions

### LINKED ISSUE #147: "Overwrite theme" option fails silently and erases the user config
## What?
If the "Overwrite theme" option fails, (for example, because of file permission problems) there are 2 problems:
- The user configuration is lost 
- The process fails silently saying it was successful.

## Expected
If the process fails:
- The user config (styles, templates, etc) should not be erased
- It should display an error message instead of a success message. 

## Screencast
In this screencast, you will be able to see the success message and the user config removed after a failure to overwrite the theme folder.

https://user-images.githubusercontent.com/1310626/201138974-65a1c8a9-c880-4f80-a078-4613bffda416.mp4

matiasbenedetto:
## what?
Check permission before running functions that need file write permissions

 If WordPress lack permission to write the needed paths an error is displayed.
 

## Screenshot:

 Error message example:
 image
 
 Fixes: #147

### DIFF
--- a/admin/class-create-block-theme-admin.php
+++ b/admin/class-create-block-theme-admin.php
@@ -834,6 +834,11 @@ Tags: one-column, custom-colors, custom-menu, custom-logo, editor-style, feature
 			}
 
 			if ( $_GET['theme']['type'] === 'save' ) {
+				// Avoid running if WordPress dosn't have permission to overwrite the theme folder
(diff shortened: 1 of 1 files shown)


mikachan (approved):
LGTM 👍

PR #153, opened 2022-11-15
Allow previewing system font

### LINKED ISSUE #151: Preview system font in the font manager
Currently, the system fonts are not previewed in the "Manage Fonts" section.
We should preview that font too.

image

jffng:
Fixes #151 . This isn't exactly specific to system fonts, but rather fonts that have a fontFamily definition but no fontFace definition(s).

Using TT3 as an example:

Before | After
--- | ---
Screen Shot 2022-11-15 at 11 31 21 AM | Screen Shot 2022-11-15 at 11 31 11 AM

I'm not sure it makes sense to be able to remove the system font, so I've disabled it for now.

### DIFF
--- a/src/font-face.js
+++ b/src/font-face.js
@@ -2,23 +2,21 @@ import { Button } from '@wordpress/components';
 
 const { __ } = wp.i18n;
 
-function FontFace ( { fontFace, demoText, deleteFontFace } ) {
+function FontFace ( { fontFamily, fontWeight, fontStyle, demoText, deleteFontFace } ) {
 
     // Handle cases like fontWeight is a number instead of a string or when the fontweight is a 'range', a string like "800 900".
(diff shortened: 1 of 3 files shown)


matiasbenedetto (approved):
Nice job!

I added a few commits to:
- Fix a regression consisting of the lacking of the remove option for font families without font faces defined.

- Allow slide-up/down of the preview of font families without font faces defined.

- Added normal as default for font weight and font-variant if that info is missing.

- Fix a pre-existing problem consisting of an index error iterating font families without font faces. 
(You can reproduce it, for example, by removing a font face from TT3)

| Before the latest commits | After |
| --- | --- |
| image  | image |

jffng:
Nice @matiasbenedetto thanks for those improvements and bug fix! I didn't realize about the index error.

PR #166, opened 2022-12-18
Replace "current theme" with "active theme" (or "currently active theme")

audrasjb:
This brings consistency with WordPress Core.
See https://core.trac.wordpress.org/ticket/54831


mikachan (approved):
Thanks for this, looks good to me!

PR #173, opened 2023-01-19
Specify node and npm versions, add nvmrc file

vcanales:
Tiny experience upgrade; I keep running into the following when installing dependencies:

```
npm WARN read-shrinkwrap This version of npm is compatible with lockfileVersion@1, but package-lock.json was generated for lockfileVersion@2. I'll try to do my best with it!
```

Specifying our expected minimum versions and adding an `.nvmrc` file that installs them — at least for `nvm` users — improves this.


package.json
```diff
@@ -21,6 +21,10 @@
     "semver": "^7.3.8",
     "simple-git": "^3.14.1"
   },
+  "engines": {
+    "node": ">=16.9.1",
+    "npm": ">=7 <8"
```
vcanales, line 26: Settings this to `<8` because using `npm@8` would overwrite the current `v2` file with a `v3` one.

mikachan (approved):
LGTM!

jffng:
This works as advertised, but do you get any package-lock differences after running `nvm use && npm i`?

vcanales:
> This works as advertised, but do you get any package-lock differences after running `nvm use && npm i`?

It does, although it keeps it as a `v2` lock file. The new lock file is fixing the package name and other inconsistencies unrelated to the npm version, except for adding the engine block. I've pushed the updated file.

image

jffng (approved):
Thanks @vcanales !

PR #176, opened 2023-01-23
Fonts: remove font files from theme assets folder if the font face/family is removed.

### LINKED ISSUE #164: Remove Font Family option does not remove fonts from /assets/
When you click Remove Font Family (or Remove for individual), the fonts are not removed from the /assets/ directory. While they are no longer listed on the Manage Theme Fonts screen, this is not expected behavior. Imagine a user adding/removing 100 fonts, and having those files accumulate in the /assets/ directory.

matiasbenedetto:
## What?
This PR allows the plugin to remove the font file assets from the theme folder when a font face or family is removed from the theme using the plugin font management tool.

## How?
The font face/family selected to be removed is marked with the `shouldBeRemoved: true` property in the frontend and sent to the backend. The backend code understands that it needs to remove the font face/family definition from the JSON and also removes the font files from the theme folder.

## Why? 
Fixes: https://github.com/WordPress/create-block-theme/issues/164

### DIFF
--- a/admin/class-manage-fonts.php
+++ b/admin/class-manage-fonts.php
@@ -229,6 +229,51 @@ class Manage_Fonts_Admin {
 	<?php
 	}
 
+    function delete_font_asset ( $font_face ) {
+        // if the font asset is a theme asset, delete it
+        $theme_folder = get_stylesheet_directory();
+        $font_path = pathinfo( $font_face['src'][0] );
+        $font_dir = str_replace("file:./", "", $font_path['dirname']);
+
+        $font_asset_path = $theme_folder . DIRECTORY_SEPARATOR . $font_dir . DIRECTORY_SEPARATOR . $font_path['basename'];
+
+        if ( ! is_writable( $theme_folder . DIRECTORY_SEPARATOR . $font_dir ) ) {
+            return add_action( 'admin_notices', [ $this, 'admin_notice_font_asset_removal_error' ] );
+
+            if ( file_exists( $font_asset_path ) ) {
+                return unlink( $font_asset_path );
+            }
+        }
+        
+        return false;
+	}
+
+    protected function prepare_font_families_for_database( $font_families ) {
+		$prepared_font_families = array();
+
+		foreach ( $font_families as $font_family ) {
+			if ( isset ( $font_family['fontFace'] ) ) {
+				$new_font_faces = array();
+				foreach ( $font_family['fontFace'] as $font_face ) {
+					$updated_font_face = $font_face;
+					if ( !isset ( $font_face['shouldBeRemoved'] ) ) {
+						$new_font_faces[] = $updated_font_face;
+					} else {
+						$this->delete_font_asset( $font_face );
+					}
+				}
+
+				$font_family['fontFace'] = $new_font_faces;
+			}
+			if ( ! isset ( $font_family[ 'shouldBeRemoved' ] ) ) {
+				$prepared_font_families[] = $font_family;
+			}
+			
+		}
+
+		return $prepared_font_families;
+	}
+
     function save_manage_fonts_changes () {
         if (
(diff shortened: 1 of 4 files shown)


admin/class-manage-fonts.php
```diff
@@ -229,6 +229,51 @@ function google_fonts_admin_page() {
...
+        $theme_folder = get_stylesheet_directory();
+        $font_path = pathinfo( $font_face['src'][0] );
+        $font_dir = str_replace("file:./", "", $font_path['dirname']);
+
+        $font_asset_path = $theme_folder . DIRECTORY_SEPARATOR . $font_dir . DIRECTORY_SEPARATOR . $font_path['basename'];
+
+        if ( ! is_writable( $theme_folder . DIRECTORY_SEPARATOR . $font_dir ) ) {
+            return add_action( 'admin_notices', [ $this, 'admin_notice_font_asset_removal_error' ] );
+
+            if ( file_exists( $font_asset_path ) ) {
```
jffng: I think this needs to be moved outside of the current `if` statement.

admin/class-manage-fonts.php
```diff
@@ -477,6 +523,15 @@ function admin_notice_embed_font_file_error () {
 		<?php
 	}
 
+    function admin_notice_font_asset_removal_error () {
+		$theme_name = wp_get_theme()->get( 'Name' );
+		?>
+			<div class="notice notice-error is-dismissible">
+				<p><?php printf( esc_html__( 'Error removing font asset. WordPress lack permissions to remove these font assets.', 'create-block-theme' ), esc_html( $theme_name ) ); ?></p>
```
jffng:
```suggestion
				<p><?php printf( esc_html__( 'Error removing font asset. WordPress lacks permissions to remove these font assets.', 'create-block-theme' ), esc_html( $theme_name ) ); ?></p>
```

jffng: Would it always be a permissions issue? Should we make the error statement less specific?

jffng (commented):
Hello this mostly works as advertised, except it does not actually remove the font file. I think there is a small bug with the nesting of the unlinking, if I am reading the code correctly.

admin/class-manage-fonts.php
matiasbenedetto: This is an error message specific to this problem. There are other messages for other problems. Maybe we can do the error message handling more generic to reduce the number of them but I think we can handle that in another PR.

admin/class-manage-fonts.php
matiasbenedetto: Thanks for catching this! It's fixed now.

admin/class-manage-fonts.php
```diff
@@ -229,6 +229,51 @@ function google_fonts_admin_page() {
...
+					if ( !isset ( $font_face['shouldBeRemoved'] ) ) {
+						$new_font_faces[] = $updated_font_face;
+					} else {
+						$this->delete_font_asset( $font_face );
+					}
+				}
+
+				$font_family['fontFace'] = $new_font_faces;
+			}
+			if ( ! isset ( $font_family[ 'shouldBeRemoved' ] ) ) {
```
jffng, line 268: Should this also contain the `else` logic to remove all of the fontFaces within the family?

jffng (commented):
It works for removing individual fontFaces, but when I try to remove the whole font family, the font files still persist.

admin/class-manage-fonts.php
matiasbenedetto: Thanks for the catch! I added the condition but on a previous `if` , to reuse the delete_font_asset() call. https://github.com/WordPress/create-block-theme/pull/176/commits/8692b1f12e6e3b2319805170a43089157f22c72b

jffng (approved):

PR #178, opened 2023-01-23
Refactor to read raw theme.json data instead of using core methods

### LINKED ISSUE #170: Adding fonts removes schema (and appearanceTools and anything else?) from theme.json
Hi.

Here's what I do. 

Create a blank theme. Edit: Then, of course, switch to that theme.

Then I add a font.

Then I look at the theme.json, and the schema and "settings": "appearanceTools" parts are now missing. They were there before I added a font.

jffng:
**What**
This PR fixes #170

**How** 
Refactors the `Manage_Fonts_Admin` class to read the active theme.json directly and manipulate it, so that no data (e.g. `appearanceTools` and `$schema` are lost when we add, update, and remove fonts.  

**Why**
Previously we used core's theme.json class methods. However those methods sanitize and transform certain aspects of the theme.json (e.g. appearanceTools & $schema). So it is not 1:1 from the raw file and the class / data that is created.

The intent with this PR is to preserve the theme.json 1:1, except for the specific changes we are making to fontFamilies.

**To Test**
- Create a blank theme
- Add and remove some fonts
- Verify it works as expected

### DIFF
--- a/admin/class-manage-fonts.php
+++ b/admin/class-manage-fonts.php
@@ -394,7 +394,7 @@ class Manage_Fonts_Admin {
 
     function replace_all_theme_font_families ( $font_families ) {
         // Construct updated theme.json.
-        $theme_json_raw = json_decode( file_get_contents( get_template_directory() . '/theme.json' ), true );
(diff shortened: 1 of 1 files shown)


matiasbenedetto (approved):
LGTM :rocket:

PR #179, opened 2023-01-23
Manage fonts: Demo text editable

matiasbenedetto:
This PR makes the demo text editable.

:warning: This functionality is not yet available for the "Add Google font" section because that section was implemented before the React app. We need to add this functionality there, too, but I think we should refactor that section in React before. 

## Screencast

[Screencast from 23-01-23 18:05:03.webm](https://user-images.githubusercontent.com/1310626/214058309-16d8cf37-a6ba-4d28-af8d-8c829195abb0.webm)

### DIFF
--- a/src/fonts-context.js
+++ b/src/fonts-context.js
@@ -1,0 +1,25 @@
+import { useState, createContext } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
+
+export const ManageFontsContext = createContext();
+export const DEFAULT_DEMO_TEXT = __( "The quick brown fox jumps over the lazy dog.", "create-block-theme" );
+
+
+export function ManageFontsProvider( { children } ) {
+    const [ demoText, setDemoText ] = useState( localStorage.getItem( "create-block-theme_default-demo-text" ) || DEFAULT_DEMO_TEXT );
+
+    const handleDemoTextChange = ( newDemoText ) => {
+        setDemoText( newDemoText );
+        localStorage.setItem( "create-block-theme_default-demo-text", newDemoText );
+    }
+
+    const resetDemoText = () => {
+        setDemoText( DEFAULT_DEMO_TEXT );
+    }
+
+    return (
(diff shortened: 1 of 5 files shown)


jffng:
> I think we should refactor that section in React before.

Let's do it. Is it possible to refactor the whole thing in React? The main app screen containing the form is rough and I think it could benefit from some components and general tidying.

matiasbenedetto:
Yep, I started that, and it's mostly working, but due to being a low-impact task, I'm prioritizing other things as the template images bundler.

matiasbenedetto:
If this is working as expected it would be nice to merge it because it contains some elements useful for this refactor: https://github.com/WordPress/create-block-theme/pull/194

src/fonts-context.js
```diff
@@ -0,0 +1,25 @@
...
+export const ManageFontsContext = createContext();
+export const DEFAULT_DEMO_TEXT = __( "The quick brown fox jumps over the lazy dog.", "create-block-theme" );
+
+
+export function ManageFontsProvider( { children } ) {
+    const [ demoText, setDemoText ] = useState( localStorage.getItem( "create-block-theme_default-demo-text" ) || DEFAULT_DEMO_TEXT );
+
+    const handleDemoTextChange = ( newDemoText ) => {
+        setDemoText( newDemoText );
+        localStorage.setItem( "create-block-theme_default-demo-text", newDemoText );
```
jffng, line 13: Is the benefit of using localStorage that the custom text will be preserved in the browser when the user closes + navigates away from this page?

jffng (approved):
Just left one question, but this works as expected so I'm giving the approval. Thanks for the change!

Once it works for Google Fonts too, it might be better to have an explicit input area for the demo text at the top or something, as right now it feels a little awkward to click on one of the text fields and have everything update.

src/fonts-context.js
matiasbenedetto: Yes, and also this is not a pure Single Page Application so we need to store the data somewhere when the page reloads.

PR #185, opened 2023-01-25
Add theme screenshot uploading

### LINKED ISSUE #74: Add Theme Screenshot Upload to Export Form
It would be helpful if the theme creation options allowed users to upload a unique screenshot that would be packaged with the theme download. This will make the newly created themes look much nicer when they are shared and installed on another WordPress site and make them easier to distinguish from other themes that might have been generated this way.

jffng:
Fixes #74 

To test, try out adding a screenshot and generating a theme with the Create child, Clone, and Blank theme options.

### DIFF
--- a/admin/class-create-block-theme-admin.php
+++ b/admin/class-create-block-theme-admin.php
@@ -189,7 +189,6 @@ class Create_Block_Theme_Admin {
 			'style.css',
 			$css_contents
 		);
-		die(var_dump($screenshot));
 
 		if ( is_uploaded_file( $screenshot['tmp_name'] && $screenshot['type'] === 'image/png' ) ){
 			// Add user uploaded screenshot.png.
--- a/admin/class-create-block-theme-admin.php
+++ b/admin/class-create-block-theme-admin.php
@@ -190,7 +190,7 @@ class Create_Block_Theme_Admin {
 			$css_contents
 		);
 
-		if ( is_uploaded_file( $screenshot['tmp_name'] && $screenshot['type'] === 'image/png' ) ){
+		if ( is_uploaded_file( $screenshot['tmp_name'] ) && $screenshot['type'] === 'image/png' ) {
 			// Add user uploaded screenshot.png.
 			$zip->addFile(
 				$screenshot['tmp_name'],
--- a/admin/class-create-block-theme-admin.php
+++ b/admin/class-create-block-theme-admin.php
@@ -189,8 +189,9 @@ class Create_Block_Theme_Admin {
 			'style.css',
 			$css_contents
 		);
+		die(var_dump($screenshot));
 
-		if ( is_uploaded_file( $screenshot['tmp_name'] ) ){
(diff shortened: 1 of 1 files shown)


admin/class-create-block-theme-admin.php
```diff
@@ -135,9 +135,14 @@ function create_sibling_theme( $theme ) {
 			$css_contents
 		);
 
-		// Add screenshot.png.
+		// Add screenshot.
+		$screenshot_path = __DIR__ . '/../screenshot.png';
+		if ( is_uploaded_file( $screenshot['tmp_name'] ) && $screenshot['type'] === 'image/png' ) {
```
mikachan: We're checking the type of the file here is an image/png, but on the upload form, we say it can be a .png, .jpg, or .jpeg. I'm guessing we should probably test if the image is png or jpg/jpeg?

admin/class-create-block-theme-admin.php
```diff
@@ -814,7 +834,13 @@ function create_admin_form_page() {
 									<?php _e('Author URI:', 'create-block-theme'); ?><br />
 									<small><?php _e('The URL of the authoring individual or organization.', 'create-block-theme'); ?></small><br />
 									<input placeholder="<?php _e('https://wordpress.org/', 'create-block-theme'); ?>" type="text" name="theme[author_uri]" class="large-text code" />
-								</label><br />
+								</label><br /><br />
+								<label for="screenshot">
+									<?php _e('Screenshot:', 'create-block-theme'); ?><br />
+									<small><?php _e('Upload a new theme screenshot (2mb max | .png,.jpg,.jpeg allowed | 1200x900 recommended)', 'create-block-theme'); ?></small><br />
+									<input type="hidden" name="MAX_FILE_SIZE" value="20000000" />
+									<input type="file" accept=".png"  name="screenshot" id="screenshot" class="upload"/>
```
mikachan, line 835: We're only accepting .png here too.

mikachan (commented):
This is fab!

I've left a couple of comments. I also wanted to check if we should be validating the image files server-side as well as the HTML form validation. Things such as the file size, and if there's a way to check if the file is actually an image, maybe by checking the `mimetype` as well as the `type`.

admin/class-create-block-theme-admin.php
jffng: I modified this to only accept png, since that seems to be the standard for screenshots. What do you think?

jffng:
I refactored this and I think it's ready for another review. 

One thing is that it fails silently if the screenshot is not valid or too big, maybe we should add in some error handling as a follow up?

admin/class-create-block-theme-admin.php
mikachan: Yeah I think that makes sense, .png definitely seems to be the standard for screenshots.

mikachan (approved):
Nice, I really like the `is_valid_screenshot` function! This is testing well for me.

> One thing is that it fails silently if the screenshot is not valid or too big, maybe we should add in some error handling as a follow up?

Yep, this sounds like a good thing to follow up on separately.

PR #165, opened 2022-12-16
Improve translation process by removing trailing spaces

audrasjb:
Hey there, thanks for this amazing plugin!
This changeset makes it easier for polyglots to translate the plugin :)

PR #188, opened 2023-01-25
Add missing spaces to option labels

mikachan:
Quick PR to add missing spaces to some of the form labels. It looks like some were accidentally removed in https://github.com/WordPress/create-block-theme/pull/165.

### DIFF
--- a/admin/class-create-block-theme-admin.php
+++ b/admin/class-create-block-theme-admin.php
@@ -716,7 +716,14 @@ Tags: one-column, custom-colors, custom-menu, custom-logo, editor-style, feature
 
 							<label>
 								<input checked value="export" type="radio" name="theme[type]" class="regular-text code" onchange="toggleForm( 'new_theme_metadata_form', true );toggleForm( 'new_variation_metadata_form', true );" />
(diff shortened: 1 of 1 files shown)


matiasbenedetto (commented):
Could it be better to provide context to the translators with the use of `_x()` instead of `__()` and the code comment for translators? It seems like `_x()` was created for that: https://developer.wordpress.org/reference/functions/_x/

mikachan:
> Could it be better to provide context to the translators with the use of _x() instead of __() and the code comment for translators? It seems like _x() was created for that:

Yeah, _x() may work better for all these translations. Maybe it would be better to update all translations using this function in a separate PR?

richtabor (approved):

matiasbenedetto:
> > Could it be better to provide context to the translators with the use of _x() instead of __() and the code comment for translators? It seems like _x() was created for that:
> 
> Yeah, _x() may work better for all these translations. Maybe it would be better to update all translations using this function in a separate PR?

Sounds good!

matiasbenedetto (approved):

PR #196, opened 2023-01-30
Add: code linting scripts

mikachan:
This adds a couple of scripts and dependencies to find linting errors following the WordPress coding standards:

- `lint:css` & `lint:css:fix`
- `lint:js` & `lint:js:fix`
- `lint:php` & `lint:php:fix`

It also adds a pre-commit hook to run the linter on staged files before they're commited.

The setup is based on the tools in Gutenberg.

PR #200, opened 2023-02-01
Fix: Adding or removing fonts fails in some Windows environments

t-hamano:
This PR solves the problem of not being able to add or remove fonts in a Windows environment.

https://user-images.githubusercontent.com/54422211/216050896-f7335d09-aa1a-4810-a085-e87cb61860a3.mp4

The PHP core function `is_writable()` may return invalid values in Windows environments, as mentioned in [this Trac ticket](https://core.trac.wordpress.org/ticket/22900). This causes an error that the permissions are incorrect in some Windows environments, when adding or removing fonts.

I have replaced this function with `wp_is_writable()`, an OS-aware WordPress function.

The removal of spaces was done automatically by my code editor, so don't worry about it 🙂

### DIFF
--- a/admin/class-create-block-theme-admin.php
+++ b/admin/class-create-block-theme-admin.php
@@ -240,7 +240,7 @@ class Create_Block_Theme_Admin {
 			'style.css',
 			$this->build_child_style_css( $theme )
 		);
-		
+
 		// Add / replace screenshot.
 		if ( $this->is_valid_screenshot( $screenshot ) ){
 			$zip->addFile(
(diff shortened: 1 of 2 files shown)


matiasbenedetto (approved):
Good catch @t-hamano! thanks for submitting this PR.

I'm not currently able to test it on a Windows machine but [as per function docs](https://developer.wordpress.org/reference/functions/wp_is_writable/)  it should continue working without changes on all the other platforms. I can confirm that works as expected on Linux.

I removed the not needed code formatting from the PR in the last commmit. I know the current formatting of this repo is not ideal, but soon it will be much better: https://github.com/WordPress/create-block-theme/pull/196

LGTM :rocket:

PR #202, opened 2023-02-02
Adding demo text settings

matiasbenedetto:
## What ?
This PR adds demo text settings to be able to preview the fonts more flexibly.

## Extra info
The branch of this Pr was detached from https://github.com/WordPress/create-block-theme/pull/194 so in the diff will appear more changes than the strictly necessary to implement these changes until that#194 is merged.

## Screencast
https://user-images.githubusercontent.com/1310626/216234715-68ce33f0-3e70-4a90-8277-a904b4126aae.mp4

PR #206, opened 2023-02-02
Clean version of Adding demo text settings

jffng:
Not sure what was going on exactly with #202 but I just made a different PR to see the actual diff. 

It looks good to me @matiasbenedetto! I made a couple minor adjustment in de75563f03e99de4f957c0f7be702c44c7b08b28:

- The initial value was not working if it came from local storage, I think because it was parsed as a string.
- Added a number input field for the font size, I feel this is a more common use case to specify a specific value when picking a font size, rather than using the slider.
- A little css


matiasbenedetto (approved):
Looking good! :rocket:

PR #208, opened 2023-02-03
check for DISALLOW_FILE_EDIT and simplify permission check logic

### LINKED ISSUE #203: Plugin doesn't notify users when can't edit files due to a setting defined in wp.config.php
When `define( 'DISALLOW_FILE_EDIT', true )` is defined in the `wp-config.php` file, the result of the call `current_user_can( 'edit_theme' )` will be `false,` so the user is not allowed to edit the theme. The plugin works okay because it doesn't edit the theme if that setting is enabled. The problem is that the plugin doesn't tell anything to the user about this situation, so the user doesn't know what is happening and how to fix it.

We need to implement a notice that warns the users about this setting.

Kudos to [Linsoo](https://wordpress.org/support/users/linsoo/) for debugging this [here](https://wordpress.org/support/topic/this-is-a-comment-about-the-add-local-fonts-to-your-theme-feature-not-working/) and [here](https://linsoo.pe.kr/create-block-theme-%ed%94%8c%eb%9f%ac%ea%b7%b8%ec%9d%b8-manage-theme-fonts-%ec%9e%91%eb%8f%99-%ec%95%88%ed%95%98%eb%8a%94-%eb%ac%b8%ec%a0%9c-%ec%88%98%ec%a0%95%ed%95%98%ea%b8%b0/). That's very much appreciated, thanks [Linsoo](https://wordpress.org/support/users/linsoo/)!

matiasbenedetto:
This PR adds:
- Checking if the user has DISALLOW_FILE_EDIT enabled in their wp-config. More context on: #203 
- Simplify a little the permissions check

Fixes: #203

### DIFF
--- a/admin/class-manage-fonts.php
+++ b/admin/class-manage-fonts.php
@@ -514,7 +514,7 @@ class Manage_Fonts_Admin {
     function admin_notice_user_cant_edit_theme () {
 		?>
 			<div class="notice notice-error is-dismissible">
-				<p><?php printf( esc_html__( 'Error: Your user account do not have the enough permissions to edit the theme.', 'create-block-theme' ) ); ?></p>
+				<p><?php printf( esc_html__( 'Error: You do not have sufficient permission to edit the theme.', 'create-block-theme' ) ); ?></p>
 			</div>
 		<?php
 	}
--- a/admin/class-manage-fonts.php
+++ b/admin/class-manage-fonts.php
@@ -40,6 +40,28 @@ class Manage_Fonts_Admin {
 		add_submenu_page(null, $local_fonts_page_title, $local_fonts_menu_title, 'edit_theme_options', 'add-local-font-to-theme-json', [ $this, 'local_fonts_admin_page' ] );
 	}
 
+    function has_permissions () {
+        $has_user_permissions = $this->user_can_edit_themes();
+        $has_file_permissions = $this->can_read_and_write_font_assets_directory();
+        if ( $has_user_permissions && $has_file_permissions ) {
+            return true;
+        }
+        return false;
+    }
+
+    function user_can_edit_themes () {
+        if ( defined( 'DISALLOW_FILE_EDIT' ) && DISALLOW_FILE_EDIT === true ) {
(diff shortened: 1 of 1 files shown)


admin/class-manage-fonts.php
```diff
@@ -494,6 +502,24 @@ function admin_notice_delete_font_success () {
...
+			<div class="notice notice-error is-dismissible">
+				<p><?php printf( esc_html__( 'Error: You need to disable the DISALLOW_FILE_EDIT config your wp-config.php file to be able to modify your theme', 'create-block-theme' ) ); ?></p>
+			</div>
+		<?php
+	}
+
+    function admin_notice_user_cant_edit_theme () {
+		?>
+			<div class="notice notice-error is-dismissible">
+				<p><?php printf( esc_html__( 'Error: Your user account do not have the enough permissions to edit the theme.', 'create-block-theme' ) ); ?></p>
```
jffng:
```suggestion
				<p><?php printf( esc_html__( 'Error: You do not have sufficient permission to edit the theme.', 'create-block-theme' ) ); ?></p>
```

admin/class-manage-fonts.php
```diff
@@ -494,6 +502,24 @@ function admin_notice_delete_font_success () {
 			</div>
 		<?php
 	}
+
+    function admin_notice_file_edit_error () {
+		?>
+			<div class="notice notice-error is-dismissible">
+				<p><?php printf( esc_html__( 'Error: You need to disable the DISALLOW_FILE_EDIT config your wp-config.php file to be able to modify your theme', 'create-block-theme' ) ); ?></p>
```
jffng:
```suggestion
				<p><?php printf( esc_html__( 'Error: `DISALLOW_FILE_EDIT` cannot be enabled in wp-config.php to make modifications to the theme using this plugin.', 'create-block-theme' ) ); ?></p>
```

jffng (approved):
Pre-approving because it works as expected, left just two suggestions on the wording of the error messages.

admin/class-manage-fonts.php
matiasbenedetto: Thanks!

PR #209, opened 2023-02-03
Load google fonts data from url

### LINKED ISSUE #205: Evaluate if and how to implement code splitting for the React App
More context on: https://github.com/WordPress/create-block-theme/pull/194#pullrequestreview-1279492755

t-hamano:
Closes: #205

This PR obtains Google Web font data by fetching URLs.

The current implementation imports font data directly in JavaScript. This results in slow builds and large build sizes. To solve these problems, this PR gets the font when rendering the React app and treats it as a state. Since the React app does not know at which URL the font data resides, it uses `wp_localize_script()` and passes the asset URL generated by PHP.

As a result, the size of the build folder could be reduced from 872KB to 14.7KB.

I hope this PR helps.


matiasbenedetto (approved):
Thanks for this enhancement!
Working as advertised.
LGTM :rocket:

PR #214, opened 2023-02-08
Add spinner while google fonts load instead of showing a blank page

matiasbenedetto:
This PR adds a spinner while the google fonts data load (which can take some time in slow connections because it is a big file).
Instead of showing a blank screen, we can show part of the interface with a spinner and some text saying it is loading.

## Before: 
[Screencast from 09-02-23 00:09:40.webm](https://user-images.githubusercontent.com/1310626/217603063-80fda2ce-640e-4954-a2f8-a61c95817228.webm)

## After:
[Screencast from 09-02-23 00:05:52.webm](https://user-images.githubusercontent.com/1310626/217603189-5c02fe68-adad-42f8-bb53-5b352908e806.webm)


mikachan (approved):
Nice improvement, works well 👍

PR #219, opened 2023-02-10
Fix Depreciation Warning

cvladan:
This depreciation warning is really driving me crazy:

```
PHP Deprecated:  Creation of dynamic property Create_Block_Theme::$loader is deprecated in ... class-create-block-theme.php on line 49
```

This is just a quick fix.

### DIFF
--- a/includes/class-create-block-theme.php
+++ b/includes/class-create-block-theme.php
@@ -8,6 +8,7 @@
  * @subpackage Create_Block_Theme/includes
  * @author     WordPress.org
  */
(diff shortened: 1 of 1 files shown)


matiasbenedetto (approved):
I Tested on [PHP 8.2](https://www.php.net/manual/en/class.allowdynamicproperties.php) I could see the deprecation message. This fix removes the warning.
LGTM
Thanks, @cvladan for the PR! :rocket:

PR #224, opened 2023-02-15
Add: input for theme tags

### LINKED ISSUE #201: Consider adding input for style.css theme tags
It might be helpful to add a form input for [theme tags](https://make.wordpress.org/themes/handbook/review/required/theme-tags/) that can be added to the list of tags in the exported theme's style.css.

mikachan:
Closes https://github.com/WordPress/create-block-theme/issues/201.

This PR adds checkbox inputs for the official theme tags that can be added to a generated theme, plus a free text field to add any custom tags. The tags list is based on the theme tags [listed here](https://make.wordpress.org/themes/handbook/review/required/theme-tags/). This list is generated using `get_theme_feature_list()`.

When using the child or clone options, the tags list will be pre-populated with the theme tags from the active theme and the default tags. They can be changed before generating the new theme.

When using the blank theme option, the tags list just uses the default list, which again can be changed. 

The default tag list is currently just: 'full-site-editing'. We can probably add 'translation-ready' as well, once all strings in the generated themes are translatable.

This PR also:

- Refactors `toggleForm()`. I found this easier to work with, plus this should help if we combine some of the theme options.
- Moves the "Items indicated with (*) are required." notice to the top of the forms.
- Moves the 'Generate' button to below the form when the 'new theme' form is being used. The form is pretty long with the tags list, so this is one solution I thought of. Maybe instead of doing this, the tags list should be hidden by default, and could be expanded with a button, like "Edit Theme Tags"?

Child or clone options:

image

Blank theme option:

image

### Testing
Please try exporting a theme using the child, clone, and blank theme options. Try out leaving the tag list as its default and try adding and removing other tags.


admin/js/form-script.js
```diff
@@ -1,9 +1,144 @@
...
+		`${ checkboxesSelector }:not(:checked)`
+	);
+
+	if ( checked.length >= max ) {
+		for ( let i = 0; i < unchecked.length; i++ ) {
+			unchecked[ i ].setAttribute( 'disabled', true );
+		}
+	}
+
+	if ( checked.length < max ) {
```
jffng: Could this be an `else` statement of the conditional right before this one?

admin/js/form-script.js
```diff
@@ -1,9 +1,144 @@
...
+		for ( let i = 0; i < unchecked.length; i++ ) {
+			unchecked[ i ].removeAttribute( 'disabled' );
+		}
+	}
+
+	if ( allCheckboxes.length > max ) {
+		for ( let i = 0; i < allCheckboxes.length; i++ ) {
+			allCheckboxes[ i ].addEventListener( 'change', function () {
```
jffng: Why is this `addEventListener` called inside the `limitCheckboxSelection` function? Wouldn't it only need to be called once when the form loads?

jffng (commented):
This is a great addition, thanks for working on it!

It's working well in my testing. 

One general comment — it looks like the position of the "Generate" button moved with this PR, I think it should be kept in the same spot for UI consistency with the other options:

Trunk | This PR
---- | ----
Screenshot 2023-02-17 at 11 47 00 AM | Screenshot 2023-02-17 at 11 44 07 AM

mikachan:
Thanks for testing!

> One general comment — it looks like the position of the "Generate" button moved with this PR, I think it should be kept in the same spot for UI consistency with the other options:

Yes, I wasn't sure how best to handle this. I've moved the button to the bottom of the form when the larger form is in view, but it moves back to its original position if the variation form or none of the forms are in view. This was because I thought the button may not be as obvious on smaller screens if the larger form is in view, as the button may be scrolled out of view. But I'm not sure - I'm not particularly fond of moving it! Shall I try just keeping the button in its original position?

admin/js/form-script.js
mikachan: Yeah, I think that works the same! Done in https://github.com/WordPress/create-block-theme/pull/224/commits/6ec658bdf7c5c73edf0f0064a652bce8a16eee69.

admin/js/form-script.js
mikachan: Ugh yes, thank you! I've moved this to `validateThemeTags`, which is run when the form loads, in https://github.com/WordPress/create-block-theme/pull/224/commits/277e0caefa3f1b789cede52ff1ea11ccb2907ff8. This fixes a massive performance bug too 🙇‍♂️

jffng:
> Shall I try just keeping the button in its original position?

I see what you mean, but yeah I think keeping the button in its original position is better.

mikachan:
Yup ok, I've changed it so the button stays in its original position in https://github.com/WordPress/create-block-theme/pull/224/commits/7456e8feccc54a12bb2aeed972270a25dc36fd83.

madhusudhand (approved):
LGTM!

PR #226, opened 2023-02-15
Cleanup Manage Theme Fonts UI

jffng:
This PR implements a few changes recommended by @beafialho:

Before | After
---- | ----
Screenshot 2023-02-15 at 3 54 41 PM | Screenshot 2023-02-15 at 3 54 36 PM

I wonder if we should refactor more of the app to react before doing this.


beafialho:
Thanks for addressing this @jffng! 

One other thing I noticed recently that would be an improvement is to have the fonts dropdowns collapsed by default, each font panel is so large it fills the entire screen. That way, it would be easier to browse all the font families. A smaller default font size for the font's demo text could help with this too.

matiasbenedetto (approved):
Thanks for the enhancement.
LGTM :rocket:

jffng:
@beafialho nice suggestions and thanks for opening up https://github.com/WordPress/create-block-theme/issues/227, we can address further improvements in a follow up.

PR #232, opened 2023-02-17
Cloned themes: Add original theme name to readme.txt and style.css

### LINKED ISSUE #223: How about leaving the name of the cloned original theme somewhere?
Currently, when I clone from Create Block Theme and start customizing a new block theme, I cannot find the name of the original block theme in readme.txt or style.css.

This plugin is very promising and will definitely make block theme creation easier. Therefore, many more block theme authors will be created in the future.

As more block themes are created in the future, there will naturally be more cloned block themes.

In that case, if we find a problem or defect in a block theme, the more clones there are, the more difficult it will be to find out where the problem lies with the current system.

So I think it would be good if somewhere in the cloned block theme (`Readme.txt` or in the comment out of `style.css` would be best, in my opinion), it would be noted which theme is the original of the block theme.

This would be a respectful gesture to the author of the block theme from which it was cloned.

Thank you.

mikachan:
Resolves https://github.com/WordPress/create-block-theme/issues/223.

This adds the original theme name to the output readme.txt of a cloned theme.

It uses the following text: "[New Theme] is based on [Original Theme] ([Original Theme URL), (C) [Original Theme Author], [Original Theme License]([Original Theme License URL)"

readme.txt:
image

### DIFF
--- a/admin/class-create-block-theme-admin.php
+++ b/admin/class-create-block-theme-admin.php
@@ -165,6 +165,7 @@ class Create_Block_Theme_Admin {
 		$theme['author_uri'] = sanitize_text_field( $theme['author_uri'] );
 		$theme['slug'] = $this->get_theme_slug( $theme['name'] );
 		$theme['template'] = wp_get_theme()->get( 'Template' );
+		$theme['original_theme'] = wp_get_theme()->get( 'Name' );
 
 		// Create ZIP file in the temporary directory.
 		$filename = tempnam( get_temp_dir(), $theme['slug'] );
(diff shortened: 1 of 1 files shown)


carolinan:
To be picky :) , it needs to also include copyright and licence information.

mikachan:
That makes sense! Hmm, maybe the 'original theme' information should also sit under the Copyright section of the readme.txt file, rather than in the description. Something like:

> == Copyright ==
> 
> [New Theme] WordPress Theme, (C) 2023
> [New Theme] is distributed under the terms of the GNU GPL.
> 
> [New Theme] is based on [Original Theme], (C) 2022
> [Original Theme] is distributed under the terms of the GNU GPL.
> 
> This program is free software: you can redistribute it and/or modify...

matiasbenedetto (approved):
This is working as advertised.
LGTM :rocket:

PR #233, opened 2023-02-18
Fix: composer scripts doesn't work on Windows

t-hamano:
**Note**: I added the --no-verify option on commit to prevent the large number of diffs created by lint-staged.

This PR fixes the composer's lint/format command to work on Windows.

```
PS D:\path\to\create-block-theme> npm run lint:php

> create-block-theme@1.5.1 lint:php D:\path\to\create-block-theme
> ./vendor/bin/phpcs

'.' is not recognized as an internal or external command,
operable program or batch file.
```

As seen in [the issue in the npm repository](https://github.com/npm/npm/issues/11721), npm scripts starting with `./` do not work on Windows. Therefore, I used `composer run-script` instead, as is done in the [Gutenberg repository](https://github.com/WordPress/gutenberg/blob/356a1b1555ab7302a6e17d88c32abcace6feb1e7/package.json#L280).

In addition, `.phpcs.xml.dist` was renamed to `phpcs.xml.dist` because the ruleset file name specified by the composer script does not exist.

---

Confirm that the following commands work:

Should show errors.
```
npm run lint:php
```

Error should appear and the code should be formatted. To restore the automatically changed format, run `git checkout .`.
```
npm run lint:php:fix
```


mikachan (approved):
Thanks for fixing this! Sorry, this was my bad. I didn't realise scripts starting with `./` don't work on Windows.

I can confirm `npm run lint:php` and `npm run lint:php:fix` still work the same for me.

🚢

PR #234, opened 2023-02-18
Fix: Unexpected action when clicking Collapse chevron

### LINKED ISSUE #231: Bug: Unexpected action when clicking "Collapse" chevron
I noticed this yesterday, as you can see in the video, when I'm trying to collapse two of the font families, I see an unexpected modal asking if I want to remove another font.

https://user-images.githubusercontent.com/65220155/219647274-f641476f-fada-4698-a8e9-77598c705ce7.mp4

t-hamano:
Fixes: #231

**Note**: I added the --no-verify option on commit to prevent the large number of diffs created by lint-staged.

I think the problem is probably due to the fact that the `tr` element did not have `overflow:hidden` in effect by default. Thus, the content of the closed font will overshadow the next element while remaining invisible.

close

### DIFF
--- a/src/manage-fonts/manage-fonts.css
+++ b/src/manage-fonts/manage-fonts.css
@@ -26,6 +26,7 @@
 
 .font-family-contents .container {
     overflow: hidden;
+    display: block;
 }
 
 .font-family-contents .slide {


mikachan (approved):
Thank you for working on this! This fixes the issue for me 👍

PR #237, opened 2023-02-21
Refactor create-block-theme class

madhusudhand:
### WHY?

With increasing features, the class `Create_Block_Theme_Admin` has grown to more than 1500 lines of code and its very difficult to understand.

This change aims at splitting the class into multiple classes for better developer productivity.
Following are the newly split classes

* Theme_Blocks
* Theme_Json
* Theme_Media
* Theme_Patterns
* Theme_Styles
* Theme_Templates
* Theme_Utils
* Theme_Zip
* Create_Theme_Form
* Form_Messages

### Testing instructions:

Validate the usecases related to create block theme. They must work without any issues.


admin/create-theme/create-theme-form.php
```diff
@@ -0,0 +1,174 @@
+<?php
+
+require_once( __DIR__ . '/theme-tags.php' );
+
+class Create_Theme_Form {
```
draganescu: This is unusual at least as a name for a class to be named like an action. Maybe `Theme_Form` would suffice?

admin/create-theme/create-theme-form.php
```diff
@@ -0,0 +1,174 @@
...
+							<input type="hidden" name="nonce" value="<?php echo wp_create_nonce( 'create_block_theme' ); ?>" />
+						</div>
+					</div>
+				</div>
+			</form>
+		</div>
+	<?php
+	}
+
+	public static function form_script() {
```
draganescu: same here a more descriptive `enqueue_form_assets` could work better

draganescu:
Breaking up a [god object](https://en.wikipedia.org/wiki/God_object) is always good refactoring.

I would refine this more strictly by naming things to sound more like what they are. 

For example:
- If a class is a helper or utility (which all classes made entirely of static functions kind of are) the name should say so. 
- If the method sets or gets we should make the distinction in the name (e.g. `build_readme_txt` is good but `original_theme_credits` is ambiguous as it sounds like a property more than a method.

madhusudhand:
Thanks @draganescu for the great feedback.

For this iteration I restricted myself to not to do any major changes to the functions except for moving them  into logical chunks, so that it will be easy to test and avoid too many conflicts with ongoing code merges.

I agree, the function names aren't great. Also, some of the names have theme specific keywords `blockbase_save_theme` that stayed from the early prototype.

In the followup I intend to target all function names and rewrite to more meaningful ones. Your examples should definitely help me get there.

admin/create-theme/create-theme-form.php
madhusudhand: Makes sense. Renamed the class name. 👍

admin/create-theme/theme-tags.php
```diff
@@ -0,0 +1,166 @@
+<?php
+/**
+ * Handles the theme tags functionality.
+ *
+ * @package Create Block Theme
```
jffng, line 5: Should we add a doc block statement like this to each of the classes, along with a `@ since`?

jffng (approved):
This works against in testing against all six of the existing CBT options.

> In the followup I intend to target all function names and rewrite to more meaningful ones. Your examples should definitely help me get there.

Sounds good to me 👍

jffng:
Merge conflicts are resolved, I tested and all cases still working as expected. I think we should merge to avoid more conflicts as this touches a large surface area of code of the main plugin class. We can address further work like doc block + renaming functions in follow ups. Nice work @madhusudhand !

PR #238, opened 2023-02-21
Replacing mkdir() calls with WordPress wp_mkdir_p() function

matiasbenedetto:
## What?
In this PR we are removing all the remaining calls to PHP's `mkdir()` and replacing them with the WordPress `wp_mkdir_p()` function.

## Why?
To simplify the code and reduce the potential error surface. By leveraging `wp_mkdir_p()` we can remove the permissions parameter and let WordPress take care of that.

This user report seems to be related to that: https://wordpress.org/support/topic/permissions-on-assets-directory/
This PR should fix it.


mikachan (approved):
This looks good to me 👍

PR #241, opened 2023-02-21
Use Gutenberg Theme JSON resolver if its available

### LINKED ISSUE #240: Custom CSS from Global Styles is not being added to the themes created with the plugin
## How to reproduce:

- Add custom CSS from global styles
- Save
- Export / clone the theme
- Check that the CSS you added is not in the exported theme

image

matiasbenedetto:
## What?
Use Gutenberg Theme JSON resolver to get user config if it's available.

## Why?
There are certain new features as the custom CSS in global styles that depends on the Gutenberg resolver because they are not coming from the core resolver.

Fixes: #240

### DIFF
--- a/admin/resolver_additions.php
+++ b/admin/resolver_additions.php
@@ -61,7 +61,11 @@ function augment_resolver_with_utilities() {
  				$theme->merge( $theme_theme );
 			}
 
-			$theme->merge( static::get_user_data() );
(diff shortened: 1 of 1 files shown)


jffng (approved):
Works for me with and without Gutenberg activated. Verified css is correctly saved in theme.json on export.

PR #247, opened 2023-02-24
Is is better placeholder URL changing to TT3 ?

Olein-jp:
Currently, the standard theme is Twenty Twenty Three, so a URL pointing to that theme would be fine.


madhusudhand (approved):
Thanks for working on this.

LGTM!

Olein-jp:
Thank you!

Issue #217, opened 2023-02-09
Could we run the pre-commit linter/prettier just on the modified lines of a commit and not on the entire file?


matiasbenedetto:
The idea is to make the PRs that modify large unformatted files readable.

mikachan:
I'm not sure I've seen a linter work like this before, and if we implemented this it would take much longer for the whole project to be linted correctly. There aren't actually that many files that need to be fixed - as soon as they're linted, this won't be a problem. We could probably tackle a couple of files in one PR, such as all the CSS.

But I'm happy to be persuaded otherwise! It looks like there's an ESLint plugin that helps you lint only the changed lines, so that would cover the JS files: https://www.npmjs.com/package/eslint-plugin-diff

t-hamano:
Personally, I think it is better to format all files according to the current rule set. For example, format CSS, JS, PHP and others (yml, json) with their respective PRs. If we can complete the formatting first, we will be able to see only pure diff in PRs.

mikachan:
As a start, I've run the linter on all the CSS files here: https://github.com/WordPress/create-block-theme/pull/235

madhusudhand:
Please hold with the PHP linter until #237 is merged to avoid conflicts.

madhusudhand:
Fixed lint issues with auto fixer here #252

madhusudhand:
All the list issues are fixed in #252, #253, #254, #255
Any new commit will notify the lint issues and they are expected to be fixed before creating the PR.
Closing this issue as it is no longer needed.

PR #253, opened 2023-03-01
Fix JS lint issues

madhusudhand:
Fixes JS lint issues

Related: #217

### DIFF
--- a/src/manage-fonts/index.js
+++ b/src/manage-fonts/index.js
@@ -1,6 +1,7 @@
-import { useState, useEffect } from 'react';
+import { useState, useEffect } from '@wordpress/element';
 import FontFamily from './font-family';
 import {
+	// eslint-disable-next-line
 	__experimentalConfirmDialog as ConfirmDialog,
 	Modal,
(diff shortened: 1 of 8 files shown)


src/manage-fonts/index.js
```diff
@@ -181,16 +182,26 @@ function ManageFonts() {
 				{ fontToDelete?.fontFamilyIndex !== undefined &&
 				fontToDelete?.fontFaceIndex !== undefined ? (
 					<h3>
-						{ __(
-							`Are you sure you want to delete "${ fontFaceToDelete?.fontStyle } - ${ fontFaceToDelete?.fontWeight }"  variant of "${ fontFamilyToDelete?.fontFamily }" from your theme?`,
-							'create-block-theme'
+						{ sprintf(
```
jffng, line 185: Maybe not an issue, but I find it a strange that the i18n implements its own string formatting function that mirrors the php one instead of just using ES6+ template literals.

jffng (approved):

PR #260, opened 2023-03-06
Fonts outline sidebar

matiasbenedetto:
## What?

- Adds a sidebar with the fonts selected to add / available in the theme (depending on the screen.
- Add extra info as the number of variants selected/available and the file size of the fonts.
- Improves the look of the Manage Fonts and Google fonts screen
- Refactor some parts of the react app to make it more modular and easy to follow.

## How to test:
- Use the manage fonts functionality
- Use the google fonts functionality
- Test that those functionalities work correctly as before

This PR is
Part of: https://github.com/WordPress/create-block-theme/issues/227

PR #261, opened 2023-03-06
Manage fonts minor refactor. Move elements from PHP to react

matiasbenedetto:
## What?

In this PR, I'm moving the rendering of the UI elements from PHP to the React app.

## Why?
- To make the code simpler
- To make us able to implement a more complex UI as the one we are implementing here: https://github.com/WordPress/create-block-theme/pull/260

## How to test?
Use the "Manage fonts" screen to check if it works as before.


jffng (approved):
LGTM, the "Manage Fonts" page header is still rendering and functioning as expected.

PR #263, opened 2023-03-07
Try a workflow for submitting theme updates to git.

jffng:
An experiment for a workflow to submit local changes to a remote repository.


pbking:
git integration looks like it's not going to be a part of CBT (the closest being the playground integration).

Closing as "an interesting experiment".

PR #268, opened 2023-03-10
Local fonts section implementation in React

matiasbenedetto:
## What?
In this PR:
- Implement the local font asset upload in the React App
- Add variable fonts preview capabilities 
- Add variable wieght font logic to build the theme.json (doesn't add fontWeight property if the fonts has a a `wght`variable axis.
- Change the styles of this section to make it look as the rest of the font section as implemented in https://github.com/WordPress/create-block-theme/pull/260

## Why?
To allow us to implement the user ability to add more configuration options.

### For classic fonts:
- Go to the "Add local font" page
- Add a few fonts a test if everything is working as before.

### For variable fonts: 
- Go to the "Add local font" page
- Upload a variable font (preferably one that implements the `wght` variable axis).
- Try the variable font preview options
- Upload the font


src/local-fonts/index.js
```diff
@@ -0,0 +1,100 @@
...
+	const debounceOnFormDataChange = useDebounce( onFormDataChange, 500 );
+
+	useEffect( () => {
+		debounceOnFormDataChange();
+	}, [ formData ] );
+
+	return (
+		<div className="layout">
+			<main>
+				<h1>Local Fonts</h1>
```
jffng: Should we mark this for localization?

src/local-fonts/upload-font-form.js
```diff
@@ -0,0 +1,239 @@
...
+	const nonce = document.querySelector( '#nonce' ).value;
+
+	const onFileSelectChange = ( event ) => {
+		const file = event.target.files[ 0 ];
+
+		if ( ! file ) {
+			setFormData( INITIAL_FORM_DATA );
+		}
+
+		// Use FileReader to, well, read the file
```
jffng, line 29: 😅

src/local-fonts/upload-font-form.js
```diff
@@ -0,0 +1,239 @@
...
+						value={ fontVariationSettings }
+					/>
+				) }
+			</form>
+
+			{ /* Render the range controls for each available axis of a variable font */ }
+			{ formData.variable && (
+				<div className="variable-settings">
+					<div className="header">
+						<p>Variable font settings:</p>
```
jffng: Mark for i18n?

src/local-fonts/upload-font-form.js
```diff
@@ -0,0 +1,239 @@
...
+						<Button
+							isSmall
+							icon={ update }
+							variant="secondary"
+							onClick={ resetDefaultVariableSettings }
+						>
+							Default settings
```
jffng: Mark for i18n?

src/local-fonts/upload-font-form.js
```diff
@@ -0,0 +1,239 @@
...
+						type="hidden"
+						name="font-variation-settings"
+						value={ fontVariationSettings }
+					/>
+				) }
+			</form>
+
+			{ /* Render the range controls for each available axis of a variable font */ }
+			{ formData.variable && (
+				<div className="variable-settings">
```
jffng: I think it might make more sense if the axis controls for the variable font were incorporated into the preview settings somehow. Where they are placed right now feels like that by changing the variable font axis you are somehow changing the settings that font will be uploaded to the theme with. 

Suggestion:
Screenshot 2023-03-14 at 10 43 31 AM

You could also just have one "reset" button that way too.

jffng (commented):
My comments are mostly UI related but it's mostly working as expected, thanks!

> Change the styles of this section to make it look as the rest of the font section as implemented in https://github.com/WordPress/create-block-theme/pull/260

So on PR 260 I see this is what adding a Google Font looks like:

Screenshot 2023-03-14 at 10 18 34 AM

On this PR, the local fonts page looks like this:

Screenshot 2023-03-14 at 10 15 01 AM

Should we aim to bring these closer together in this PR? For example, the padding and the layout are different. I think the initial state looks a bit "broken" until you load a font because there is no indication that the right column is reserved for the font preview:

Screenshot 2023-03-14 at 10 26 30 AM

src/local-fonts/upload-font-form.js
matiasbenedetto: Good point! I moved the variable fonts preview options to the demo option preview panel

matiasbenedetto:
> I think the initial state looks a bit "broken" until you load a font because there is no indication that the right column is reserved for the font preview

Good point, I added a message for the initial state.
image

When a variable font is loaded it looks like this now:
image

jffng (approved):
Nice improvements, thanks! I think it's in a spot we can merge and iterate. Pre-approving with a couple notes:

1. Let's try and bring the Google Fonts and Local Fonts layout closer together in a follow up.
2. I noticed a security warning when I ran `npm install`. So maybe running `npm audit` and fixing the package vulnerabilities would be of good conscience.

Issue #269, opened 2023-03-10
Fatal Error on plugin setup (PHP 8.0)


kishanjasani:
**How to Replicate:**
1. Clone plugin from repo: `git clone git@github.com:WordPress/create-block-theme.git`
2. Run `composer install` and `npm install`
3. If you forget to run `npm run start` or `npm run build`
4. It will throw an error on `Apperance -> Manage Theme Fonts`

**Screenshot:**
https://prnt.sc/50oJvHvkyJaQ

matiasbenedetto:
Hi, @kishanjasani ! :wave: 

I think this is the expected behavior (failing if the javascript app is not built yet).

The need to run `npm run build` is documented here https://github.com/WordPress/create-block-theme/blob/trunk/CONTRIBUTING.md#how-can-i-contribute so I'm closing this issue. Feel free to re-open if you think we should change something around this.

PR #270, opened 2023-03-10
Fix fatal error on plugin setup

kishanjasani:
Issue: #269 
- If you forget to run npm run start it will not through fatal error.

Screenshot: 
https://prnt.sc/GXp8rMOETYId

### DIFF
--- a/admin/manage-fonts/fonts-page.php
+++ b/admin/manage-fonts/fonts-page.php
@@ -33,7 +33,13 @@ class Fonts_Page {
 	public static function load_fonts_react_app() {
 		// Load the required WordPress packages.
 		// Automatically load imported dependencies and assets version.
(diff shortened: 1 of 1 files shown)


jffng (commented):
Hmm if the app has not been built, I think in some ways the fatal error reporting is more useful than showing nothing. Could we provide some messaging / error reporting instead?

kishanjasani:
Yes, We can show error/warning message. I will update my pull request.

matiasbenedetto:
Since the issue motivating this PR [doesn't seem to be valid ](https://github.com/WordPress/create-block-theme/issues/269#issuecomment-1501974565)I think we could close this PR. Feel free to re-open if you think we need to implement this.

Thanks for your work trying to improve the plugin! it is very much appreciated.

Issue #230, opened 2023-02-17
Font Management: accessibility improvements


mikachan:
I noticed a couple of accessibility improvements that could be made to the font management page:

- [ ] The expand/collapse icon on each font family heading is missing a label. I think we should add a label based on [these best practices](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#icon_buttons).

<img width="362" alt="image" src="https://user-images.githubusercontent.com/1645628/219607833-bacd8a07-f662-46d2-a431-5a096e296def.png">

- [ ] It would be good to add the name of the font family in each remove button, so it's clear which font family they are linked to. For example, add an `aria-label` with the contents, "Remove DM Sans font family".

<img width="269" alt="image" src="https://user-images.githubusercontent.com/1645628/219614141-3eabfdfd-d13a-4f34-be15-64d47814c0f7.png">

- [ ] Similar to above, it would be good to add an `aria-label` to the font face buttons too.

<img width="282" alt="image" src="https://user-images.githubusercontent.com/1645628/219614392-4ced68d3-0fda-4850-af2b-f665e7c232be.png">

- [ ] The tab order of the font size range control is reversed because the flex-direction is set to `row-reverse`. This is probably because the `RangeControl` component only allows for the input to be shown on the right, and in this case, it works better on the left. I'm not sure what's possible here but thought it was worth mentioning. Maybe we could look into changing the tab order somehow.

<img width="347" alt="image" src="https://user-images.githubusercontent.com/1645628/219611600-ad3e7d19-203b-4e3f-89a1-adac671166b1.png">

kishanjasani:
@mikachan there is one more accessibility related issue is with color contrast for buttons.

kishanjasani:
@mikachan  For the last point, 

> The tab order of the font size range control is reversed because the flex-direction is set to row-reverse. This is probably because the RangeControl component only allows for the input to be shown on the right, and in this case, it works better on the left. I'm not sure what's possible here but thought it was worth mentioning. Maybe we could look into changing the tab order somehow.

Using the flex-direction property with values of row-reverse or column-reverse will create a disconnect between the visual presentation of content and DOM order. This will adversely affect users experiencing low vision navigating with the aid of assistive technology such as a screen reader. If the visual (CSS) order is important, then screen reader users will not have access to the correct reading order.

**Alternative solution,**
So we can add `RangeControl` with `withInputField={ false }` props. Currently, It's `withInputField={ true }`

[...]

PR #271, opened 2023-03-11
Improve accesibility for Manage theme fonts page

kishanjasani:
Issue: #230 

- [x] The expand/collapse icon on each font family heading is missing a label. I think we should add a label based on [these best practices](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#icon_buttons).
 -> Add `aria-expands: true/false` based on the user action, Also added `aria-label` to the button and `aria-hidden: true` for icon.

- [x] It would be good to add the name of the font family in each remove button, so it's clear which font family they are linked to. For example, add an aria-label with the contents, "Remove DM Sans font family".
  -> Add `aria-label` for button and `aria-hidden: true` for icon.

- [x]  Similar to above, it would be good to add an aria-label to the font face buttons too.
  -> Similarlly, Adds `aria-label` for button and `aria-hidden: true` for icon.

- [ ]  The tab order of the font size range control is reversed because the flex-direction is set to row-reverse. This is probably because the RangeControl component only allows for the input to be shown on the right, and in this case, it works better on the left. I'm not sure what's possible here but thought it was worth mentioning. Maybe we could look into changing the tab order somehow.

All checked item has been improved, the last one is pending. Last item needs discussion, how we can improve that? Needs 
 you opinion.

### DIFF
--- a/src/demo-text-input/index.js
+++ b/src/demo-text-input/index.js
@@ -3,6 +3,8 @@ import {
 	RangeControl,
 	SelectControl,
 	// eslint-disable-next-line
+	__experimentalNumberControl as NumberControl,
+	// eslint-disable-next-line
 	__experimentalInputControl as InputControl,
 } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
@@ -47,14 +49,30 @@ function DemoTextInput() {
 					onChange={ handleDemoTextChange }
 				/>
 
-				<div>
-					<RangeControl
+				<div className="font-size-input">
+					<NumberControl
 						label={ __( 'Font size (px)', 'create-block-theme' ) }
+						aria-label={ __(
+							'Font size (px)',
+							'create-block-theme'
+						) }
+						inputMode="decimal"
+						max={ 140 }
+						min={ 8 }
+						onChange={ handleDemoFontSizeChange }
+						shiftStep={ 1 }
+						value={ demoFontSize }
+					/>
+					<RangeControl
+						aria-label={ __(
+							'Font size (px)',
+							'create-block-theme'
+						) }
 						value={ demoFontSize }
 						onChange={ handleDemoFontSizeChange }
 						min={ 8 }
 						max={ 140 }
-						withInputField={ true }
+						withInputField={ false }
 					/>
 				</div>
 
--- a/src/manage-fonts/font-family.js
+++ b/src/manage-fonts/font-family.js
@@ -3,7 +3,7 @@ import { Button, Icon } from '@wordpress/components';
 import FontFace from './font-face';
 import { ManageFontsContext } from '../fonts-context';
 
-const { __, _n } = wp.i18n;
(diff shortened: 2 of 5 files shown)


src/demo-text-input/index.js
```diff
@@ -47,14 +49,30 @@ function DemoTextInput() {
 					onChange={ handleDemoTextChange }
 				/>
 
-				<div>
-					<RangeControl
+				<div className="font-size-input">
+					<NumberControl
```
mikachan, line 53: Does the `NumberControl` work OK if it's placed after the `RangeControl`? I believe this is the intended design, so it would be great if we could keep it like that:

image

src/manage-fonts/font-family.js
```diff
@@ -59,19 +59,40 @@ function FontFamily( {
 									e.stopPropagation();
 									deleteFontFamily( fontFamilyIndex );
 								} }
+								aria-label={ sprintf(
+									/* translators: %s: Font Family. */
+									__( 'Remove %s Font Family' ),
+									fontFamily.name || fontFamily.fontFamily
+								) }
```
mikachan, line 66: Nice!

src/manage-fonts/font-family.js
```diff
@@ -59,19 +59,40 @@ function FontFamily( {
...
+										? __(
+												'Collapse Font Family',
+												'create-block-theme'
+										  )
+										: __(
+												'Expand Font Family',
+												'create-block-theme'
+										  )
+								}
+							>
```
mikachan, line 88: 👍

src/manage-fonts/index.js
```diff
@@ -192,8 +192,9 @@ function ManageFonts() {
 					<Button
 						onClick={ toggleIsHelpOpen }
 						style={ { padding: '0', height: '1rem' } }
+						aria-label={ __( 'Alert Popup', 'create-block-theme' ) }
```
mikachan, line 195: I'd maybe label this as something like 'More information' or 'Additional information', rather than 'Alert'.

mikachan (commented):
Thanks so much for working on this!

I've left a couple of inline comments, but this needs to be rebased so it's up-to-date with the latest changes from the plugin. Please could you do that, and I'll take another look 🙇

src/manage-fonts/index.js
kishanjasani: @mikachan Thanks for your feedback.

kishanjasani:
@mikachan I think we need to create another PR for it. there are many changes in the font page.

mikachan:
> I think we need to create another PR for it. there are many changes in the font page.

Yeah that's true, it's probably easier to handle the changes in a new PR 👍

vcanales:
Thanks for your contribution! Font management is being handled by core now, so we're working towards removing it from the plugin on the next release.

PR #272, opened 2023-03-13
Allow otf font file upload

matiasbenedetto:
## What? 
Allows uploading OTF font files 

## Why? 
Because it is a valid font file supported widely by browsers.
https://caniuse.com/ttf
https://www.w3schools.com/css/css3_fonts.asp

It was requested by a user here:
https://wordpress.org/support/topic/font-loading-not-working/?view=all#post-16550790


jffng (approved):
Successfully uploaded .otf font, thanks!

PR #279, opened 2023-03-16
Variable font weight range

matiasbenedetto:
## What ?
Fonts with variable weight axis will autocomplete a font-weight range instead of nothing.

## Why ?
- To follow css specification: https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-weight
- To avoid the 400 font-weight applied by default by WordPress when the font-weight definition is missing in `theme.json`.

## Screenshot
image

### DIFF
--- a/admin/class-manage-fonts.php
+++ b/admin/class-manage-fonts.php
@@ -146,6 +146,7 @@ class Manage_Fonts_Admin {
 			wp_verify_nonce( $_POST['nonce'], 'create_block_theme' ) &&
 			! empty( $_FILES['font-file'] ) &&
 			! empty( $_POST['font-name'] ) &&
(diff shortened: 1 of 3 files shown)


jffng (approved):
Works as advertised. 

I noticed a potential bug, or is it the expected behavior for the font preview to disappear if font-weight is cleared?


https://user-images.githubusercontent.com/5375500/226335040-2386fdda-2cff-4f4e-8ca9-ad446f12db26.mov

matiasbenedetto:
Thanks for the detailed review!

> I noticed a potential bug, or is it the expected behavior for the font preview to disappear if font-weight is cleared?

I think It's ok to work like that because allowing the preview without the complete data can lead to inconsistencies between the preview and the final result.

PR #281, opened 2023-03-17
Fix manage fonts UI and backend when no settings are defined in theme.json

matiasbenedetto:
## What ?
Fix manage fonts functionality when the `settings` object is not defined by `theme.json`.

If the `settings.typography.fontFamilies` is not defined in theme.json the manage fonts UI fails to render and the PHP backend throws warnings.

With this PR we are fixing both the UI and the backend.

## How to test
- Remove entirely the `settings` key of your `theme.json`.
- Check if the manage fonts screen is rendering ok
- Add google or local fonts
- Navigate the manage fonts screen and check is working as expected.

Fixes: https://wordpress.org/support/topic/found-an-error-2/


jffng (approved):
Works as advertised, thanks!

PR #282, opened 2023-03-21
fix issue where package-lock is not updated on version bump

madhusudhand:
Right now the version bump script updates all versions except `package-lock.json`
It is required to be updated when `package.json` is updated.

This PR addresses it with the help of `npm version <version>` command.


matiasbenedetto (approved):
Tested locally and working ok. Thanks for the fix!
LGTM

matiasbenedetto (changes requested):
I stressed the test a bit more, and I think I found a problem. I will post the findings here. Meanwhile, don't merge this PR yet, please.

matiasbenedetto (approved):
I added a commit with a little fix to avoid erroring when the tag is not found in the repo.
Seems ready to merge.

PR #283, opened 2023-03-21
Export style variations just with the changes made by the user

### LINKED ISSUE #150: Creating a style variation enhancements
1. Write the name the user enters in the form to the style variation file as the title attribute

Currently, when creating a global styles variation using the option in the plugin, the title attribute for the new style variation file is not set. This could be taken from the name field and written to the style variation JSON.

2. Consider only writing the saved JSON in the database to the style variation file

Additionally, the entire theme.json is read, and then the updated styles are added to the JSON object. Then that entire object is written to the style variation file. However, it is possible to just write the JSON object stored in the database for that new style variation.

matiasbenedetto:
## What ?
- Export a style variation with just the changes made by the user instead of a full copy of the theme.json
- Add the variation title to the variation json.

## Why ?
Because without this change:
- A full copy of theme.json is exported as a variation.
- Variation title is not added to the variant json

Fixes: #150

### DIFF
--- a/admin/class-create-block-theme-admin.php
+++ b/admin/class-create-block-theme-admin.php
@@ -48,7 +48,7 @@ class Create_Block_Theme_Admin {
 	}
 
 	function save_variation( $export_type, $theme ) {
-		Theme_Json::add_theme_json_variation_to_local( $export_type, $theme );
+		Theme_Json::add_theme_json_variation_to_local( 'variation', $theme );
 	}
 
 	/**
--- a/admin/create-theme/theme-json.php
+++ b/admin/create-theme/theme-json.php
@@ -27,9 +27,16 @@ class Theme_Json {
 
 		$_POST['theme']['variation_slug'] = $variation_slug;
 
+		$extra_theme_data = array(
+			'version' => WP_Theme_JSON::LATEST_SCHEMA,
+			'title'   => $theme['variation'],
+		);
+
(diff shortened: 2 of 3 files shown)


mikachan (commented):
Thanks for working on this, this will be a great improvement!

When I test this out, I can see the variation name is correctly saved in the title attribute. And if I create a variation based on the base theme, I can see only the changes I've made are saved in the variation .json file.

However, if I create a variation using the plugin that's based on an existing variation (e.g. if I switch to a variation, customise it, and then create a new variation using the plugin), then the full content of the .json file is copied over. Is it possible to have the same functionality when starting with a variation as well?

matiasbenedetto:
> if I create a variation using the plugin that's based on an existing variation (e.g. if I switch to a variation, customise it, and then create a new variation using the plugin), then the full content of the .json file is copied over. Is it possible to have the same functionality when starting with a variation as well?

I haven't tried this use case, but after reading your description, I think the current behavior makes sense because the new variation is applied to the base theme and not over the original variation, so it should include the styles of the original variation + the new variation ones.  What do you think?

mikachan:
Oh, right! That makes sense. Sorry, I was testing with a variation that had already copied everything from the original theme.json file, so of course that's going to be in the variation created with the plugin. I've now tested with an existing variation that only includes minimal changes, and the newly created variation only includes these minimal changes, plus the changes I've made in the site editor.

In that case, I think this is working well! ✨

mikachan (approved):

PR #290, opened 2023-03-27
Adding default value to an to avoid error when calling export_theme_data()

matiasbenedetto:
Adding default value to an argument to avoid error when calling  `export_theme_data` the function with only one parameter instead of 2.

This error was introduced here: https://github.com/WordPress/create-block-theme/pull/283
The error is only on `trunk`, and it is not been released publicly.


mikachan (approved):
Looks good 👍

PR #291, opened 2023-03-27
Fixing image downloading not working in some cases

### LINKED ISSUE #280: Exported assets/images with 0 bytes
Hi!

When I export the theme with the plugin, the image files used in the templates are exported (as mentioned in #213) but they contain no info, 0 bytes.

To reproduce (WP 6.1.1, Create Block Theme 1.7.0)
- On this plugin, create a blank theme and activate
- On the footer template, add and Image block and upload to the library a png file

imagen
 
- On this plugin, Export the theme
imagen

- The generated .zip file contains the image file in /assets/images, but its size is 0 bytes. So it fails to load when I try to upload this theme into another installation

imagen

p.d. I'm quite new to FSE and so on, git included, so maybe I'm missing something, sorry in advance and thank you for this plugin, I'm learning a lot!

matiasbenedetto:
## What ?
Fixing image downloading and bundling into the theme not working in some cases.

## How ?
Instead of PHP function `file_get_contents()` we are using the WordPress function `download_url()` that seems to be working in most scenarios.

## Why ?
Some users reported this functionality not working in their environments. See #280 as example.

## How to test:
1. Add an image to a template or template part
2. Export / Overwrite the theme
3. See if the image on the assets folder downloaded correctly (not 0-byte size).

Fixes: #280

### DIFF
--- a/admin/create-theme/theme-media.php
+++ b/admin/create-theme/theme-media.php
@@ -137,15 +137,12 @@ class Theme_Media {
 
 	public static function add_media_to_local( $media ) {
 		foreach ( $media as $url ) {
-			$download_file = file_get_contents( $url );
+			$download_file = download_url( $url );
 			$media_path    = get_stylesheet_directory() . DIRECTORY_SEPARATOR . self::get_media_folder_path_from_url( $url );
 			if ( ! is_dir( $media_path ) ) {
 				wp_mkdir_p( $media_path );
 			}
-			file_put_contents(
-				$media_path . basename( $url ),
-				$download_file
-			);
+			rename( $download_file, $media_path . basename( $url ) );
 		}
 	}
 }
--- a/admin/create-theme/theme-zip.php
+++ b/admin/create-theme/theme-zip.php
@@ -156,9 +156,12 @@ class Theme_Zip {
 	static function add_media_to_zip( $zip, $media ) {
 		$media = array_unique( $media );
 		foreach ( $media as $url ) {
-			$folder_path   = Theme_Media::get_media_folder_path_from_url( $url );
-			$download_file = file_get_contents( $url );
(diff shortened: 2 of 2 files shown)


matiasbenedetto:
This zip file contains the plugin version in case you want to download it and test it in your environment: 

[create-block-theme.zip](https://github.com/WordPress/create-block-theme/files/11082108/create-block-theme.zip)

mikachan (commented):
I'm struggling to reproduce the original issue in my testing, so I'm unable to confirm if this fixes the issue. However, I can confirm that this PR works just the same for me, and I see my included images.

It would be awesome if @decrecementofeliz has time to test this, or someone else who has also seen this issue before.

I was also thinking that perhaps we should add improved error handling around the asset management, to handle potential silent errors such as those caused by server settings. e.g. should the plugin check for PHP settings such as `upload_max_filesize` before running `add_media_to_local `? (Not for this PR though!)

matiasbenedetto:
> I was also thinking that perhaps we should add improved error handling around the asset management, to handle potential silent errors such as those caused by server settings. e.g. should the plugin check for PHP settings such as upload_max_filesize before running add_media_to_local ? (Not for this PR though!)

Yes! that would be very useful.

mikachan (approved):
Approving based on the testing by @decrecementofeliz [here](https://github.com/WordPress/create-block-theme/issues/280#issuecomment-1489080176). Thank you!

matiasbenedetto:
Thanks @mikachan  @decrecementofeliz !

carolinan:
Is this part of the released plugin? or do I need to use the development version for images to be added to the assets?

carolinan:
OK, I found the changelog, but I am still experiencing issues.

PR #292, opened 2023-03-27
Add Export (Clone) to site editor

jffng:
This PR is a WIP 🔧 experiment to add CBT's export / clone functionality directly to the site editor.

Screenshot 2023-03-27 at 5 52 12 PM

### DIFF
--- a/admin/class-create-block-theme-admin.php
+++ b/admin/class-create-block-theme-admin.php
@@ -221,8 +221,7 @@ class Create_Block_Theme_Admin {
 				'methods'             => 'POST',
 				'callback'            => array( $this, 'rest_export_theme' ),
 				'permission_callback' => function () {
-					return true;
-					// return current_user_can( 'edit_theme_options' );
+					return current_user_can( 'edit_theme_options' );
 				},
 			)
 		);
--- a/admin/class-create-block-theme-admin.php
+++ b/admin/class-create-block-theme-admin.php
@@ -30,6 +30,7 @@ class Create_Block_Theme_Admin {
 		add_action( 'admin_menu', array( $this, 'create_admin_menu' ) );
 		add_action( 'admin_init', array( $this, 'blockbase_save_theme' ) );
 		add_action( 'enqueue_block_editor_assets', array( $this, 'create_block_theme_enqueue' ));
+		add_action( 'rest_api_init', array( $this, 'register_theme_export' ));
 	}
 
 	function create_block_theme_enqueue() {
@@ -202,9 +203,27 @@ class Create_Block_Theme_Admin {
 		header( 'Content-Disposition: attachment; filename=' . $theme['slug'] . '.zip' );
 		header( 'Content-Length: ' . filesize( $filename ) );
 		flush();
-		echo readfile( $filename );
-		die();
+		readfile( $filename );
+		unlink( $filename );
+		exit;
 	}
+
+	function rest_export_theme( $request ) {
+		$theme = $request->get_params();
+		$this->clone_theme($theme, null);
+	}
+
+	public function register_theme_export() {
+		register_rest_route('create-block-theme/v1', '/export', array(
+			'methods' => 'POST',
+			'callback' => array( $this, 'rest_export_theme' ),
+			'permission_callback' => function () {
+				return true;
+				// return current_user_can( 'edit_theme_options' );
+			}
+		));
+	}
+
 	/**
 	 * Create a child theme of the activated theme
 	 */
--- a/admin/class-create-block-theme-admin.php
+++ b/admin/class-create-block-theme-admin.php
@@ -163,6 +163,11 @@ class Create_Block_Theme_Admin {
 		$theme['template']       = wp_get_theme()->get( 'Template' );
 		$theme['original_theme'] = wp_get_theme()->get( 'Name' );
 
+		// Use previous theme's tags if custom tags are empty.
+		if ( empty( $theme['tags_custom'] ) ) {
+			$theme['tags_custom'] = implode( ', ', wp_get_theme()->get( 'Tags' ) );
+		}
+
 		// Create ZIP file in the temporary directory.
 		$filename = tempnam( get_temp_dir(), $theme['slug'] );
 		$zip      = Theme_Zip::create_zip( $filename );
--- a/admin/class-create-block-theme-admin.php
+++ b/admin/class-create-block-theme-admin.php
@@ -29,6 +29,21 @@ class Create_Block_Theme_Admin {
 	public function __construct() {
 		add_action( 'admin_menu', array( $this, 'create_admin_menu' ) );
 		add_action( 'admin_init', array( $this, 'blockbase_save_theme' ) );
+		add_action( 'enqueue_block_editor_assets', array( $this, 'create_block_theme_enqueue' ));
+	}
+
+	function create_block_theme_enqueue() {
+		$asset_file = include( plugin_dir_path( dirname( __FILE__ ) ) . 'build/editor.asset.php');
+	
+		wp_register_script(
+			'create-block-theme-slot-fill',
+			plugins_url( 'build/editor.js', dirname( __FILE__ ) ),
+			$asset_file['dependencies'],
+			$asset_file['version']
+		);
+		wp_enqueue_script(
+			'create-block-theme-slot-fill',
+		);
 	}
 
 	function create_admin_menu() {
--- a/admin/class-create-block-theme-admin.php
+++ b/admin/class-create-block-theme-admin.php
@@ -34,11 +34,11 @@ class Create_Block_Theme_Admin {
 	}
 
 	function create_block_theme_enqueue() {
-		$asset_file = include( plugin_dir_path( dirname( __FILE__ ) ) . 'build/editor.asset.php' );
+		$asset_file = include( plugin_dir_path( dirname( __FILE__ ) ) . 'build/plugin-sidebar.asset.php' );
 
 		wp_register_script(
 			'create-block-theme-slot-fill',
-			plugins_url( 'build/editor.js', dirname( __FILE__ ) ),
+			plugins_url( 'build/plugin-sidebar.js', dirname( __FILE__ ) ),
 			$asset_file['dependencies'],
 			$asset_file['version']
 		);
--- a/admin/class-create-block-theme-admin.php
+++ b/admin/class-create-block-theme-admin.php
@@ -29,13 +29,13 @@ class Create_Block_Theme_Admin {
 	public function __construct() {
 		add_action( 'admin_menu', array( $this, 'create_admin_menu' ) );
 		add_action( 'admin_init', array( $this, 'blockbase_save_theme' ) );
-		add_action( 'enqueue_block_editor_assets', array( $this, 'create_block_theme_enqueue' ));
-		add_action( 'rest_api_init', array( $this, 'register_theme_export' ));
+		add_action( 'enqueue_block_editor_assets', array( $this, 'create_block_theme_enqueue' ) );
+		add_action( 'rest_api_init', array( $this, 'register_theme_export' ) );
 	}
 
 	function create_block_theme_enqueue() {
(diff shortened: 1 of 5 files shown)


admin/class-create-block-theme-admin.php
```diff
@@ -187,9 +203,31 @@ function clone_theme( $theme, $screenshot ) {
...
+	public function register_theme_export() {
+		register_rest_route(
+			'create-block-theme/v1',
+			'/export',
+			array(
+				'methods'             => 'POST',
+				'callback'            => array( $this, 'rest_export_theme' ),
+				'permission_callback' => function () {
+					return true;
+					// return current_user_can( 'edit_theme_options' );
```
jffng: This keeps returning `401` unauthorized, I don't know why.

src/editor.js
```diff
@@ -0,0 +1,154 @@
...
+			method: 'POST',
+			headers: {
+				'Content-Type': 'application/json',
+			},
+			body: JSON.stringify( theme ),
+		} )
+			.then( ( response ) => response.blob() )
+			.then( ( blob ) => {
+				const url = URL.createObjectURL( blob );
+				window.location.href = url;
```
jffng: I don't know if this is a good approach, but didn't know how to trigger the file download another way. Also the filename is not being named correctly.

pbking:
I think the "save" functionality is more important than the "export" functionality.  Perhaps we could add that as the first option?

If we "save" (rather than export) and then the native "export" functionality of Gutenberg is used, does that have the same result?  (Perhaps we could just make the bits we add to the Editor a way to change the metadata and save (or "save as" if the slug is changed) simplifying what we're building and how it works with a user's workflow.

jffng:
> I think the "save" functionality is more important than the "export" functionality. Perhaps we could add that as the first option?

This isn't quite the feedback I've heard from designers, but I hear you that "save" makes sense to implement here. 

> If we "save" (rather than export) and then the native "export" functionality of Gutenberg is used, does that have the same result?

Not quite, there is functionality within CBT that does not exist within Gutenberg's export — particularly the escaping of relative URLs and patternizing of templates that contain images.

src/editor.js
scruffian, line 1: We should use a more descriptive file name

src/editor.js
```diff
@@ -0,0 +1,154 @@
+import { registerPlugin } from '@wordpress/plugins';
+import { PluginSidebar, PluginSidebarMoreMenuItem } from '@wordpress/edit-site';
+import { blockDefault } from '@wordpress/icons';
```
scruffian: Not sure if we should use this icon

src/editor.js
```diff
@@ -0,0 +1,154 @@
...
+					label={ __( 'Author URI', 'create-block-theme' ) }
+					value={ theme.author_uri }
+					onChange={ ( value ) =>
+						setTheme( { ...theme, author_uri: value } )
+					}
+					placeholder={ __(
+						'https://wordpress.org/',
+						'create-block-theme'
+					) }
+				/>
```
scruffian: This button looks weird!

src/editor.js
```diff
@@ -0,0 +1,154 @@
...
+				<TextControl
+					label={ __( 'Theme URI', 'create-block-theme' ) }
+					value={ theme.uri }
+					onChange={ ( value ) =>
+						setTheme( { ...theme, uri: value } )
+					}
+					placeholder={ __(
+						'https://github.com/wordpress/twentytwentytwo/',
```
scruffian:
```suggestion
						'https://github.com/wordpress/twentytwentythree/',
```

?

src/editor.js
scruffian: It would be good to use the slug but you'd need a workaround (https://stackoverflow.com/questions/19327749/javascript-blob-filename-without-link)

src/editor.js
```diff
@@ -0,0 +1,154 @@
...
+
+	const handleSubmit = () => {
+		fetch( '/wp-json/create-block-theme/v1/export', {
+			method: 'POST',
+			headers: {
+				'Content-Type': 'application/json',
+			},
+			body: JSON.stringify( theme ),
+		} )
+			.then( ( response ) => response.blob() )
```
matiasbenedetto: What do you think about using using async/await instead of chaining then and catch? Could that increase code readbility?

src/editor.js
scruffian: This should probably be a follow up

src/editor.js
matiasbenedetto: It's the only call to action in the panel. Should it be "primary"?

src/editor.js
```diff
@@ -0,0 +1,154 @@
...
+					value={ theme.author_uri }
+					onChange={ ( value ) =>
+						setTheme( { ...theme, author_uri: value } )
+					}
+					placeholder={ __(
+						'https://wordpress.org/',
+						'create-block-theme'
+					) }
+				/>
+				<Button variant="secondary" onClick={ handleSubmit }>
```
matiasbenedetto: Should we add validation to the form that enable/disable the button?

src/editor.js
```diff
@@ -0,0 +1,154 @@
...
+			name: themeData.name.raw,
+			description: themeData.description.raw,
+			author: themeData.author.raw,
+			author_uri: themeData.author_uri.raw,
+			theme_uri: themeData.theme_uri.raw,
+		} );
+	}, [] );
+
+	const handleSubmit = () => {
+		fetch( '/wp-json/create-block-theme/v1/export', {
```
matiasbenedetto: We would need to make this URL absolute. Otherwise it will fail when WordPress is not served from the root of the domain.

matiasbenedetto: `apiFetch` takes care of this, so it should be fixed with my latest commit: https://github.com/WordPress/create-block-theme/pull/292/commits/16f3c5cc0f59955498eaccf8d367389ebab22620

admin/class-create-block-theme-admin.php
matiasbenedetto: This is because the ajax request is not authenticated. With the implementation of `apiFetch` this hack is no longer necessary: https://github.com/WordPress/create-block-theme/pull/292/commits/16f3c5cc0f59955498eaccf8d367389ebab22620

matiasbenedetto:
I left a few comments about the code. In terms of UI it looks good.

I'm just wondering about:
1. Is there consensus around moving all the interfaces out from wp-admin and into the editor even the plugin ones?
2. If the answer to 1 is Yes: Will we add just one of the options of the plugin to the editor or all of them (export, clone, child, variation, etc)?

jffng:
Thanks for the reviews!

> Is there consensus around moving all the interfaces out from wp-admin and into the editor even the plugin ones?

I think moving all of the interfaces to the site editor is good goal to shoot for. I think the next phase of Gutenberg also focuses on extensibility, so iterating on CBT in the site editor will help inform what needs to be developed in Gutenberg. What is your opinion? 

> If the answer to 1 is Yes: Will we add just one of the options of the plugin to the editor or all of them (export, clone, child, variation, etc)?

Yes, but I think it makes sense to prioritize Export, Overwrite, and Style Variations.

src/editor.js
jffng: Thank you!

src/editor.js
jffng: How about `editor-plugin-sidebar.js`, or just `plugin-sidebar.js`?

src/editor.js
jffng: I'm not sure about primary since there is already the "Save" primary button. What do you think of this?

Screenshot 2023-03-29 at 5 26 47 PM

src/editor.js
jffng: Done in [b5c6a17](https://github.com/WordPress/create-block-theme/pull/292/commits/b5c6a17d5321c2618b363ce38fef49f8fcb7cf6a)

src/editor.js
jffng: Done in [6efb440](https://github.com/WordPress/create-block-theme/pull/292/commits/6efb440d1cafcff3ffd1a8e4548227ab29eb0a80)

src/editor.js
jffng: I updated to use the `tool` icon but I'm not sure about this either, it's also rather generic.

src/editor.js
scruffian: WFM

scruffian:
> Is there consensus around moving all the interfaces out from wp-admin and into the editor even the plugin ones?
If the answer to 1 is Yes: Will we add just one of the options of the plugin to the editor or all of them (export, clone, child, variation, etc)?

I believe all the UI should be in the site editor. That was always the intention but I just stuck it in wp-admin as a quick first pass. If I'd known how popular the plugin would become I'd have never put it there!

scruffian:
I think you can merge this when the linter is happy.

pbking:
>> If the answer to 1 is Yes: Will we add just one of the options of the plugin to the editor or all of them (export, clone, child, variation, etc)?

> Yes, but I think it makes sense to prioritize Export, Overwrite, and Style Variations.

I definitely agree with all of the above that we should be moving all aspects to the site editor.  There are two things to consider doing this:

* Any action other than "export" changes the Theme and the editor will need to re-load those changes.  I've experimented with doing this for other tools (Revisr) but haven't found a good solution yet.  It's not a matter when the actions are in wp-admin, but moving the actions into the Site Editor means that for those actions we'll need to plan to handle the changes.

* It may be wise to revisit the actions and consider alternate paths; potentially consolidating or refactoring that UI rather than just moving the "here are all of the possible options" interface that's currently in wp-admin.  For example we may want to introduce "New", "Save" and "Save As..." options that may better fit in people heads and simplify the actions.  This could separate the "Metadata" editor from "export" since there are multiple use-cases for editing the metadata.

Moving the functionality into the Site Editor is the PERFECT opportunity to refactor that UI and I think that we should do that carefully.  I'm not convinced that we should merge this in until we've had an opportunity to create a target for what that UI could be.

jffng:
@pbking great feedback, thanks for raising those issues.

> Any action other than "export" changes the Theme and the editor will need to re-load those changes

Generating a style variation was demo'd in #96, I'm not sure if it still requires a GB change but that is worth revisiting.

Overwrite does seem trickier.

> It may be wise to revisit the actions and consider alternate paths; potentially consolidating or refactoring that UI rather than just moving the "here are all of the possible options" interface that's currently in wp-admin

Yep, I agree let's not just move all the options into the sidebar as is, but take the time to consider how they could be refactored or improved to fit better into the site editor model. 

>  I'm not convinced that we should merge this in until we've had an opportunity to create a target for what that UI could be.

Do you have suggestions for how the Export functionality and the UI presented in this PR could be reworked? Or are you just suggesting to come up with a more holistic rework before merging anything and iterating?

scruffian:
Personally I don't see a need to put this PR on hold for the sake of future UX improvements. We don't need to maintain backwards compatibility, so we can just merge this and then improve it in another PR. Done is better than perfect.

matiasbenedetto:
I added [a commit ](https://github.com/WordPress/create-block-theme/pull/292/commits/cd5f862db849f65897ffb485809cf2d695c04634) to set the filename of the zip exported. Without it, the user receives a zip file with a random name (example: `736a3448-d2c4-4b05-9e4b-13f42f1658d2.zip`). If the user uploads the zip file with that filename, it becomes a problem because that random name is used as the theme slug name, and some parts of the site are broken.

Apart from that, it looks good to me. I think we can merge and continue iterating over the UI later.

matiasbenedetto (approved):

jffng:
I just thought of a small bug with this PR related to the theme's metadata — adding / editing tags is not supported yet in this flow, so I think we should carry through the tags of the current theme.

Fixed in [fd779d3](https://github.com/WordPress/create-block-theme/pull/292/commits/fd779d30b1b17b1101ef4d1b3f488dedb2c121f8)

src/editor.js
jffng: Fixed by @matiasbenedetto in [cd5f862](https://github.com/WordPress/create-block-theme/pull/292/commits/cd5f862db849f65897ffb485809cf2d695c04634) 🎉

Issue #186, opened 2023-01-25
Merge Clone and Export options


jffng:
Currently Export and Clone are two different options but they effectively do the same thing. Clone just allows you to overwrite the theme metadata (name, author, etc.).

I think it would make sense to merge these two options, and just have the form pre-filled with existing theme metadata. And if the user wants, they can overwrite the metadata if they want. 

What do others think?

mikachan:
Yeah, I like the idea of merging these two options, makes sense to me 👍

matiasbenedetto:
I agree with your proposal if the difference is just the ability to edit the metadata.

tinjure20:
IMO, the three options: export, clone and rewrite could be combined in one, with a select drop-down menu added in the form, to allow users to choose which function to perform. Depending on the selected entry, display the form (for clone) or the notification (to rewrite).

If pattern feature (similar to style variation) is added to the plugin, then it actually allows creating a block theme without writing code. This could be huge for average WordPress users with no/little coding skill.

pbking:
This has been merged in the Site Editor interface.  Closing as complete.

PR #294, opened 2023-03-28
Try: refactor clone to copy theme to current site

### LINKED ISSUE #284: Clone + Overwrite
Given the workflow: 
* Activate a Base Theme to start from (such as Blank Canvas)
* Make Changes using the Site Editor and save them
* (optional) Save the User changes to the activated Base Theme
* Clone the Theme (exporting it with a new name)

At this point a .zip file exists with the changes to the Theme, however those changes exist EITHER:
* In the user space 
* Saved in a modified version of the Base Theme

Unless we upload the newly exported theme .zip to the site we don't have those changes in the site.

I suggest a means of cloning the theme but keeping it on the site rather than exporting a .zip file.  This way changes can continue to be made to the theme and saved locally using the "overwrite" option which would overwrite to the NEW theme rather than to the original Base Theme.

This would also be an alternative to "Creating a Blank Theme" which is currently the only way to CREATE a new theme in the environment.

### LINKED ISSUE #288: Allow installing and activating the theme on creation
When you create a new theme, allow installing it and activating it on the current site, so that we don't need to manually install it from the downloaded theme zip file.

jffng:
This PR is an attempt to address #284 #288 and #186, by modifying the behavior of the Clone option.

Instead of generating a zip, _Clone_ creates a copy of the currently active theme, along with any user modifications, to the current site — providing a convenient way for users to effectively fork an existing theme. 

_Export_ was updated to work like Clone did — allowing users to edit the metadata for an export.

**notes**
- The PR does not activate the theme — that still needs to be done manually.
- If we like this idea, I think we should add the ability to edit the metadata to the Export option as a fast follow up, which would solve #186 EDIT this PR was refactored to solve #186 also.

**To test**
- Make some changes to templates and global styles via the site editor
- Use the Clone option
- Verify the cloned theme appears as expected in Appearance > Themes
- Verify that the export and overwrite options still work as expected


admin/class-create-block-theme-admin.php
```diff
@@ -147,49 +147,53 @@ function clone_theme( $theme, $screenshot ) {
...
 
-		// Add / replace screenshot.
-		if ( $this->is_valid_screenshot( $screenshot ) ) {
-			$zip->addFile(
-				$screenshot['tmp_name'],
-				'screenshot.png'
-			);
-		}
+			// Add theme templates including user modifications.
+			Theme_Templates::add_templates_to_dest( $export_type, $cloned_theme_dir, $theme['slug'] );
```
mikachan: I'm seeing the following warning from this line:

`Notice: Undefined variable: export_type in /class-create-block-theme-admin.php on line 179`

Maybe this variable should be passed into `clone_theme`?

mikachan (commented):
I've tested this out using the latest version of GB with Archeo and Iotix.

> Verify the cloned theme appears as expected in Appearance > Themes

The cloned theme appears with the correct edits (screenshot, name) in Themes  ✅

However, when I view the cloned theme in the Site Editor, I'm not seeing most of the content:

image

I _think_ this is because of the way I named the theme. I was testing with Archeo here and I named the clone, "Archeo clone 2". The textdomain has come through as `archeoclone2clone2`, and the template parts and patterns are using`archeoclone2` as a theme slug. I'm guessing this mismatch has caused these references to fail.

I also tried with a single word name, "ClonedArcheo", but the textdomain came through as `clonedclonedarcheo`, so I saw the same issue with the parts and patterns.

> Verify that the export and overwrite options still work as expected

Yes! ✅

mikachan:
> If we like this idea, I think we should add the ability to edit the metadata to the Export option as a fast follow up

I think this is a great idea!

admin/create-theme/theme-zip.php
```diff
@@ -23,11 +23,16 @@ public static function add_theme_json_to_zip( $zip, $export_type ) {
 		return $zip;
 	}
 
-	public static function copy_theme_to_zip( $zip, $new_slug, $new_name ) {
-
+	public static function copy_theme_to_dest( $dest, $new_slug, $new_name ) {
```
matiasbenedetto: Could we move this method outside the Theme_Zip class ? The names do make not much sense.

matiasbenedetto (commented):
Hi :wave: 
This looks very good :)

Could we add some theme name/slug validation? I ask this because if, for example, there is already a theme with that name in the themes folder, the plugin fails silently. Let's say I export a "TT3" clone with the name "hola," and after that, I activate the "Disco" theme, and I clone it with the same name ("hola"). The resulting folder /themes/hola have the TT3 content.

We could get a list of the themes/directories in the themes folder to know if the name/slug is valid.

Apart from that, having the option to edit the theme data when you export seems important; otherwise, the user could not get a zip of a theme with a name and data.

admin/create-theme/theme-zip.php
jffng: Yes I had considered to move it to the `Theme_Utils`, but wanted the diff to be cleaner to get feedback first. What do you think about moving it to `Theme_Utils`?

matiasbenedetto:
Using this branch, I get 0-byte zip files when using the "export" option.

admin/create-theme/theme-zip.php
matiasbenedetto: Sounds good to me :+1:

admin/class-create-block-theme-admin.php
jffng: Good call, added in de3b091

admin/create-theme/theme-zip.php
jffng: Done in [9534830](https://github.com/WordPress/create-block-theme/pull/294/commits/9534830ce3cf87aabcb23262a553408597baa6b5)

jffng:
Thanks for the reviews @mikachan @matiasbenedetto !

> I think this is because of the way I named the theme. 

@mikachan you're right, there was a bug in the `replace_namespace` function — it did not handle "overlapping" cases e.g. where the old name is present in the new one. I think I fixed it [here](https://github.com/WordPress/create-block-theme/pull/294/commits/9534830ce3cf87aabcb23262a553408597baa6b5), but it could use another test and review because my regex is not very specific.

> Using this branch, I get 0-byte zip files when using the "export" option.

This should be fixed with the latest commit [f4c3f2c](https://github.com/WordPress/create-block-theme/pull/294/commits/f4c3f2c43ad1cdc7c0b93223e7c85a51148ec9a5)

jffng:
> Could we add some theme name/slug validation?

Good idea, I will add this to the PR.

mikachan:
> there was a bug in the replace_namespace function

With the very latest commit, I see this error when I try to clone: 'Theme name already exists, please choose a different name.'

If I use https://github.com/WordPress/create-block-theme/pull/294/commits/f4c3f2c43ad1cdc7c0b93223e7c85a51148ec9a5, it looks like the name of the theme does prevent it from being cloned, but I'm struggling to find a pattern. I'm using Archeo again, and these names didn't seem to work: 'ArcheoClone', 'Clone Archeo', 'New Archeo'. But these did: 'Archeos', 'Theme', 'New Theme', 'Archeo2'.

jffng:
Okay I rebased this PR. Since #292 was merged, it made sense to also solve #186 at the same time. 

> With the very latest commit, I see this error when I try to clone: 'Theme name already exists, please choose a different name.'

Sorry there was a bug in this, I think I fixed it in [c1797be](https://github.com/WordPress/create-block-theme/pull/294/commits/c1797bebb8645d103646b28d612d75481c27f948).

If you wouldn't mind testing the following flows again, that would be greatly appreciated:

- Clone
- Export
- Export from the site editor

mikachan (commented):
I've tested with Archeo again. I've added a custom image to the home template and changed the headings in global styles to a different colour (red).

**Clone:**

I cloned the theme and named it 'Oak':
- I get a cloned theme with the correct name
- The change to global styles is correctly included in the clone

I'm seeing intermittent problems with the custom image. Sometimes the image is included in the cloned theme assets folder, but on other attempts, the image is added to the existing version of Archeo (e.g. when I check my themes directory with `git status`).

I'm not sure what's triggering the differences - the only difference in my testing is different theme names, but I've seen both happen on the same theme name.

**Export:**

- Works well if I don't change the name of the theme
- If I change the name from Archeo (I tried 'Export', 'Oak', 'Test') I get the following warning from the `add_media_to_zip` function:

`Warning: file() expects parameter 1 to be a valid path, object given in /admin/create-theme/theme-zip.php on line 106`

Edit: I've just been testing out something else and I'm still seeing this warning, so I don't think it's related to this PR. It seems to consistently happen with Archeo but not any other theme - so I've either messed up my local version of Archeo or there's a bug in that theme...

**Export from the site editor:** 

- Works well 👍
- I get a zip file that includes my custom image and my change to global styles

I'm sorry that I'm still seeing errors with export - does it sound like I'm testing correctly? I feel like I'm doing something particularly awkward. Let me know if you need more info!

jffng:
Thanks for the thorough test and reporting @mikachan ! I am not sure what's going on with the **Export** there, I tried with Archeo and it seems to be working after I rebased this branch.

What is not working for me is the Export from Site Editor option. I narrowed it down fail when I include a media asset in the template, and the [`download_url`](https://github.com/WordPress/create-block-theme/blob/trunk/admin/create-theme/theme-zip.php#L168) fails. Any ideas? cc @matiasbenedetto who recently worked on a bug with this I think.

vcanales:
> What is not working for me is the Export from Site Editor option. I narrowed it down fail when I include a media asset in the template, and the [download_url](https://github.com/WordPress/create-block-theme/blob/trunk/admin/create-theme/theme-zip.php#L168) fails.

Clone from the Site Admin has a similar issue; the resulting theme does not include custom fonts, although it fails silently.

mikachan:
> I narrowed it down fail when I include a media asset in the template, and the [download_url](https://github.com/WordPress/create-block-theme/blob/trunk/admin/create-theme/theme-zip.php#L168) fails.

This sounds like the same error I was seeing with `add_media_to_zip` above, but I'm pretty sure it's not related to this PR as I was seeing the same error on different branches.

vcanales:
I created a PR with fixes for the Export/download_url problem. https://github.com/WordPress/create-block-theme/pull/338

vcanales:
>I created a PR with fixes for the Export/download_url problem. https://github.com/WordPress/create-block-theme/pull/338

Forgot to mention; as @mikachan said, it doesn't seem to be related to this PR. I tested it both on trunk and here, and it looks good.

jffng:
Closing this one as outdated.

PR #296, opened 2023-04-01
Add current WordPress version to style.css and readme.txt

mikachan:
The plugin currently uses a static version of WordPress in the style.css and readme.txt files for the "Tested up to" version. This PR changes this so the current WordPress version is used instead.

I've used [`get_bloginfo( 'version' )`](https://developer.wordpress.org/reference/functions/get_bloginfo/) to grab the current version.


jffng (approved):
Nice, makes sense to me.

PR #297, opened 2023-04-01
Add labels around Google font family checkbox controls

mikachan:
This adds labels around the text related to the Google font checkboxes when adding Google fonts to a theme. This makes the options slightly easier to select, as you can now click any text in the corresponding table row to select a font family or variant.

### Video
https://user-images.githubusercontent.com/1645628/229313331-e5539554-45e4-4c7d-bf5d-c3fe5a30d9c9.mov

### DIFF
--- a/src/google-fonts/font-variant.js
+++ b/src/google-fonts/font-variant.js
@@ -32,21 +32,31 @@ function FontVariant( { font, variant, isSelected, handleToggle } ) {
 			} );
 	}, [ font, variant ] );
 
+	const formattedFontFamily = font.family.toLowerCase().replace( ' ', '-' );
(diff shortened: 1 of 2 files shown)


jffng (commented):
This works as advertised! Code looks good, I just wonder about one thing UX wise — does it make sense that when you click on the row headers ("Weight", "Style", "Preview") it selects all the variants?

mikachan:
> does it make sense that when you click on the row headers ("Weight", "Style", "Preview") it selects all the variants?

I also wasn't too sure about this, but I went with it because these labels are next to the 'select all' checkbox, so they're essentially the labels for it. I don't feel too strongly about it though. They may easily be clicked by accident too.

jffng (approved):
Cool, let's go with it then!

Issue #195, opened 2023-01-30
Consider adding input for theme credits


jffng:
It might be useful to add a form input so users can append credits (for fonts, images, etc) or other information to the readme when they export the theme.

- [x] Handle license info for Google fonts: https://github.com/WordPress/create-block-theme/pull/298
- [x] Handle license info for local fonts https://github.com/WordPress/create-block-theme/issues/348
- [x] Handle license info for media https://github.com/WordPress/create-block-theme/pull/380

mikachan:
I think we can close this now as both fonts and images can now be credited from within the plugin.

PR #298, opened 2023-04-01
Automatically add font license info for Google fonts

mikachan:
This adds a function that automatically handles adding the font license information for each Google font to the theme's readme.txt file.

image

Part of https://github.com/WordPress/create-block-theme/issues/195.


admin/class-manage-fonts.php
```diff
@@ -296,7 +305,82 @@ function ( $font_family ) use ( $font_slug ) {
...
+
+				// Build font source URL
+				$font_source = 'https://www.google.com/fonts/specimen/' . $google_formatted_font_name;
 
+				// Get license info from font file
+				$font = \FontLib\Font::load( get_stylesheet_directory() . '/assets/fonts/' . $file_name );
+				$font->parse();
+				$license_info = $font->getNameTableString( 13 ) ? $font->getNameTableString( 13 ) : '';
+
+				if ( stripos( $font->getNameTableString( 13 ), 'SIL Open Font License' ) ) {
```
matiasbenedetto: Could get all the information from the file metadata instead of searching for specific things?

matiasbenedetto (commented):
Thanks for working on this! Very useful enhancement! :heart_eyes: 

My main concern is this: We already have a client-side library to read the font metadata. Do you think it is worth bundling another entire library for the backend? 

I don't have an answer to that question, but maybe this could help to decide:
Would the use of just the client-side library to extract the metadata increase or decrease code complexity?

matiasbenedetto:
When I add a font, 3 deprecation notices are displayed:

```
Deprecated: file_get_contents(): Passing null to parameter #4 ($offset) of type int is deprecated in create-block-theme/includes/FontLib/Font.php on line 31

Deprecated: Function strftime() is deprecated in /create-block-theme/includes/FontLib/BinaryStream.php on line 285

Deprecated: Function strftime() is deprecated in create-block-theme/includes/FontLib/BinaryStream.php on line 285

```

matiasbenedetto:
Not all the fonts available on Google Fonts are licensed under SIL license, there are a few licensed under Apache or Ubuntu licenses. If I add for example the font called Arimo, it adds a SIL license notice when Arimo uses Apache licensing. https://fonts.google.com/attribution

Example:
```
Arimo Font
Licensed under the SIL Open Font License, Version 1.1 (http://scripts.sil.org/OFL)
Source: https://www.google.com/fonts/specimen/Arimo
```

matiasbenedetto:
Should we add the copyright notice and the author's name to be compliant with the attribution part of the SIL license? I guess that it won't hurt and we can we that info [from the file metadata](https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6name.html). In this way, we could make this work for local font files too.

Apparently, we need to add the copyright notice too: https://themes.trac.wordpress.org/ticket/136638#comment:2

admin/class-manage-fonts.php
```diff
@@ -296,7 +305,82 @@ function ( $font_family ) use ( $font_slug ) {
...
+
+		// If file_name exists, then add font license to readme.txt
+		if ( 'remove' !== $file_name && is_string( $file_name ) ) {
+			// Check that the font is not already credited in readme.txt
+			if ( false === stripos( $readme_file_contents, $font_name ) ) {
+				// Require php-font-lib
+				require_once( __DIR__ . '/../includes/FontLib/Autoloader.php' );
+
+				// Build font source URL
+				$font_source = 'https://www.google.com/fonts/specimen/' . $google_formatted_font_name;
```
matiasbenedetto: Could we get rid of this and use the info from the font file asset metadata? In this way we could use this function for local font files too.

composer.json
```diff
@@ -32,12 +32,16 @@
...
+		"phenx/php-font-lib": "^0.5.4"
 	},
 	"scripts": {
 		"format": "phpcbf --standard=phpcs.xml.dist --report-summary --report-source",
 		"lint": "phpcs --standard=phpcs.xml.dist",
 		"test": "phpunit",
-		"test:watch": "phpunit-watcher watch < /dev/tty"
+		"test:watch": "phpunit-watcher watch < /dev/tty",
+		"post-autoload-dump": [
+			"cp -r vendor/phenx/php-font-lib/src/FontLib includes"
```
matiasbenedetto: I'm not sure what is this doing. Could you explain it, please :pray: ?

admin/class-manage-fonts.php
```diff
@@ -296,7 +305,82 @@ function ( $font_family ) use ( $font_slug ) {
...
+				$font = \FontLib\Font::load( get_stylesheet_directory() . '/assets/fonts/' . $file_name );
+				$font->parse();
+				$license_info = $font->getNameTableString( 13 ) ? $font->getNameTableString( 13 ) : '';
+
+				if ( stripos( $font->getNameTableString( 13 ), 'SIL Open Font License' ) ) {
+					// If license is SIL Open Font License, use custom format
+					$license_info = 'Licensed under the SIL Open Font License, Version 1.1';
+				}
+
+				$license_url = $font->getNameTableString( 14 ) ? "({$font->getNameTableString(14)})" : '';
```
madhusudhand: Should we include author information to give them credits?

image

madhusudhand (commented):
update `gitignore` to exclude `includes` dir?

admin/class-manage-fonts.php
mikachan: I'm not sure what you mean here, could you give me an example? This is to check if the SIL license is being used, and then replace the information from the file with a custom message so it's more consistent with the other types of license.

composer.json
mikachan: This is copying the FontLib files to the `includes` directory, as this seemed like a good way to include the library and expose the functionality. It's the first time I've needed to include an external library like this though, I'm not sure if this is the best way!

mikachan:
> My main concern is this: We already have a client-side library to read the font metadata. Do you think it is worth bundling another entire library for the backend?

I'd rather not include a whole additional library that does the same thing, so I'd like to try and use the client-side library here instead. Hopefully it doesn't increase the code complexity too much.

admin/class-manage-fonts.php
mikachan: Yes this is a good idea, I'll try adding it from the metadata!

mikachan:
> Not all the fonts available on Google Fonts are licensed under SIL license, there are a few licensed under Apache or Ubuntu licenses. If I add for example the font called Arimo, it adds a SIL license notice when Arimo uses Apache licensing.

This was the main reason I added the FontLib library 😅 Hmm, I was testing this using the Aclonica font, and the Apache license was coming through correctly from the metadata. Maybe this information isn't always correct with Google fonts... I'll do some more testing around this. Maybe it'll work better with the other library.

admin/class-manage-fonts.php
mikachan: Yes good idea, I was hoping to extend this to local fonts once I got this version working ✨

admin/class-manage-fonts.php
matiasbenedetto: I mean that we can extract the entire license information from the metadata tables of the font file asset. 
Example:
image

This online font metadata inspector could be helpful in looking for the name of the metadata property: https://opentype.js.org/font-inspector.html

mikachan: Ah yes, that's what I'm using the FontLib library for here, but the information for the SIL license is a lot longer and looks different from the usual descriptions we use for the fonts in WP themes. So I decided to add a check for the SIL license and then replace it with information in our usual format.

mikachan:
Thank you @matiasbenedetto @madhusudhand for the reviews!

I've updated the PR with the following:

- Tried out using the client-side font library to read the font files and removed the PHP font library
- It's now passing the font credit info from the font metadata through the Google font form, along with all the other existing font data (name, weight etc)
- Included the copyright info
- It's now using the source URL from the font metadata
- Added an 'ending tag' to the font credit info in the readme.txt, as this makes it a lot easier to remove the credits when the font is removed from the theme

I think this covers the feedback so far. I prefer using a single font library!

matiasbenedetto (commented):
Adding some particular fonts, for example `Abyssinica SIL`, the entire SIL license text is printed in the `readme.txt` file. Is this expected?

image

mikachan:
Thanks for testing again!

> Is this expected?

Ah, no! I've not seen that before. I'll look into how we can handle that.

mikachan:
I think I've improved this with a bit of a refactor. 

- I've changed this so it's using the `get()` function to get the data from the font file, and this seems more reliable than before.
- I've limited the license info content to 200 characters, as this seemed reasonable. The full license can always be read at the license URL, which we're also including.

matiasbenedetto:
> I've limited the license info content to 200 characters

I haven't read the code yet, but can it be a good idea to limit it to one line instead of counting the characters?

mikachan:
> but can it be a good idea to limit it to one line instead of counting the characters?

I considered this too, and couldn't decide which one was best! I've updated it so it will now limit the content to the first line, and still limit it to 200 characters as well in case there's one long single-line string.

matiasbenedetto (changes requested):
I found a little problem: let's say a family has several font faces; if you delete just one font face, but the rest of the family is there, the license is gone from `readme.txt` . Apart from that this PR seems almost ready to go!

Github video upload is not working for me today so I upload the screencast of featuring the issue here:
https://i.imgur.com/nNjO55W.mp4

mikachan:
Good spot, @matiasbenedetto! I've fixed this by moving the 'remove' functionality to the `prepare_font_families_for_database` function, so it can be run based on the font family logic rather than the font face. Thanks!

matiasbenedetto (approved):
It seems to be working great now. Thanks for this enhancement. Very useful.

As a follow up, I created an issue to extend this functionality to the local font file assets uploaded: https://github.com/WordPress/create-block-theme/issues/348

LGTM :rocket:

PR #304, opened 2023-04-05
Add theme name validation

### LINKED ISSUE #299: Check for available theme name
It would be such a time saver if the theme name could be checked for availability when you enter the name into the theme name input field.

- It should check against the Theme directory svn first, since this is doable today.

matiasbenedetto:
## What?
Add theme name validation.
The user won't be able to create a theme using a name if it:
- Already exists in the WordPress.org theme directory (#299)
- Doesn't follow the naming guidelines: https://meta.svn.wordpress.org/sites/trunk/wordpress.org/public_html/wp-content/plugins/theme-directory/class-wporg-themes-upload.php

##  Why?
To early warning theme creators about not valid theme names

## Screencast

https://user-images.githubusercontent.com/1310626/230172711-a1e8a747-3b3d-4e98-b08b-8e7ed945d45e.mp4

Fixes https://github.com/WordPress/create-block-theme/issues/299

### DIFF
--- a/admin/create-theme/theme-form.php
+++ b/admin/create-theme/theme-form.php
@@ -131,7 +131,7 @@ class Theme_Form {
 								<p><em><?php _e( 'Items indicated with (*) are required.', 'create-block-theme' ); ?></em></p>
 								<label>
 									<?php _e( 'Theme Name (*):', 'create-block-theme' ); ?><br />
-									<input placeholder="<?php _e( 'Theme Name', 'create-block-theme' ); ?>" type="text" name="theme[name]" class="large-text" />
+									<input placeholder="<?php _e( 'Theme Name', 'create-block-theme' ); ?>" type="text" name="theme[name]" class="large-text" id="theme-name" autocomplete="off" />
 								</label>
 								<br /><br />
 								<label>
--- a/admin/js/form-script.js
+++ b/admin/js/form-script.js
@@ -143,17 +143,10 @@ const ERROR_NAME_CONTAINS_WORDPRESS = __(
 	'Theme name cannot contain the word "WordPress"',
 	'create-block-theme'
 );
-const ERROR_NAME_CONTAINS_MARKUP_WORDS = __(
-	'Theme name cannot contain the words as "HTML", "CSS", or "PHP"',
-	'create-block-theme'
-);
-const ERROR_NAME_CONTAINS_RELATED_WORDS = __(
-	'Theme name cannot contain the words as "Blog", "Web log", "Template", "Skin", etc.'
-);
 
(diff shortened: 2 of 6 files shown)


carolinan:
I don't think those guidelines from 2013 are valid anymore. It may be easiest to use the same "blockers" as the uploader in the directory?

matiasbenedetto:
> It may be easiest to use the same "blockers" as the uploader in the directory?

I have made a quick search and could not find those blockers. Apart from this https://make.wordpress.org/themes/handbook/review/required/#7-naming-spelling-and-trademarks are there any other guidelines that still apply?

@carolinan could you share a link, please?

carolinan:
https://meta.svn.wordpress.org/sites/trunk/wordpress.org/public_html/wp-content/plugins/theme-directory/class-wporg-themes-upload.php

matiasbenedetto:
> https://meta.svn.wordpress.org/sites/trunk/wordpress.org/public_html/wp-content/plugins/theme-directory/class-wporg-themes-upload.php

Thanks!

I removed the old naming restrictions from validation.
The restrictions left are just "WordPress" and "Theme" words can't be part of the theme name. And the theme name should not be already in the theme directory.

madhusudhand:
> And the theme name should not be already in the theme directory.

I don't think this should be a hard restriction to create a theme. 

Given that plugin can be used to create themes without restrictions and not every theme is published to `.org`, this check should only be informative not a requirement.

Anyone should still be able to create a theme with any name of their choice and published it to their own platform.

carolinan:
The main reason is that if the theme on wordpress.org has a higher version number, the site will get notifications that there is an update, but it won't be the same theme. This is very annoying and can be devastating if you don't know about it.

carolinan:
If we lift the restriction, I think those of us who do submit to .org would like a notice or warning at least.

jffng (approved):
This is working as advertised so I'm giving the ✅ 

There are a couple ideas I had for improving the UX a bit but these could be done in a follow up if folks agree:

1. it could be nice to persist the warning if you navigate away from the input
2. supply a link to the documentation / theme name guidelines if a warning is issued

Regarding 1: 

Warning is shown | Warning disappears when you navigate away
--- | ---
Screenshot 2023-04-27 at 9 41 17 AM | Screenshot 2023-04-27 at 9 41 19 AM

matiasbenedetto:
> This is working as advertised so I'm giving the white_check_mark
> 
> There are a couple ideas I had for improving the UX a bit but these could be done in a follow up if folks agree:

Thanks for testing!
Yes, feel free to open an issue or a PR to improve that. Seems like a good enhancement.

PR #305, opened 2023-04-06
Replace theme slug in templates after getting media urls from them

matiasbenedetto:
## What?

Replace theme slug in templates after getting media urls from them to avoid non-existing urls.

## Why?
If we replace the slug before getting the media URLs and those urls contain the old slug name, we get non-existing URLs.

## Thanks!

Thanks, @henriqueiamarino for reporting this issue.
Thanks, @vcanales for pairing on this!


vcanales (approved):
LGTM!

PR #310, opened 2023-04-11
Fixing error checking

### LINKED ISSUE #309: The media bundling functionality is not creating the assets when running overwrite theme option
If you want to overwrite your theme, the media bundler is not creating the assets in your theme assets folder. In the conversion from a template to a pattern works but the media assets are not written to disk.

Raised originally in this forum thread: https://wordpress.org/support/topic/issue-with-exporting-images/

matiasbenedetto:
In the function `add_media_to_local,` we checked for an error to try to write the asset when we should check the opposite.

The problem was introduced in this recent PR: https://github.com/WordPress/create-block-theme/pull/305

Fixes #309

### DIFF
--- a/admin/create-theme/theme-media.php
+++ b/admin/create-theme/theme-media.php
@@ -139,7 +139,7 @@ class Theme_Media {
 		foreach ( $media as $url ) {
 			$download_file = download_url( $url );
 			// TODO: implement a warning if the file is missing
-			if ( is_wp_error( $download_file ) ) {
+			if ( ! is_wp_error( $download_file ) ) {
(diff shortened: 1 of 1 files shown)


jffng (approved):
Here's what I did to test:

- Added image to header
- Used the "Overwrite" option
- Verified that the image was written and appeared as expected

Approving since it fixes that issue. Though the original [wp forum](https://wordpress.org/support/topic/issue-with-exporting-images/) report suggested that it wasn't working on "Clone" — I tried on Clone and that options seems to work correctly (writes the asset) with and without this fix.

PR #315, opened 2023-04-12
Avoid adding Template info to style.css if it's empty

### LINKED ISSUE #314: Template: in style.css header of parent theme
I generated a parent theme using the plugin and I found, `Template:` empty in the style.css header. That template is only for child themes. Can we remove that from the parent theme and only use in the child theme? The empty Template: will create confusion for the user.

image

matiasbenedetto:
Avoid adding Template info to style.css if it's empty

Fixes: #314


admin/create-theme/theme-styles.php
```diff
@@ -29,7 +30,7 @@ public static function build_child_style_css( $theme ) {
 Version: 0.0.1
 License: GNU General Public License v2 or later
 License URI: http://www.gnu.org/licenses/gpl-2.0.html
-Template: {$template}
+{$template_text}
```
madhusudhand: It still keeps a blank line

image

madhusudhand (commented):
Also, it is preserved in create blank theme option (create blank theme -> activate -> export)

admin/create-theme/theme-styles.php
matiasbenedetto: fixed in the latest commit.

matiasbenedetto:
Thanks for testing! The new line should is now avoided after the latest commit.

admin/create-theme/theme-styles.php
```diff
@@ -29,10 +30,17 @@ public static function build_child_style_css( $theme ) {
 Version: 0.0.1
 License: GNU General Public License v2 or later
 License URI: http://www.gnu.org/licenses/gpl-2.0.html
-Template: {$template}
-Text Domain: {$text_domain}
+";
+
+		if ( ! empty( $template ) ) {
+			$style_css .= "Template: {$template}";
```
jffng: I tried creating a child theme, and the output is missing a line break:

Screenshot 2023-04-14 at 12 44 33 PM

Should we add one?

```suggestion
			$style_css .= "Template: {$template}\n";
```

matiasbenedetto: Thanks for the catch!

jffng (commented):
This is now working as advertised, but why is the diff so large with `fallback-fonts-list.json`?

mikachan (approved):
Looks like the changes to `fallback-fonts-list.json` should just be removed from this PR, but otherwise this looks good to bring in.

matiasbenedetto:
Thanks for taking a look!

> Looks like the changes to `fallback-fonts-list.json` should just be removed from this PR

That was solved by merging with the latest trunk version.

PR #328, opened 2023-04-17
Fix delete font family/face when name is different from family

matiasbenedetto:
## What?
A fix for deleting font family and font faces.

## Why?
When the font name is different from the font family the delete function is not working.
With this change, we look for both name and family.

## How to test:
1. Activate a theme including fonts with different name that fontFamily as `Disco` theme.
2. Try to remove the font using the font manager screen and see that it is failing silently.

Example from disco:
```
 "fontFamilies": [
				{
					"fontFamily": "SpaceMono, serif",   <--------------------
					"name": "SpaceMono", <-----------------------------------
					"slug": "spacemono",
					"fontFace": [
						{
							"fontDisplay": "block",
							"fontFamily": "SpaceMono",
							"fontWeight": "400",
							"fontStyle": "normal",
							"fontStretch": "normal",
							"src": [
								"file:./assets/fonts/SpaceMono-Regular.ttf"
							]
						},
...
```

### DIFF
--- a/src/manage-fonts/index.js
+++ b/src/manage-fonts/index.js
@@ -84,7 +84,10 @@ function ManageFonts() {
 
 	function deleteFontFamily( fontFamily ) {
 		const updatedFonts = newThemeFonts.map( ( family ) => {
-			if ( fontFamily === family.fontFamily ) {
+			if (
+				fontFamily === family.fontFamily ||
+				fontFamily === family.name
+			) {
 				return {
 					...family,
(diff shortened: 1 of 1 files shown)


jffng:
Can you provide an example to test this PR with?

matiasbenedetto:
> Can you provide an example to test this PR with?

Yes! Sorry for not including them earlier. I updated the PR description with test instructions.

jffng (approved):
Works as advertised!

PR #347, opened 2023-05-02
Fix console error in `prepareThemeNameValidation` function

mikachan:
This fixes a small console error triggered from the `prepareThemeNameValidation` function, by checking for the `themeNameInput` element before adding the event listener.

Here's an example of the error, which is currently showing on WP admin pages:

image

To test, check that the above error is no longer shown and that the `prepareThemeNameValidation` function still runs correctly when the `themeNameInput` element exists (i.e. where you can edit the theme name for a generated theme at `/wp-admin/themes.php?page=create-block-theme`).


madhusudhand (approved):
LGTM!

PR #357, opened 2023-05-09
Adding files to zip subfolder called as theme slug

matiasbenedetto:
## What?
Create theme files inside a subfolder called as the theme slug.

## Example of zip file structure: 
**Before:**
- style.css
- screeenshot.png
- templates

**After:**
- themename/style.css
- themename/screeenshot.png
- themename/functions.php  

## Why?
Avoid problems with theme paths when exporting/cloning themes as zip files. 

For example, when you download the theme zip file with a filename different from the expected theme slug. It's common to download files as 'themename (5).zip'. Let's see what happen if you upload that file to a WordPress instance: the theme will live in  `wp-content/themes/themename-5` instead of `wp-content/themes/themename` meanwhile the theme slug will continue being `themename`. There could be references to the theme slug in the theme, for example, if your theme is linking template parts in patterns. The end result will be missing template parts because the parts will be searched in `wp-content/themes/themename/parts` when they will be located in `wp-content/themes/themename-5/parts`.

With this change no matter how is the zip file called. The theme will be created in `wp-content/themes/themename`  so we will avoid the errors described before.


mikachan (commented):
I get quite a few warnings in WP admin when I apply this branch:

`Warning: Declaration of CbtZipArchive::addEmptyDir(string $dirname, int $flags = 0) should be compatible with ZipArchive::addEmptyDir($dirname) in /admin/create-theme/cbt-zip-file.php on line 24`

`Warning: Declaration of CbtZipArchive::addFromString(string $name, string $content, int $flags = ZipArchive::FL_OVERWRITE) should be compatible with ZipArchive::addFromString($name, $content) in /admin/create-theme/cbt-zip-file.php on line 14`

`Warning: Declaration of CbtZipArchive::addFile(string $filepath, string $entryname = '', int $start = 0, int $length = 0, int $flags = ZipArchive::FL_OVERWRITE) should be compatible with ZipArchive::addFile($filepath, $entryname = NULL, $start = NULL, $length = NULL) in /admin/create-theme/cbt-zip-file.php on line 19`

And a couple of: `Warning: Cannot modify header information - headers already sent...` warnings.

It looks like `CbtZipArchive` doesn't quite match the `ZipArchive` class.

admin/create-theme/cbt-zip-file.php
```diff
@@ -0,0 +1,31 @@
+<?php
```
jffng: What do you think about naming this file `cbt-zip-archive` instead?

jffng (commented):
I got the same warnings as @mikachan.

admin/create-theme/cbt-zip-file.php
matiasbenedetto: Yes, updated!

matiasbenedetto:
> I get quite a few warnings in WP admin when I apply this branch:
> It looks like `CbtZipArchive` doesn't quite match the `ZipArchive` class.

It seems like overriding the ZipArchive class methods is not the best idea because the declaration of the methods is not exactly the same in all PHP versions. To work around this, I called the functions we are using with another name to avoid conflicts. It should work as expected now.

jffng (commented):
Here's a workflow that I tried that doesn't seem to be working as expected:

- Export "Adventurer" (a theme that lives in `wp-content/themes/themes/adventurer` 
- There already exists `adventurer.zip` in my downloads, so the export is named `adventurer (2).zip`
- Upload that theme zip to the same site and receive the following error:

```
The package could not be installed. The theme is missing the style.css stylesheet.

Theme installation failed.
```

When I expand the zip locally, it results in a folder `themes` containing a subfolder `adventurer`.

matiasbenedetto:
> Here's a workflow that I tried that doesn't seem to be working as expected:

Thanks for testing @jffng ! I could not reproduce this issue even with `adventurer` or any other theme I tried.

jffng:
@matiasbenedetto what is the path of the themes you are testing with? I did some more testing and get the error for exported themes that live in a subfolder within the themes folder, e.g. `wp-content/themes/community-themes/stacks`.

matiasbenedetto:
@jffng The problem was only reproducible when the original theme came from a subfolder. In that case, the slug we get is a path instead of just the slug of the theme (ex: `subfolder/themeslug` instead of `themeslug`). Because of this, the files created in the zip file were added in this path `subfolder/themeslug/file.html` instead of `themeslug/file.html` that was causing the errors. With the latest commit that is fixed, and the zip files are created in `themeslug/...`.

mikachan (approved):
This is working well for me now using themes in subfolders. I followed @jffng's testing steps from [here](https://github.com/WordPress/create-block-theme/pull/357#pullrequestreview-1453429359):

> Export "Adventurer" (a theme that lives in wp-content/themes/themes/adventurer) ✅
> Upload that theme zip to the same site, no errors and theme activates successfully ✅
> Theme installation works ✅
> Expanded the zip locally, I can see the theme files in the root of the zip folder ✅

I think this is good to come in now 👍

PR #364, opened 2023-05-10
Added 'Update' and 'Create' panels to work with Revisr plugin flow experiments

pbking:
Moved 'export' to a panel and added 'update' and 'create' panels.

Done mostly to introduce a way to move a theme from one directory to another: 

image

Done while working on "Revisr" workflows, in an effort to move an already-created theme to a directory that's under version control (such as a `/pub` folder).


pbking:
This was a spike to test things.
Implemented elsewhere, closing this.

PR #369, opened 2023-05-15
Add linebreaks before hyphen lists in readme to fix plugin repository display

mrwweb:
The readme.txt omits line breaks before lists made with hyphens. This prevents the lists from correctly being rendered as HTML lists in the wp.org plugin repository.

This PR adds those linebreaks to fix this issue.

-----

Readme preview courtesy of WPReadme.com

Before:

The start of the plugin's readme uses the hyphen character for a list, and it's not converted to a real HTML list

After:

The list from the previous image now uses real bullets and has an empty line before it

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -13,6 +13,7 @@ A WordPress plugin to create block themes.
 == Description ==
 
 This plugin allows you to:
+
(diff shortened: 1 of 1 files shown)


mikachan (approved):
Nice, thank you!

PR #373, opened 2023-05-16
Prevent additional white space in font credits in readme.txt

### LINKED ISSUE #361: Font removal: additional spaces are left in readme.txt from font license
When a font family is removed from a theme, there are additional spaces left behind from the font license in the readme.txt file:

image

We should ensure that the additional spaces are also removed when the font license content is removed.

Originally mentioned here: https://github.com/WordPress/create-block-theme/pull/359#pullrequestreview-1420424962

mikachan:
When the font credits are added to the readme.txt file, they add an unnecessary line break above the font license info. This PR should fix this so that a new line break is only added when the readme.txt does not already have a line break at the end of the file. This line break above the font credits block should also be removed from the readme.txt when a font family is removed from the theme.

Testing instructions:

1. Go to the Manage Theme Fonts page
2. Add a new font
3. Check that the font license info has been added to the theme readme.txt and that there is a line break between the previous end of the readme contents and the new font credits
4. Go back to the Manage Themes Fonts page
5. Remove a font family
6. Check that the font license info has been removed from the readme.txt and there are no unnecessary line breaks left behind

Closes https://github.com/WordPress/create-block-theme/issues/361.

### DIFF
--- a/admin/class-manage-fonts.php
+++ b/admin/class-manage-fonts.php
@@ -355,11 +355,15 @@ class Manage_Fonts_Admin {
 
 				// Build the font credits string
 				$font_credits = "
-
 {$font_name} Font
 {$copyright} {$license_info} {$license_url} {$font_source}
 {$end_credits_note}
 ";
+
+				// Check if readme.txt ends with a new line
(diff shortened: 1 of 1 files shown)


vcanales (approved):
Works as advertised; LGTM.

PR #374, opened 2023-05-16
Escape special characters to avoid syntax errors

### LINKED ISSUE #349: Apostrophe in alt text crashes site with critical error
## How to reproduce the bug
1. Start with the Twenty Twenty Three theme.
2. Add an image to the Homepage template through the Editor in an Image block.
3. Include an apostrophe in the alt text of that image.
4. Export the theme with the Create Block Theme plugin. (Issue occurs when the theme is exported either as a parent or child theme.)
5. Import the theme into a new site.
6. Activate the theme, and the site will crash with a critical error.

## Error message

```
Parse error: syntax error, unexpected identifier "s", expecting ")" in .../wp-content/themes/theme-name/patterns/home.php on line ###
There has been a critical error on this website. Please check your site admin email inbox for instructions.
```

## Environment
I've been able to reproduce this both on a live environment, and locally.

vcanales:
Special characters were causing patterns to crash the site due to syntax errors. In this change, we make sure that these characters are escaped before being printed to the pattern.

Tested using the string "ABC/'"-.`\';//\\''``" as alt text for a template image.

Fixes: https://github.com/WordPress/create-block-theme/issues/349

### DIFF
--- a/admin/create-theme/theme-patterns.php
+++ b/admin/create-theme/theme-patterns.php
@@ -58,7 +58,8 @@ class Theme_Patterns {
 
 	static function escape_text_for_pattern( $text ) {
 		if ( $text && trim( $text ) !== '' ) {
-			return "<?php echo esc_attr_e( '" . $text . "', '" . wp_get_theme()->get( 'Name' ) . "' ); ?>";
+			$escaped_text = addslashes( $text );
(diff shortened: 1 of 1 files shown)


mikachan (approved):
I tested this using TT3. I inserted an image on the home template using, "ABC/'"-.`';//\''``", as the alt text. The theme was exported and activated successfully with no error messages, and I can see the image with the correct alt text.

LGTM 👍

PR #382, opened 2023-05-25
Update theme `version` logic to use isset()

mikachan:
I was seeing the error, `Undefined index: version`, when creating a zip from the wp-admin pages of the plugin. This PR should fix this error by checking if `version` exists before using the value.


pbking:
I can't confirm the error, nor understand why it would be happening.  But there's certainly no harm in bringing this in so I'm giving it a 👍 since it works either way for me.

pbking (approved):

PR #389, opened 2023-06-05
Fix double replacement in replace_namespace

### LINKED ISSUE #387: Site Editor Plugin sidebar: Clone: Theme Name is duplicated
If my theme name is for example "testar" and I go into the Site Editor plugin sidebar and select "Create theme" and then
clone, it will tell me that the theme already exists. This is fine. But when I enter a new name in the name field,
for example, "testar2", the theme ends up as "testar222"; if I set the theme name to "testarx" it ends up as "testarxxx" "testar-again" ends up as "testar-againagainagain".

The theme name presented in the UI in the Site Editor plugin sidebar (theme info panel) matches the name that ends up in the style.css file.
But it does not match the theme folder name: If the theme name is "testarenxxx" then the folder name is "testarenx".

MacOS
Local by flywheel
PHP 8.0.022
WP 6.2.2
Create block theme Version 1.12.0 
Gutenberg Version 1.12.0 

The theme that is being cloned is a "blank" theme generated through the create block theme plugin.

jffng:
Fixes #387. This PR alters the `replace_namespace` function to use placeholders before performing the actual string replacement. 

When performing multiple replacements in the same string, some of the replacements can result in a string that matches an earlier search pattern. I think this is what's happening in #387.

**To test**
Follow the instructions in 387 on trunk and this PR, verify you don't get duplicated aspects of the old theme slug in the new theme slug.

### DIFF
--- a/admin/create-theme/theme-utils.php
+++ b/admin/create-theme/theme-utils.php
@@ -30,15 +30,25 @@ class Theme_Utils {
 	}
 
 	public static function replace_namespace( $content, $new_slug, $new_name ) {
-
 		$old_slug            = wp_get_theme()->get( 'TextDomain' );
 		$new_slug_underscore = str_replace( '-', '_', $new_slug );
 		$old_slug_underscore = str_replace( '-', '_', $old_slug );
 		$old_name            = wp_get_theme()->get( 'Name' );
 
-		$content = str_replace( $old_slug, $new_slug, $content );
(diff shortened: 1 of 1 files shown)


pbking:
I was finally able to reproduce this!  

Thank you for your persistence in debuggin it @jffng and once I saw the fix I understood what I was missing in my steps to reproduce.

This fixes the issue for me.

pbking (approved):

carolinan:
Thank you!

PR #391, opened 2023-06-06
Try: Only run google font update on WordPress repo

### LINKED ISSUE #386: Consider not running cron GitHub Actions on forked repositories
As I understand it, [the GitHub Actions cron checks for Google Fonts updates](https://github.com/WordPress/create-block-theme/blob/b400afa3086d002c29221e3e388498fd64bcc9db/.github/workflows/update-google-fonts-data.yml#L9) every Tuesday and a pull request is created.

It seems that this cron runs even on forked repositories. And I am getting weekly notifications that the GitHub Actions have failed:

gmail

The results of failed GitHub Actions on a forked repository can be found [here](https://github.com/t-hamano/create-block-theme/actions).

This problem can probably be solved by changing the forked repository to not perform GitHub Actions:

disable-actions

However, I think it would make sense to add a setting in GitHub Actions itself to exclude forked repositories.

vcanales:
Checks if the repo triggering the workflow is owned by WordPress; otherwise, skip running it.

Fixes: https://github.com/WordPress/create-block-theme/issues/386

### DIFF
--- a/.github/workflows/update-google-fonts-data.yml
+++ b/.github/workflows/update-google-fonts-data.yml
@@ -13,6 +13,7 @@ on:
 jobs:
     # This workflow contains a single job called "update-google-fonts-json"
     update-google-fonts-json:
+        if: github.repository_owner == 'WordPress'
         # The type of runner that the job will run on
         runs-on: ubuntu-latest
 


pbking (approved):
Works as advertised.

Issue #395, opened 2023-06-13
500 Internal Server Error on Save Changes


eirichmond:
After editing theme templates, template parts, global styles and manual changes made in the theme.json file, I'm now getting an Internal Server Error when trying to 'Save Changes'

<img width="836" alt="Screenshot 2023-06-13 at 18 18 09" src="https://github.com/WordPress/create-block-theme/assets/525490/4af0597d-8af8-4586-b2bc-034167ab89e7">

Steps taken:
1. Create Theme
2. Makes edits in the FS UI
3. Save Changes
4. Make edits directly in the theme.json file
5. Makes edits in the FS UI
6. Save Change

For me this resulted an Internal Server Error.

carolinan:
@pbking   I wonder if this is the same error I was getting that I could not reproduce after resetting the database.

@eirichmond Have you enabled error logging and error notice displaying?
It is also possible that the manual changes to theme.json broke theme.json.

eirichmond:
@carolinan error log shows call to undefined function in theme-media.php:140

`[14-Jun-2023 05:32:40 UTC] PHP Fatal error:  Uncaught Error: Call to undefined function download_url() in /Users/elliottrichmond/valet-sites/wooblocktheme/wp-content/plugins/create-block-theme/admin/create-theme/theme-media.php:140
`

I've noticed that where `download_url()` is used elsewhere, the following is called before the function is used:
```
if ( ! function_exists( 'download_url' ) ) {
    require_once ABSPATH . 'wp-admin/includes/file.php';
}
```

So I added the above to the main create-block-theme.php file just below the `WPINC` as follows:

```
if ( ! defined( 'WPINC' ) ) {
	die;
}

if ( ! function_exists( 'download_url' ) ) {
	require_once ABSPATH . 'wp-admin/includes/file.php';
}
```

Now, the previously generated theme gets removed/deleted completely from the /themes/ folder, which in turn results in a "theme not found" error on reload.

vcanales:
Closed via https://github.com/WordPress/create-block-theme/pull/398

PR #398, opened 2023-06-14
Move check for `download_url` higher up

mikachan:
This moves the check for `download_url` into the main plugin file, `create-block-theme.php`. This means we shouldn't need to repeat this check whenever the `download_url()` function is called.

Partly addresses a comment from another issue: https://github.com/WordPress/create-block-theme/issues/395#issuecomment-1590535195


vcanales (approved):
Good point; works for me :)

mikachan:
Just pushed a commit to add @eirichmond as a co-author, as this was spun up from your comment 😊 

Will bring this in, thanks for testing @vcanales!

PR #401, opened 2023-06-15
Avoid white spaces or other weird characters on font asset paths.

matiasbenedetto:
## What ?
If you upload a font with spaces in any of the components that end up in the file name of the font asset the resulting font URL has a space and that causes problems.

**For example:** if you upload a variable font with a font weight of "100 900", the resulting font path in theme.json would be `font-name_normal_100 900.ttf` and that can be problematic.

With this change, we would have `font-name_normal_100-900.ttf`.

## How to test:
1. Upload a font with variable width. Example: [Literata variable](https://drive.google.com/file/d/1kPE9nLUOsaQlOrB_ueqOhgAUQkiRnQx8/view?usp=drive_link) (external link because GitHub doesn't allow ttf files).
2. Test that the font asset filename and font path in theme.json don't have white spaces.


mikachan (approved):
Nice, LGTM 👍

PR #402, opened 2023-06-19
Add "Recommended Plugins" input box

### LINKED ISSUE #399: Add a text input to handle recommended plugins
We should consider adding a text input to allow theme authors to list recommended plugins that complement the theme.

This is detailed in the Theme Handbook here: https://make.wordpress.org/themes/handbook/review/required/#6-plugins

We should include the restrictions close to the input:

> Themes may only recommend plugins that are hosted on WordPress.org.
> Plugins may only be installed by user action, not automatically.

Originally suggested by @jamiemarsland.

mikachan:
This PR adds a "Recommended Plugins" input to the wp-admin page and editor versions of the plugin.

Closes https://github.com/WordPress/create-block-theme/issues/399.

### Testing Instructions:
1. Open the wp-admin page or the editor sidebar panel
2. Add some text to the recommended plugins input box
3. Export the theme
4. The text should be added to the theme's readme, under a "Recommended Plugins" heading

### Screenshots:

wp-admin:
image

Editor sidebar panel:
image

### DIFF
--- a/admin/create-theme/theme-readme.php
+++ b/admin/create-theme/theme-readme.php
@@ -4,7 +4,7 @@ class Theme_Readme {
 	/**
 	* Build a readme.txt file for CHILD/GRANDCHILD themes.
 	*/
-	public static function build_readme_txt( $theme ) {
+	public static function build_readme_txt( $theme, $update_current_theme = false ) {
 		$slug                = $theme['slug'];
 		$name                = $theme['name'];
 		$description         = $theme['description'];
@@ -25,6 +25,12 @@ class Theme_Readme {
 
 		// Handle recommended plugins section.
 		$recommended_plugins_section = self::recommended_plugins_section( $recommended_plugins );
+
+		// Handle updating current theme.
+		if ( $update_current_theme ) {
+			$updated_readme = self::update_readme( $image_credits, $recommended_plugins );
+			return $updated_readme;
+		}
 
 		return "=== {$name} ===
 Contributors: {$author}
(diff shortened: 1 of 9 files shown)


pbking:
I see the additional form on the editor.  However, nothing sticks when I edit it.
image

I don't see the form on the wp-admin 'Create Blank' form.  (Though I do see it on `Clone` and `Create Child`).

The forms don't seem to add the value to the exported theme though.

mikachan:
Thanks for testing, @pbking!

> I see the additional form on the editor. However, nothing sticks when I edit it.

It looks like the editor version was only updating the style.css file, so I've updated this PR so that the readme.txt file is also updated from the editor.

> I don't see the form on the wp-admin 'Create Blank' form.

This is intentional (similar to how the image credits input works), as I was thinking that people wouldn't want to add recommended plugins to a blank theme. I've also removed the 'Recommended Plugins' input from the 'Create' sidebar panel, as I'm now thinking it's only useful on the 'Update' panel for now. We could revisit this when we align the wp-admin version and editor versions closer together.

admin/create-theme/theme-readme.php
```diff
@@ -4,37 +4,35 @@ class Theme_Readme {
...
+		$recommended_plugins = $theme['recommended_plugins'] ?? '';
+		$is_parent_theme     = $theme['is_parent_theme'] ?? false;
+		$original_theme      = $theme['original_theme'] ?? '';
+
+		// Handle copyright section.
 		$new_copyright_section  = $is_parent_theme || $original_theme ? true : false;
 		$original_theme_credits = $new_copyright_section ? self::original_theme_credits( $name, $is_parent_theme ) : '';
+		$copyright_section      = self::copyright_section( $new_copyright_section, $original_theme_credits, $name, $copy_year, $author, $image_credits );
 
-		$default_copyright_section = "== Copyright ==
```
mikachan, line 22: I've refactored this and moved it lower down so that it's contained in `copyright_section()`.

vcanales (commented):
Could we add a placeholder with an example? It's been hard to find an explanation of this section; for instance, it's not listed in the docs https://developer.wordpress.org/themes/basics/main-stylesheet-style-css/

I wondered if I should input a URL or if just the plugin's name would suffice, for example. I can also picture folks wondering if they should separate plugins with a comma or if the textarea will be printed into the document as is.

src/editor-sidebar/update-panel.js
```diff
@@ -32,10 +33,26 @@ export const UpdateThemePanel = () => {
...
+	async function getThemeReadmeData() {
+		return apiFetch( {
+			path: '/create-block-theme/v1/get-readme-data',
+			method: 'GET',
+		} ).then( ( response ) => {
+			if ( response.status === 'SUCCESS' ) {
+				return response.data;
+			}
+			return {
+				recommendedPlugins: '',
```
jffng: Why do we need to return an object with only this key as an empty string? Seems like a strange exception.

jffng (commented):
I tested and verified this is working as expected in the wp-admin and site editor flows.

While some of the changes are just formatting, I do think this is a lot of code to add for what is effectively just reading/writing some unstructured text to the readme, without enforcement or validation of those plugins coming from wp.org, for example. Unless I am missing something? It's a fair suggestion, but what's the real utility? For example, Gutenberg has a mechanism to prompt users to install respective plugins for blocks that are missing from templates / patterns a theme supplies.

I don't mean to block the PR, just adding my general thoughts.

jffng (approved):
On second thought, I can see how having this info is useful for theme authors to inform users what plugins to install, and a lot of the changes are just refactoring / formatting, so giving it the ✅ .

src/editor-sidebar/update-panel.js
mikachan: I've updated this so it's now returning an empty object with no specific keys.

mikachan:
Thanks, both, for testing this out!

> It's been hard to find an explanation of this section

I was also looking for an example, and the best reference I can find is here: https://make.wordpress.org/themes/handbook/review/required/#6-plugins. I think a placeholder is a good idea, I've added one here: https://github.com/WordPress/create-block-theme/pull/402/commits/1b293ab6db297e2e874eebe3f8abbab480bf6503, which looks like this:

```
Plugin Name
https://wordpress.org/plugins/plugin-name/
Plugin Description
```

> and a lot of the changes are just refactoring / formatting

Yes... sorry about that, this snowballed as I saw loads of improvements to the existing readme logic! Thank you for working through it.

@pbking and I paired on this earlier today, and with our latest changes, the 'Recommended Plugins' textarea in the editor will now pre-populate with data from the current readme file, if this section already exists. This functionality also makes it easier to expand in the future for other parts of the readme file (e.g. image credits). I think this enhances the UX, as the user sees some feedback from updating the theme data.

vcanales (approved):
Thanks for adding the placeholder! I gave it another go and it LGTM.

pbking:
I made a teeny tiny tweak to the regex processing to ensure that if another section is in readme.txt following the recommended plugins section that the title `== TITLE HERE ==` of the section remains unchanged (it was removing the first `==`'s)

Otherwise this cleaned up great!  Well done @mikachan !  This was a blast to pear on!

PR #406, opened 2023-06-26
Add default value for recommended plugins section

vcanales:
Add a default value to `$recommended_plugins_section` to avoid errors when providing no recommended plugins.

Error:
```
Warning: Undefined variable $recommended_plugins_section in wp-content/plugins/create-block-theme/admin/create-theme/theme-readme.php on line 47
```

### DIFF
--- a/admin/create-theme/theme-readme.php
+++ b/admin/create-theme/theme-readme.php
@@ -24,9 +24,7 @@ class Theme_Readme {
 		$copyright_section      = self::copyright_section( $new_copyright_section, $original_theme_credits, $name, $copy_year, $author, $image_credits );
 
 		// Handle recommended plugins section.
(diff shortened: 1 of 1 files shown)


pbking:
I can't say that I was able to reproduce the error you mentioned.  I tried all the ways I could think of both in the wp-admin interface as well as site editor.

I'm not sure what is different to have caused you to get the error, however I can see how the change you made would mitigate that, and running it in all the scenarios I could caused no harm so I'm fine bringing this change in.

pbking (approved):

vcanales:
@pbking Thanks for reviewing! Here's a screen capture of what I experience

https://github.com/WordPress/create-block-theme/assets/1157901/1708beff-31e4-4a19-9054-0ac9eb81a8a5

PR #411, opened 2023-07-07
Set the initial version to 1.0.0 insted of 0.0.1

matiasbenedetto:
## What ?
Set the initial version of the exported themes to 1.0.0 instead of 0.0.1

## Why?
Versions below 1.0.0 imply that the product is not production ready. Block themes are production ready.


mikachan (approved):
Makes sense 👍

madhusudhand (approved):

PR #413, opened 2023-07-10
Updated correct template part location when creating patterns on save

eirichmond:
When editing templates and including media when you "Save Changes" the block is converted to a pattern, the template part references are then included in the pattern in a directory /patterns/ this breaks the template part references.

This fixes the header and footer reference, moving it back to the template from the pattern.

### DIFF
--- a/admin/create-theme/theme-templates.php
+++ b/admin/create-theme/theme-templates.php
@@ -118,6 +118,14 @@ class Theme_Templates {
 
 			// If there are images in the template, add it as a pattern
 			if ( ! empty( $template_data->media ) ) {
+
+				// get all template parts to create template links
+				$template_parts = get_block_templates( array(), 'wp_template_part' );
+				$template_links = self::create_template_links( $template_parts );
+
+				// remove template strings from template data content
+				$template_data->content = str_replace( array_values( $template_links ), '', $template_data->content );
+
 				// If there is no templates folder, create it.
 				if ( ! is_dir( get_stylesheet_directory() . DIRECTORY_SEPARATOR . 'patterns' ) ) {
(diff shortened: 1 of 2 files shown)


pbking:
Thank you for this @eirichmond!  This is a topic I've been struggling with for a while.

This is definitely something that I've manually had to do for a few themes.  Template Parts referenced from Patterns (for various reasons) continues to be something we're struggling to work around.

Unfortunately, while this change you're introducing works for MOST themes I've recently come across some where this would not work.  The most obvious example I've seen is where the header template is not the first thing in the page template but is instead embedded in a column block or nested in some other way.  I don't believe this change would work for that scenario.  I believe we can't assume the structure of the template, even though 'header is first, footer is last' is definitely the most commonplace.

eirichmond:
@pbking I completely understand what you mean, and I actually had a similar thought after I created this pull request.

I was thinking that perhaps, before creating the patterns and templates, we could generate a temporary unique string reference in the pattern and store it as the index in an array. This way, we can keep track of where the template parts should logically be placed and reference them from that array to ensure that the template parts and patterns are outputted in the correct order within the templates themselves.

However, it's just an idea that came to me later, and I haven't figured out yet how it would work in the code. I can visualise it in my head, but I need to dive deeper into it. If I can figure out a solution, I'll definitely update this pull request.

pbking:
This was a good idea (I had a very similar one) but ultimately I think the correct fix here is to
1) ensure that patterns are created without references to template parts
2) change gutenberg to allow patterns to reference template parts without assigning a 'theme' attribute.  I opened a [Gutenberg issue](https://github.com/WordPress/gutenberg/issues/53194) to hopefully address that.

I'm going to close this item.

PR #417, opened 2023-07-14
Introduce basic `wp-env` environment

t-hamano:
Currently, I think either of the following methods is mainly used to develop this project locally.

- Set up your own local environment and clone this repository in the `plugins` directory
- Use the globally installed `env` package, and define a path relative to this project in the `plugins` property of `.wp-env.json`.

I don't think the introduction of wp-env has been discussed in this project yet, but it would be useful to be able to set up a local environment with only the npm packages and commands that this project depends on, without having to prepare a special environment.

This PR introduces a basic wp-env execution environment.

- run `npm ci`
- run `npm run wp-env start`
- Access `localhost:8888/wp-admin` and confirm that the Create Block Theme plugin works.

### DIFF
--- a/.wp-env.json
+++ b/.wp-env.json
@@ -1,0 +1,4 @@
+{
+	"core": "WordPress/WordPress",
+	"plugins": [ "." ]
+}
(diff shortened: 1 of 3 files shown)


vcanales (approved):
Thanks, I agree that this is missing! This is good to go as is.

PR #419, opened 2023-07-14
Try: Apply theme color schema to Manage Theme Fonts page

t-hamano:
This PR attempts to apply the user's theme color scheme to a Manage Theme Fonts page rendered as a React app.

| Before | After |
|--------|--------|
| before | after |

---

The WordPress admin page loads the CSS file according to the color schema selected by the user and applies the colors. 

admin-css

Therefore, the "Create Block Theme" page will correctly reflect that color.

create-block-theme

However, components of the Gutenberg package, such as `Button`, do not apply this CSS because their colors are controlled by their own CSS variables.

React-based canvases, such as the Post Editor and Site Editor, use [the `base-styles` package](https://github.com/WordPress/gutenberg/blob/trunk/packages/base-styles/README.md) and use CSS variables generated from its [mixins](https://github.com/WordPress/gutenberg/blob/9879bdb7abebf9f45299f38c4a05682b7dc5d730/packages/base-styles/_mixins.scss#L439).

Therefore, on a React page that is built independently, we need to explicitly call this mixin in order to apply the user's color schema to the components.

I have first added `@wordpress/base-styles` to the npm package to achieve this. Then I defined `index.scss` in the `src` directory to import the styles according to [the documentation](https://github.com/WordPress/gutenberg/blob/trunk/packages/base-styles/README.md#scss-utilities-and-variables).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,6 +15,7 @@
 			"devDependencies": {
 				"@actions/core": "^1.10.0",
 				"@emotion/babel-plugin": "^11.11.0",
+				"@wordpress/base-styles": "^4.28.0",
 				"@wordpress/browserslist-config": "^5.16.0",
 				"@wordpress/element": "^5.10.0",
 				"@wordpress/env": "^8.3.0",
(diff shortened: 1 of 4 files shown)


mikachan (approved):
Nice enhancement, thank you for working on this! Works well in my testing, happy to bring this in.

PR #427, opened 2023-07-27
Update .distignore

swissspidy:
Stop wastefully deploying unneeded files to WordPress.org. See https://plugins.svn.wordpress.org/create-block-theme/trunk/


vcanales (approved):
Tested by deleting everything included in `.distignore` after building; LGTM.

PR #428, opened 2023-07-27
Use wp_add_inline_script for createBlockTheme object

vcanales:
This change ensures we follow accepted good practices, as `wp_localize_script` should only be used when localizing strings.

It's more verbose, but it doesn't lend itself to confusion.

Relevant comment: https://developer.wordpress.org/reference/functions/wp_localize_script/#comment-3213

### DIFF
--- a/admin/class-react-app.php
+++ b/admin/class-react-app.php
@@ -38,7 +38,8 @@ class React_App {
 				esc_url( plugins_url( 'assets/google-fonts/fallback-fonts-list.json', __DIR__ ) ),
 				esc_url( admin_url() ),
 				esc_url( get_template_directory_uri() )
-			)
+			),
+			'before'
 		);
 	}
 }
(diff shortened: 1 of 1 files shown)


mikachan (approved):
This is working well for me. I can see the `createBlockTheme` global variable is created in the same format, and looks more readable in the page source:

image

The "Add Google fonts to your theme" page still works as expected ✅

I did notice that the fonts json file is now included after the build script (ID `create-block-theme-app-js`), whereas currently the file is included before:

image

I don't think it affects the way the app works, but it looks like we can switch the order by adding `'before'` to the `wp_add_inline_script` call:

```
wp_add_inline_script(
	'create-block-theme-app',
	sprintf(
		$script_content,
		esc_url( plugins_url( 'assets/google-fonts/fallback-fonts-list.json', __DIR__ ) ),
		esc_url( admin_url() ),
		esc_url( get_template_directory_uri() )
	),
	'before'
);
```

I don't think it matters either way, just wanted to point out the order change!

vcanales:
> I don't think it matters either way, just wanted to point out the order change!

Good catch; I agree that it doesn't seem to affect things, but in order to keep things as close as they were before, I've added your suggestion.

PR #430, opened 2023-07-28
[I18n] Fix plugin internationalization

pedro-mendonca:
This PR addresses many i18n issues:

- [x] Fix missing `wp_set_script_translations()` to translate JavaScript scripts
- [x] Fix missing gettext wrappers
- [x] Fix missing text domains
- [x] Fix extra leading slash on form assets path
- [x] Add i18n to Font Styles (Normal/Italic)
- [x] Use correct i18n for strings with plurals - include variable on string ( %d Variant / %d Variants )

Solves issue reported on https://wordpress.org/support/topic/the-screen-manage-theme-fonts-is-not-translated/

### DIFF
--- a/admin/create-theme/theme-form.php
+++ b/admin/create-theme/theme-form.php
@@ -223,7 +223,7 @@ Plugin Description',
 	}
 
 	public static function form_script() {
-		wp_enqueue_script( 'form-script', plugin_dir_url( dirname( __FILE__ ) ) . '/js/form-script.js' );
-		wp_enqueue_style( 'form-style', plugin_dir_url( dirname( __FILE__ ) ) . '/css/form.css' );
+		wp_enqueue_script( 'form-script', plugin_dir_url( dirname( __FILE__ ) ) . 'js/form-script.js' );
+		wp_enqueue_style( 'form-style', plugin_dir_url( dirname( __FILE__ ) ) . 'css/form.css' );
 	}
 }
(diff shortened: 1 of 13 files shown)


src/utils.js
```diff
@@ -21,13 +23,21 @@ export function getGoogleVariantFromStyleAndWeight( style, weight ) {
 	return weight + style;
 }
 
+export function getFontStyleLocalized( style ) {
```
vcanales: I think naming this function `localizeFontStyle` works best in this context.

```suggestion
export function localizeFontStyle( style ) {
```

admin/create-theme/theme-form.php
```diff
@@ -223,7 +223,10 @@ public static function create_admin_form_page() {
 	}
 
 	public static function form_script() {
```
vcanales, line 225: This is a great catch; it was frustrating me quite a bit, as it wasn't working with the extra `/` present.

It's working for me now.

vcanales (approved):
This is a great improvement; thanks for taking the time to implement it @pedro-mendoca!

We should address any further improvements and as a follow-up, in order to get translations out ASAP.
You [mentioned in slack](https://wordpress.slack.com/archives/C02RP50LK/p1690559505577999?thread_ts=1690429456.937199&cid=C02RP50LK) that `/assets/boilerplate/theme.json` isn't being parsed; we should add an issue to get to that later.

Salve for the small naming nitpick, this is good to go.

PR #438, opened 2023-08-09
Add wp-admin menu for git repository configuration

madhusudhand:
This adds Admin menu for git repository configuration.

It enables to add settings such as Repository URL, Access Token etc.. and Disconnect repository.
It saves the set configuration using 'update_option'.

Note: This just saves and deletes the settings. Does not clone the actual repository.

image

image

### DIFF
--- a/admin/class-git-integration.php
+++ b/admin/class-git-integration.php
@@ -1,5 +1,6 @@
 <?php
 
+require_once( __DIR__ . '/class-react-app.php' );
 require_once( dirname( __DIR__ ) . '/includes/class-create-block-theme-settings.php' );
 
 class Git_Integration_Admin {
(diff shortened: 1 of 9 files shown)


pbking:
Considering how in previous incarnations of this work, from the site editor there was the option to connect individual themes to a repository.  This (from a UX perspective at least) seems to connect the entire SITE to a single repository.

If we're going to keep the repository settings per-theme then I believe there should be two options for this:
* Connect THIS theme to a repository
* Install a theme from a repository

Does that line up with your thinking?

madhusudhand:
@pbking I created this as small chunk to speedup the review process and bring the changes in :)
But I agree, having both options (connecting with all themes, and active theme) should give the overall picture for review.

Pushed next set of changes in this PR and following are few screenshots.

image

image

image

pbking:
We no longer have plans for CBT to manage a theme via git so closing this as an "interesting experiment".

PR #445, opened 2023-08-15
Updated Requires PHP version to 7.4

### LINKED ISSUE #444: PHP Required >= 7.4
The plugins readme.txt has `Requires PHP: 7.0` but currently the plugin requires PHP 7.4 or higher.

On PHP 7.3 the following error is triggered: [(support thread)](https://wordpress.org/support/topic/fatal-error-when-plugin-activated-5/)
> Parse error: syntax error, unexpected 'string' (T_STRING), expecting function (T_FUNCTION) or const (T_CONST) in /wordpress/wp-content/plugins/create-block-theme/admin/create-theme/cbt-zip-archive.php on line 8

On PHP 5.6, 7.0, 7.1, and 7.2 the following error is triggered:
> Parse error: syntax error, unexpected ')' in /wordpress/wp-content/plugins/create-block-theme/admin/class-create-theme.php on line 52

Note: I've tested this using https://playground.wordpress.net/?plugin=create-block-theme&php=7.3&wp=6.2&storage=temporary

_I've temporarily overridden the `Requires PHP` version on wordpress.org/plugins for this plugin to 7.4, upon next update it'll revert to the value in readme.txt_

kafleg:
Fixes #444

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -4,7 +4,7 @@ Tags: themes, theme, block-theme
 Requires at least: 6.0
 Tested up to: 6.3
 Stable tag: 1.13.2
-Requires PHP: 7.0
+Requires PHP: 7.4
 License: GPLv2 or later
 License URI: https://www.gnu.org/licenses/gpl-2.0.html
 


mikachan (approved):
Thanks for addressing this, @kafleg!

PR #461, opened 2023-10-24
Update google fonts on release instead of on a schedule

vcanales:
Currently, the data from the Google Fonts API is being updated every Tuesday, but pulling the new data makes sense closer to a release since, before that, current plugin installations won't consume it.

Here, I propose moving the automated workflow of the update to the Release Workflow so that this may be done when needed.

### Testing:
- On a fork, checkout this branch as `trunk`
- If you want to test the actual file update, modify your copy of `assets/google-fonts/fallback-fonts-list.json`.
- Comment out [this line](https://github.com/WordPress/create-block-theme/pull/461/files#diff-85dcdbd0202633cb8228a007e37c479fe3568a8052aa67414669e1b7e453bd1cR18), otherwise, the step won't run.
- Run the workflow (Go to Actions > Deploy to WorkPress.org, choose any version increase)

The "Update Google Fonts JSON file" and "Commit Changes" steps should pass without errors.

Note that in your fork, the full workflow will report a failure, since you'll be missing the secrets necessary to actually deploy the plugin; this is expected.


.github/workflows/deploy-to-dotorg.yml
```diff
@@ -14,6 +14,44 @@ on:
...
+        # Steps represent a sequence of tasks that will be executed as part of the job
+        steps:
+            - uses: actions/checkout@v3
+            - uses: actions/setup-node@v3
+              with:
+                  node-version: 18
+
+            - name: Get current date
+              id: date
+              run: echo "{date}={$(date +'%Y-%m-%d')}" >> $GITHUB_OUTPUT
```
mikachan: Where is this date used? I couldn't see a reference to it so I was wondering what it was used for.

mikachan (approved):
Thanks for working on this! Tested in a fork and the "Update Google Fonts JSON file" and "Commit Changes" steps passed without errors. Code looks good to me!

I think this is ready to bring in - I have one question but it's not a blocker.

.github/workflows/deploy-to-dotorg.yml
vcanales: It was just printing the date on the debug log; removing!

creativecoder:
I reviewed this post-merge, mostly to learn more about how create-block-theme releases work.

Thanks for the specific testing instructions, everything worked as expected on my fork (I did need to add a GOOGLE_FONTS_API_KEY env secret first, since I didn't have that).

One risk this approach seems to create is that it removes a manual review before committing the Google fonts configuration, since that's now tied to an automated release process. For example, if the fonts API request returned successfully, but had erroneous data, like an empty array of fonts, I think that would be committed automatically to the plugin as is was released. Is that right?

It's an unlikely scenario, so not a huge risk. But a nice improvement might be some basic checking of the fonts API response in the update script to make sure it looks to as expected, and throwing an error if any problems are detected.

PR #469, opened 2023-11-06
Adds error handling for themes.svn.wordpress.org request

creativecoder:
Adds error handling so that if the request to themes.svn.wordpress.org to check theme names fails for some reason, the request fails gracefully and outputs the error to the JS console.

### Testing instructions

- Spoof an error when making request to themes.svn.wordpress.org, e.g. add domain to /etc/hosts file or manually return a WP_Error from `WP_Theme_Directory::get_theme_names`
- See that the with this change, the error is logged to the JS console, rather than being unhandled.


pbking (approved):
Works as advertised

creativecoder:
Thanks for the review @pbking !

PR #470, opened 2023-11-06
Update theme form markup and styles

creativecoder:
This change adds a handful of improvements to the markup and styles for the main plugin page, including

- Radio button labels and form field descriptions match other wp-admin page style and structure more closely
- All description text is tied to its input field with an `aria-describedby` attribute
- Theme tags have responsive column styles
- URLs are removed from translation strings

### Testing instructions

- The changes should be markup structure and styles only
- Test theme creation at `/wp-admin/themes.php?page=create-block-theme` and make sure the form submits successfully

| **Before** | **After** |
| - | - |
| image | image |
| image | image |

### DIFF
--- a/admin/create-theme/theme-form.php
+++ b/admin/create-theme/theme-form.php
@@ -160,7 +160,7 @@ class Theme_Form {
 							<label for="theme-name">
 								<?php _e( 'Theme Name (*)', 'create-block-theme' ); ?>
 							</label>
-							<input id="theme-name" placeholder="<?php _e( 'Theme Name', 'create-block-theme' ); ?>" type="text" name="theme[name]" class="large-text" id="theme-name" autocomplete="off">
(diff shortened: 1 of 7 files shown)


pbking:
They look like good changes, though I only did a cursory glance.

To be honest I would rather just completely remove the `wp-admin` interface rather than spend efforts improving it.

The only parts of the workflow that I know aren't available in the site editor are:
* export as a child theme/sibling theme
* include/manage screenshot

vcanales (approved):
I'm keen on getting this in, regardless of what we decide on for the future of the admin page; these changes are a valuable improvement for it in the meantime.

creativecoder:
Thanks for the reviews and feedback! To me it seems worth merging this, even if we end up throwing it out soon in favor of other changes to the wp-admin section of the plugin.

Note that I added a couple more small tweaks based on feedback from running the form markup through validator.w3.org.

This seems ready to go, so I'll go ahead and merge it!

creativecoder:
For anyone following here, I made a mistake in my final markup changes that removed a closing `>`. That's been corrected in a follow-up PR: https://github.com/WordPress/create-block-theme/pull/475

PR #467, opened 2023-11-03
Use npx for commit linting scripts

creativecoder:
This update the package.json config for pre-commit hooks to use `npx`.

This prevents errors when committing if you don't have `wp-scripts` installed globally, and makes sure to use the version specified by package.json.

### Testing instructions

Make a commit locally on this branch and make sure pre-commit hooks run as expected.

Since this PR is from a fork, you may want to use `github-cli` to test, e.g. `gh pr checkout 467` (https://cli.github.com/manual/gh_pr_checkout)

PR #476, opened 2023-11-15
Add quotes to font family names

### LINKED ISSUE #468: Fonts with invalid names are not displayed or installed properly
Font Manager fails to display & properly install some fonts due to the _Font name_ input (in the Add Local Font page) or the _Select font_ dropdown option value (in the Add Google Font page) having non-standard characters or character combinations, which are not allowed in a CSS `font-family` value. The internal name of the font is used as a initial value for the input and dropdown option values.

Problem occurs when the internal name of the font is invalid for the purpose it is used, so loading the font is broken from the start. Failure to display the font causes confusion for the user, who may think that there is something wrong with the font itself, since the UI shows no errors or indications of a problem.

Problematic values in the font name are for example:
- A number preceded by a space, for example `Exo 2`. For more information, see discussion in this [Stack Overflow topic](https://stackoverflow.com/a/8951680)
- Any special characters

#### Add Google Font:

1. Add `Exo 2` font to the project via the `Manage Theme Fonts > Add Google Font` feature.
2. Bug: Some fallback font is displayed, but there is no indication of a problem in the UI. An error is shown in the console: `DOMException: An invalid or illegal string was specified`.
3. Bug: if you select some font variants and continue installing, the font is added but is not showing correctly.

#### Add Local Font:

1. Download [Exo 2](https://fonts.google.com/specimen/Exo+2) font from Google Fonts.
2. Add one of the font files to WordPress via the `Manage Theme Fonts > Add Local Fonts` feature.
4. Bug: Some fallback font is displayed, but there is no indication of a problem in the UI. An error is shown in the console: `DOMException: An invalid or illegal string was specified`.
5. Workaround: remove the space in the prefilled _Font name_ input value, so the value is `Exo2` and then the correct font is displayed.

### Suggested fix:

1. Font names which have a number preceded by a space: the generated `font-family` value should be wrapped to quotes, for example `'Exo 2'`
2. Other non-standard characters in the name, or any other failure to display the font: _Font name_ field should have validation and display an error message.

vcanales:
~Checks if the font family name of a given font, either added through Google Fonts or Locally, ends with a number and if so, it adds single quotes to its name.~ Add single quotes to font family names to avoid errors related to trying to parse font names with special characters, white spaces, or numbers.

Also, check if the font family name of a font being uploaded includes non-alphanumeric characters and invalidates the form if so.

Currently, no validation errors are shown in the UI; I'll be making a note of this to add it in as a follow-up change.

Fixes: #468

### DIFF
--- a/src/local-fonts/index.js
+++ b/src/local-fonts/index.js
@@ -6,6 +6,7 @@ import DemoTextInput from '../demo-text-input';
 import Demo from '../demo-text-input/demo';
 import { variableAxesToCss } from '../demo-text-input/utils';
 import BackButton from '../manage-fonts/back-button';
+import { addQuotesToName } from '../utils';
 
 const INITIAL_FORM_DATA = {
 	file: null,
@@ -61,7 +62,8 @@ function LocalFonts() {
 		}
 
 		const data = await formData.file.arrayBuffer();
-		const newFont = new FontFace( formData.name, data, {
+		const sanitizedFontFamily = addQuotesToName( formData.name );
+		const newFont = new FontFace( sanitizedFontFamily, data, {
 			style: formData.style,
 			weight: formData.weight,
 		} );
--- a/src/local-fonts/index.js
+++ b/src/local-fonts/index.js
@@ -53,7 +53,7 @@ function LocalFonts() {
 			return {};
 		}
 		const style = {
-			fontFamily: formData.name,
+			fontFamily: addQuotesToName( formData.name ),
 			fontWeight: formData.weight,
 			fontStyle: formData.style,
 		};
--- a/src/local-fonts/index.js
+++ b/src/local-fonts/index.js
@@ -1,5 +1,5 @@
 import { __ } from '@wordpress/i18n';
-import { useEffect, useState } from '@wordpress/element';
+import { useCallback, useEffect, useState } from '@wordpress/element';
 import UploadFontForm from './upload-font-form';
 import './local-fonts.css';
 import DemoTextInput from '../demo-text-input';
@@ -34,7 +34,7 @@ function LocalFonts() {
 		setAxes( newAxes );
 	};
 
-	const isFormValid = () => {
+	const isFormValid = useCallback( () => {
 		// Check if font name is present and is alphanumeric.
 		const alphanumericRegex = /^[a-z0-9 ]+$/i;
 
@@ -46,7 +46,7 @@ function LocalFonts() {
 		}
 
 		return formData.file && formData.weight && formData.style;
-	};
+	}, [ formData ] );
 
 	const demoStyle = () => {
 		if ( ! isFormValid() ) {
@@ -63,33 +63,32 @@ function LocalFonts() {
 		return style;
 	};
 
-	// load the local font in the browser to make the preview work
-	const onFormDataChange = async () => {
-		if ( ! isFormValid() ) {
-			return;
-		}
+	useEffect( () => {
+		// load the local font in the browser to make the preview work
+		const onFormDataChange = async () => {
+			if ( ! isFormValid() ) {
+				return;
+			}
 
-		const data = await formData.file.arrayBuffer();
-		const sanitizedFontFamily = addQuotesToName( formData.name );
-		const newFont = new FontFace( sanitizedFontFamily, data, {
-			style: formData.style,
-			weight: formData.weight,
-		} );
-		newFont
-			.load()
-			.then( function ( loadedFace ) {
+			const data = await formData.file.arrayBuffer();
+			const sanitizedFontFamily = addQuotesToName( formData.name );
+			const newFont = new FontFace( sanitizedFontFamily, data, {
+				style: formData.style,
+				weight: formData.weight,
+			} );
+
+			try {
+				const loadedFace = await newFont.load();
 				document.fonts.add( loadedFace );
-			} )
-			.catch( function ( error ) {
+			} catch ( error ) {
 				// TODO: show error in the UI
 				// eslint-disable-next-line
 				console.error( error );
-			} );
-	};
+			}
+		};
 
-	useEffect( () => {
 		onFormDataChange();
-	}, [ formData ] );
+	}, [ formData, isFormValid ] );
 
 	return (
 		<div className="layout">
--- a/src/local-fonts/index.js
+++ b/src/local-fonts/index.js
@@ -35,9 +35,17 @@ function LocalFonts() {
 	};
 
 	const isFormValid = () => {
-		return (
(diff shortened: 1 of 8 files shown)


vcanales:
I've also added a couple of optimizations that I noticed while fixing this:

1. Refactor promise calls to use async/await syntax, for consistency's sake.
2. Avoid re renders by wrapping `isFormValid` with `useCallback`, and also move the definition of `onFormDataChange` inside the effect where it's used to avoid having to pass it as a dependency.

src/utils.js
```diff
@@ -76,3 +76,22 @@ export async function downloadFile( response ) {
...
+ * Trim a string.
+ * @param {string} str The string to trim.
+ * @return {string} The trimmed string.
+ */
+export function trim( str ) {
+	str.trim();
+}
+
+/*
+ * Add quotes to fon name if it ends with a number.
```
mikachan:
```suggestion
 * Add quotes to font name if it ends with a number.
```

src/utils.js
```diff
@@ -76,3 +76,22 @@ export async function downloadFile( response ) {
...
+ * @param {string} familyName The font family name.
+ * @return {string} The font family name with quotes if it ends with a number.
+ */
+
+export function addQuotesToName( familyName ) {
+	return trim( familyName.match( /\d$/ ) ? `'${ familyName }'` : familyName );
```
mikachan: This is only handling font family names that end with a number, so it won't cater to fonts such as [Libre Barcode 128 Text](https://fonts.google.com/specimen/Libre+Barcode+128+Text).

I also noticed that the CSS `font-family` [docs](https://developer.mozilla.org/en-US/docs/Web/CSS/font-family#valid_family_names) mention more than just numbers as a requirement for quotes:

> It is a good practice to quote font family names that contain white space, digits, or punctuation characters other than hyphens.

I'm wondering if it would be better to wrap all font names in quotes regardless of whether the name includes numbers or special characters. Are there any downsides to this? 

I had a look through the [LibFont](https://github.com/Pomax/lib-font/) library to see if there were any helpful functions, I didn't find any but I noticed that they're wrapping font names in quotes on output [here](https://github.com/Pomax/lib-font/blob/master/src/utils/fontface.js#L56).

mikachan (commented):
This works in my testing for the font Exo 2, but I don't think it'll catch the names of the following fonts:

- Libre Barcode 128 Text
- Press Start 2P
- Rock 3D
- Rubik 80s Fade

I've left a longer comment about this below, but I'm wondering if it may be best practice to always wrap font names in quotes, as long as the name isn't a [generic fallback name](https://developer.mozilla.org/en-US/docs/Web/CSS/font-family#generic-name) (which they're unlikely to be, as they're custom fonts).

> Refactor promise calls to use async/await syntax
> Avoid re renders by wrapping isFormValid with useCallback,

Nice!

Definitely not a blocker for this PR but it would be great if we could write some tests, especially for things like the utils functions which are hopefully easier to cover in isolation. I know we've not historically done this in this repo - the list of example fonts above made me think that they'd make great test cases.

vcanales:
@mikachan You're correct in that those cases should also be considered. Here's a few examples:

```console
// Normal
> new FontFace('Example1', {});
< FontFace { family: "Example1", style: "normal", weight: "normal", stretch: "normal", unicodeRange: "U+0-10FFFF", variant: "normal", featureSettings: "normal", variationSettings: "normal", display: "auto", ascentOverride: "normal" }

// Notice the empty family name:
> new FontFace('Example 2', {});
< FontFace { family: "", style: "normal", weight: "normal", stretch: "normal", unicodeRange: "U+0-10FFFF", variant: "normal", featureSettings: "normal", variationSettings: "normal", display: "auto", ascentOverride: "normal" }

// Same, empty family name:
new FontFace('Rubik 80s Fade', {});
FontFace { family: "", style: "normal", weight: "normal", stretch: "normal", unicodeRange: "U+0-10FFFF", variant: "normal", featureSettings: "normal", variationSettings: "normal", display: "auto", ascentOverride: "normal" }
```

For some reason, fonts that have white spaces and numbers in their name didn't output errors, but they're just as wrong. I'll have `addQuotesToName` just add quotes and trim names without conditions. I can't seem to find any downsides to doing this.

vcanales:
> Definitely not a blocker for this PR but it would be great if we could write some tests, especially for things like the utils functions which are hopefully easier to cover in isolation. I know we've not historically done this in this repo - the list of example fonts above made me think that they'd make great test cases.

I don't think we even have the tools for this right now, so I'll add it to my to-do for this week. I agree that this change in particular is very well fit for unit tests.

src/local-fonts/index.js
```diff
@@ -33,11 +34,19 @@ function LocalFonts() {
...
 	};
 
-	const isFormValid = () => {
-		return (
-			formData.file && formData.name && formData.weight && formData.style
-		);
-	};
+	const isFormValid = useCallback( () => {
+		// Check if font name is present and is alphanumeric.
+		const alphanumericRegex = /^[a-z0-9 ]+$/i;
```
mikachan, line 39: Do you think this regex should also include punctuation characters, based on the `font-family` spec? As this would still be a valid font name:

> It is a good practice to quote font family names that contain white space, digits, or punctuation characters other than hyphens.

I'm not sure which characters they specifically are! And I'm also struggling to immediately find many font families that include special characters.

mikachan (commented):
We should add the new `addQuotesToName` function to `demoStyle` too (around line 56 of local-fonts/index.js), something like this:

```
		const style = {
			fontFamily: addQuotesToName( formData.name ),
			fontWeight: formData.weight,
			fontStyle: formData.style,
		};
```

This allows the inline font demo on the local fonts page to apply the font family correctly.

src/local-fonts/index.js
vcanales: I think we just have to make a call, since Font Families can be special characters, unicode, numbers, etc. But for the purpose of generating class names from these font family names, it's practical to request names that are compatible with this requirement. Therefore, restricting these names to alphanumeric characters seems like a sensible approach in this context.

vcanales:
@mikachan I ended up adding tests here; it was simpler than expected. I added @wordpress/wp-scripts as a dependency now, which undoes #467.

src/local-fonts/index.js
mikachan: > But for the purpose of generating class names from these font family names, it's practical to request names that are compatible with this requirement.

Yup ok, that makes sense 👍

mikachan:
> I ended up adding tests here; it was simpler than expected.

Awesome! This is a great start. With this we can start adding tests in other areas too!

mikachan (approved):
Thanks for working on all of this! I think we should bring this in.

PR #477, opened 2023-11-16
Add @vcanales to contributors list

vcanales:
It's a silly change, but I just realized I'm not listed on the site readme or on the site :) 
https://wordpress.org/plugins/create-block-theme/#developers


mikachan (approved):
It's not silly at all!

PR #479, opened 2023-11-16
Set page titles that set only within CBT

### LINKED ISSUE #478: Admin Titles being overwritten
Most admin titles are being set to `Manage Theme Fonts` when this plugin is enabled. It may be that other plugins are missing titles, but this should be restricted to the page. 

```
$manage_fonts_page_title = _x( 'Manage Theme Fonts', 'UI String', 'create-block-theme' );
		$manage_fonts_menu_title = $manage_fonts_page_title;
		add_theme_page( $manage_fonts_page_title, $manage_fonts_menu_title, 'edit_theme_options', 'manage-fonts', array( 'Fonts_Page', 'manage_fonts_admin_page' ) );

		// Check if the admin page title is set, and if not, set one.
		// This is needed to avoid a warning in the admin menu, due to the admin page title not being set in
		// the add_submenu_page() function for the Google Fonts and Local Fonts pages.
		global $title;
		if ( ! isset( $title ) ) {
			$title = $manage_fonts_page_title;
		}

```

vcanales:
When setting a page title within the Manage Fonts page, we were leaking out of the plugin and setting the page title for any plugin admin pages with no title; this ensures that we only do it within the pages belonging to the Manage Fonts Admin class.

Fixes: https://github.com/WordPress/create-block-theme/issues/478

### DIFF
--- a/admin/class-manage-fonts.php
+++ b/admin/class-manage-fonts.php
@@ -31,25 +31,41 @@ class Manage_Fonts_Admin {
 			return;
 		}
 
+		$manage_font_page_slug     = 'manage-fonts';
+		$add_google_font_page_slug = 'add-google-font-to-theme-json';
+		$add_local_font_page_slug  = 'add-local-font-to-theme-json';
+
 		$manage_fonts_page_title = _x( 'Manage Theme Fonts', 'UI String', 'create-block-theme' );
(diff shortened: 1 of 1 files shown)


mikachan (approved):
This looks good to me and tests well 👍

PR #480, opened 2023-11-23
Make form files more specific to form page

mikachan:
This is a quick fix to make the plugin form styles more specific, and only allow the form.css and form-script.js files to load on the CBT plugin page.


vcanales (approved):
LGTM

PR #482, opened 2023-11-28
Proposal: remove the development-only warning

vcanales:
It feels like the warning might be a deterrent for some folks. While it's true that the plugin is intended for the development of themes, it doesn't seem like there's a risk behind installing the plugin in a production site, or, if there are any — eg. if the plugin would replace the active theme — we should warn the user before the action is taken.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@ Welcome to Create Block Theme - a WordPress plugin to create block themes from w
 
 It works alongside features that are already available in the Site Editor to enhance the workflow for creating block themes. After being tested in this plugin, some of the features included here may be moved into the Site Editor itself.
 
+*Disclaimer:* Create Block Theme enables development features and thus is a tool that should be treated like as such; you can think of it as a Development Mode for WordPress, and you should keep in mind that changes made through this plugin could change your site and/or theme permanently.
+
 This plugin allows you to:
 
 - Export your existing theme with all customizations made in the Site Editor
--- a/README.md
+++ b/README.md
@@ -4,8 +4,6 @@ Welcome to Create Block Theme - a WordPress plugin to create block themes from w
 
 It works alongside features that are already available in the Site Editor to enhance the workflow for creating block themes. After being tested in this plugin, some of the features included here may be moved into the Site Editor itself.
 
-_The plugin is for development only — it is not intended for use on production websites, but to be used as a tool to create new themes._
-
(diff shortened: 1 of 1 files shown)


pbking:
Without CBT changes to themes can be reversed, all the way back to the theme's design when it was installed.

With CBT installed it's possible for a user to "save" the Theme at which point the only way to "undo" those changes is to re-install the theme.  It feels like an important step and, even with a "warning" on the "save" step still doesn't seem like enough of a potential deterrent. 

I do think that, generally speaking, CBT doesn't belong on a production website, especially with people who might have permissions to use the Site Editor but not the understanding not to use CBT when they do.

vcanales:
> With CBT installed it's possible for a user to "save" the Theme at which point the only way to "undo" those changes is to re-install the theme. It feels like an important step and, even with a "warning" on the "save" step still doesn't seem like enough of a potential deterrent.

We might want to make it so that this is not necessarily the case?

> I do think that generally speaking, CBT doesn't belong on a production website.

I agree with this, and at the same time feel like we could write a better warning that doesn't make it feel like the plugin itself is in development mode. Maybe turn the message from a warning into something more encouraging, such as "This plugin is intended to be "Developer Mode" for WordPress/the Site Editor" (based on an IRL conversation I had yesterday).

pbking:
> I agree with this, and at the same time feel like we could write a better warning that doesn't make it feel like the plugin itself is in development mode. Maybe turn the message from a warning into something more encouraging, such as "This plugin is intended to be "Developer Mode" for WordPress/the Site Editor" (based on an IRL conversation I had yesterday).

Good point, I see what you're saying.  Less "this is unstable and thus shouldn't be in production" and more "this enables development features and thus is a tool that should be treated like that".

vcanales:
I've adjusted the warning to:

> *Disclaimer:* Create Block Theme enables development features and thus is a tool that should be treated like as such; you can think of it as a Development Mode for WordPress, and you should keep in mind that changes made through this plugin could change your site and/or theme permanently.

Feel free to suggest or commit changes.

pbking (approved):
I like that a lot better than warning not to use it in production (which I agree makes it sound unstable, which isn't the point) or removing the statement completely (which doesn't warn of anything).

I think this seems like good wording.

🚢

mikachan (approved):
I agree, I like this wording too 👍

PR #499, opened 2023-12-25
Add text domain to translation target

Olein-jp:
When I was using the plugin in a Japanese environment, I found a few points that were translated but not adapted.

After checking the source code, I found that the text domain was not described in the corresponding sections, so I added them.

I also found a few other cases, so we took the same measures.

### DIFF
--- a/src/editor-sidebar/update-panel.js
+++ b/src/editor-sidebar/update-panel.js
@@ -200,7 +200,8 @@ export const UpdateThemePanel = () => {
 					help={
 						<>
 							{ __(
-								'List the recommended plugins for this theme. e.g. contact forms, social media. Plugins must be from the WordPress.org plugin repository.'
+								'List the recommended plugins for this theme. e.g. contact forms, social media. Plugins must be from the WordPress.org plugin repository.',
+								'create-block-theme'
 							) }
 							<br />
 							<ExternalLink href="https://make.wordpress.org/themes/handbook/review/required/#6-plugins">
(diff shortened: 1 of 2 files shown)


src/editor-sidebar/update-panel.js
```diff
@@ -210,9 +210,12 @@ export const UpdateThemePanel = () => {
...
 					// eslint-disable-next-line @wordpress/i18n-no-collapsible-whitespace
-					placeholder={ __( `Plugin Name
+					placeholder={ __(
+						`Plugin Name
 https://wordpress.org/plugins/plugin-name/
-Plugin Description`, 'create-block-theme' ) }
+Plugin Description`,
+						'create-block-theme'
+					) }
 					value={ theme.recommended_plugins }
```
matiasbenedetto, line 219: It seems like the translation domain (`create-block-theme`) is already there.

matiasbenedetto (commented):
Hi! @Olein-jp, thanks for submitting this PR.
It seems like the translation domain for the string you updated is already there, and the PR is only changing whitespaces, or I'm missing something else?

src/editor-sidebar/update-panel.js
Olein-jp: https://github.com/WordPress/create-block-theme/pull/499/commits/bdd7389efcc9b631c9dc32f3cf849907374ac1d4

I added text domain here, but it gave me a JS syntax error, so I adjusted the whitespace separately.

mikachan (approved):
Thanks for this, @Olein-jp! Looks good to me, and I can't see any other missing text domains in any other `__()` calls.

src/editor-sidebar/update-panel.js
mikachan: Yeah, looks like the text domain was missing from here.

PR #505, opened 2024-01-14
Fix browser warning error when clicking the reset button

t-hamano:
This PR fixes warning error when clicking the reset button.

image

`resetAxes()` is called to reset these two controls for variable fonts in the font file preview settings.

resetAxes

However, on the "Manage Theme Fonts" and "Add Google Fonts" pages, these two components related to variable fonts are not rendered. In other words, the `<DemoTextInput>` component is not passed props(`axes`, `setAxes`, `resetAxes` ) regarding variable fonts.

https://github.com/WordPress/create-block-theme/blob/cef22d8a4f622589df4f486a2315141650f80cc9/src/google-fonts/index.js#L266

https://github.com/WordPress/create-block-theme/blob/cef22d8a4f622589df4f486a2315141650f80cc9/src/manage-fonts/index.js#L161

As a result, `resetAxes()` is undefined and causes an error.

To avoid this error I used optional chaining (`?.`).

### DIFF
--- a/src/demo-text-input/index.js
+++ b/src/demo-text-input/index.js
@@ -82,7 +82,7 @@ function DemoTextInput( { axes, setAxes, resetAxes } ) {
 						onClick={ () => {
 							resetDefaults( 'sentence' );
 							handleDemoTypeChange( 'sentence' );
-							resetAxes();
+							resetAxes?.();
 						} }
 					>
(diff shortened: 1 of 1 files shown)


mikachan (approved):
Very nice, thanks @t-hamano!

PR #509, opened 2024-01-24
Use WOFF2 from Google Fonts

matiasbenedetto:
## What?
- Use WOFF2 format to add font faces from Google Fonts  

## Why?
WOFF2 is a lighter format than TTF and is compatible with most browsers.

## How?
- Update the Google Fonts API URL to get the WOFF version of the fonts.
- Update the Google Fonts API data in the repo.

## Testing instructions:
- Go to "Add Google Fonts" section.
- Add a few fonts to test that the woff2 fonts are added to your theme.


mikachan (commented):
I can successfully add Google fonts and I can see they're WOFF files, however I'm seeing the following error when I select a Google font from the dropdown: 

`TypeError: e.brotliDecompressSync is not a function at K`

image

vcanales:
Font management is no longer a part of this plugin.

PR #512, opened 2024-01-30
Add files and directories not needed for release to `.distignore`

t-hamano:
I noticed that this plugin published in the plugin directory has `.wp-env.json` file.

https://plugins.trac.wordpress.org/browser/create-block-theme/trunk

image

I believe this is unnecessary for published plugins.

### DIFF
--- a/.distignore
+++ b/.distignore
@@ -1,11 +1,13 @@
 # Files to be ignored in the plugin release
 
 node_modules/
+vendor/
 .gitignore
 update-google-fonts-json-file.js
 update-version-and-changelog.js
 package.json
 src/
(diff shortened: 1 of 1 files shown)


vcanales:
Agreed. Should we also add `test/` for good measure?

t-hamano:
> Should we also add test/ for good measure?

Makes sense. I also added `vendor/` directory. Composer is not used in the current deploy action, but this is to prevent `vendor` directory from being unintentionally released if it is introduced in the future.

vcanales (approved):
LGTM! Thanks!

PR #518, opened 2024-03-13
Use CORE for Font Management

### LINKED ISSUE #451: Saving Font Assets from native WP Font Management tooling
With the introduction of the [Font Manager](https://github.com/WordPress/gutenberg/issues/54169) in WP 6.4 CBT should be able to work with that workflow to manage fonts.

Additionally the Font Management portion of the application can be removed.

pbking:
Fixes: #451

Copy ACTIVATED USER FONTS to the THEME when SAVED.

REMOVE UNACTIVATED FONTS from the THEME when SAVED.

To test:
## Install Fonts to Theme
* Install and Activate fonts from Google Fonts
* 'Save' the theme with Create Block Theme
* Note that the font assets are copied to the theme (`/assets/fonts`) and added to the theme.json
* Note that the INSTALLED fonts have been DEACTIVATED (and can be safely deleted if desired)

## Uninstall Fonts from Theme
* Deactivate fonts using the Font Library UI
* 'Save' the theme with Create Block Theme
* Note that the deactivated fonts are removed from theme.json and font assets are removed from the asset folder


.github/workflows/pr-checks.yml
```diff
@@ -81,5 +81,5 @@ jobs:
             - name: Run PHP tests
               run: |
                   mysql -uroot -h127.0.0.1 -e 'SELECT version()' \
-                  && ./bin/install-wp-tests.sh --recreate-db wordpress_test root '' > /dev/null \
+                  && ./bin/install-wp-tests.sh --wp-version=trunk --recreate-db wordpress_test root ''  > /dev/null \
```
vcanales, line 84: Interesting that it didn't work on `nightly`; it worked here https://github.com/WordPress/create-block-theme/pull/526/commits/420ad20ae94d58257255ce2d9017abea095e68a1

vcanales (approved):
This is working well for me. LGTM.

.github/workflows/pr-checks.yml
pbking: Yes!  Very strange!  I didn't expect it to, I just noticed the || in the code and gave it a whirl.

It made zero sense but I ran with it!

PR #519, opened 2024-03-15
Remove/font management

pbking:
This removes the assets that manage Font Management in the Plugin.  This is a follow-up to being able to use the Native Font Mangement tooling with the plugin and can be merged once that issue has been resolved.

https://github.com/WordPress/create-block-theme/issues/451


pbking:
Closing this, will do it in another branch

PR #520, opened 2024-03-15
Add theme json inspector

### LINKED ISSUE #491: Add a theme.json editor
I'm seeing that even theme builders who work mainly from the Global styles interface still need to directly edit theme.json files for some settings.

Adding a theme.json editor to Create Block Theme could have multiple benefits:

- Help avoid needing to switch to a text editor to make these changes
- Show theme.json structure and valid settings to speed up workflows and provide built in documentation

A super basic version of this would be enabling the existing theme file editor to access theme.json.

A more advanced version could use the schema of theme.json to show what options are available and valid for the area of the file you are working in, perhaps using autocomplete, or having a more structured interface, like checkboxes for booleans and drop-downs when there is a limited set of valid choices.

vcanales:
Add a code editor to enable viewing theme json data for the current theme.

Tasks:

- [x] Add a code editor.
~- [ ] Add instructions, including information on how to use the editor.~
~- [ ] Enable writing JSON file to theme.~

### DIFF
--- a/src/editor-sidebar/json-editor-modal.js
+++ b/src/editor-sidebar/json-editor-modal.js
@@ -29,6 +29,7 @@ const ThemeJsonEditorModal = ( { onRequestClose } ) => {
 				extensions={ [ json() ] }
 				value={ themeData }
 				onChange={ handleSave }
+				readOnly={ true }
 			/>
 		</Modal>
 	);
--- a/src/editor-sidebar/json-editor-modal.js
+++ b/src/editor-sidebar/json-editor-modal.js
@@ -1,15 +1,19 @@
 import { useState, useEffect } from '@wordpress/element';
 import { Modal } from '@wordpress/components';
+import { useSelect } from '@wordpress/data';
 import CodeMirror from '@uiw/react-codemirror';
 import { json } from '@codemirror/lang-json';
 import { fetchThemeJson } from '../resolvers';
 
 const ThemeJsonEditorModal = ( { onRequestClose } ) => {
 	const [ themeData, setThemeData ] = useState( '' );
-
+	const themeName = useSelect( ( select ) =>
+		select( 'core' ).getCurrentTheme()
+	)?.name?.raw;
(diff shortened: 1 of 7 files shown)


src/editor-sidebar/json-editor-modal.js
```diff
@@ -0,0 +1,37 @@
...
+	return (
+		<Modal
+			isFullScreen
+			title={ `theme.json for ${ themeName }` }
+			onRequestClose={ onRequestClose }
+		>
+			<CodeMirror
+				extensions={ [ json() ] }
+				value={ themeData }
+				onChange={ handleSave }
```
mikachan, line 31:
```suggestion
				onChange={ handleSave }
				readOnly={ true }
```

I think we should add this `readOnly` option until we've implemented the save functionality. It's a little confusing that you're able to edit the JSON content - users might assume it's being saved automagically. The `readOnly` option disables any editing within the CodeMirror.

src/plugin-sidebar.js
```diff
@@ -188,6 +193,22 @@ const CreateBlockThemePlugin = () => {
...
+									onClick={ () => setIsEditorOpen( true ) }
+								>
+									{ __(
+										'Inspect Theme JSON',
+										'create-block-theme'
+									) }
+								</Button>
+								<Text variant="muted">
+									{ __(
+										'Open the theme JSON file to inspect theme data.',
```
mikachan:
```suggestion
										'Open the theme.json file to inspect theme data.',
```

Just a suggestion to make this wording more specific to the file name itself, in case we add other options to open style variation JSON files or colors/typography JSON, etc.

mikachan (commented):
This looks great, thanks so much for working on it! 🎉 It provides a good first step for revealing the whole theme.json contents without needing to open a separate code editor, and I'm looking forward to hearing feedback from users.

mikachan (approved):

jeffikus:
Oh this is awesome! Will give this a test!

PR #521, opened 2024-03-18
Removed backup code when WP_HTM_Tag_Processor is missing

pbking:
Removed backup code when WP_HTM_Tag_Processor is missing and raised minimum version to 6.5

Since 6.5 will be the minimum required version for this plugin (for font support) the fallback code for 6.0 can be removed.


pbking:
Closing this.  The code can be removed as those bits change anyway.  Most of this is taken care of in other branches.

PR #522, opened 2024-03-19
Extracted any logic that may need to be tested from the api class

### LINKED ISSUE #162: Cleanup Theme Export
Currently, an exported theme includes attributes such as `ref` and `id` for certain blocks, but these have no context outside of the environment where the theme was built. They can also break some blocks (a `ref` in a nav block that doesn't exist on a different WordPress instance will currently cause the nav block to break).

Here are the cases I can think of, but there may be more:

- [x] Remove `ref` attributes from the Navigation block
- [x] Remove `id` attributes from Image and Cover blocks
- [x] Remove `wp-image-[id]` classes from Image and Cover blocks
- [x] Remove `category` IDs from query blocks
- [x] Remove `taxQuery` attribute from query blocks
- [x] ~Remove `queryId` attributes from Query block~ I don't think this needs removing, see https://github.com/WordPress/gutenberg/issues/42730#issuecomment-1428524380 for more details.

### LINKED ISSUE #336: Images are not placed in the assets folder on export
Images from the wordpress.org pattern directory are not placed in the assets folder on export.

To reproduce:

Activate Twenty Twenty-Three
In any template in the site editor, add a pattern from the pattern directory that has an image,
For example Banners > Opening hours or Forest.
Save the template.

Try to export the theme using the plugin icon in the Site Editor.
I  got an error message saying "An error occurred while attempting to export the theme" and the browser console shows
Internal Server Error 500.

Go to the Admin > Appearance > Create block theme, and select export.

Check the content of the zip file.
The pattern was created with a local link, but the image is not in the folder.
`<!-- wp:cover {"url":"<?php echo esc_url( get_stylesheet_directory_uri() ); ?>/assets/images/forest.jpg","dimRatio":60,"minHeight":800,"align":"full"} -->`

### LINKED ISSUE #489: Patternize + Escape templates with strings
**Background**
Currently, only templates with images are transformed to patterns. In this process, the image is made "local" to the theme, as well as strings get escaped and marked for translation. 

https://github.com/WordPress/create-block-theme/blob/017d52004e46d6c2ab38e68ef1be2bc08922fb2f/admin/create-theme/theme-zip.php#L114C44-L130

**Proposal**
It could be useful to allow all templates that contain strings, to follow the same behavior — transformed to a pattern and strings escaped + marked for translation.

pbking:
Fixes: #162

This change moves any logic that may need to be tested outside of the api implementation.  This logic is now in a 'Theme_Create` class.  Logic is re-used in the other `Theme-*` classes whenever possible.

Additionally this change incorporates the removal of site-specific attributes from template markup fixing #162.  That work began in #524 and was copied to this branch.

This also fixes #336 since the asset processing from the API was cleaned up and now the assets are being copied as expected.  The logic to process this is now quite a bit simpler: the block parsing is mostly gone, with only the gathering of URLs happening as it used to.  Once that happens the URLs that were gathered are transformed to a local path, wrapped in a PHP tag and replaced in the template.

This fixes #489 as it escapes (most) strings during template parsing.

Many comments were added to clarify the code and functions renamed where appropriate.

Lastly templates that have strings (currently only in paragraphs, additional blocks are needed) have those strings wrapped in PHP tags and localized.  (Any templates with PHP tags are converted to patterns).

## To Test

Complete the following actions to ensure that the resulting theme matches expected results:

* Make changes to Global Styles and save the theme (changes should be reflected in the theme.json file)
* Make changes to templates and save the theme 
* * Image assets should be included in the theme, referenced from the pattern, which is referenced from the template
* * Paragraph blocks should be localized in a pattern and referenced from a template
* * templates with NO images or paragraphs shouldn't be converted to a pattern
* Export a saved theme
* Clone a theme
* Create a Blank Theme


admin/create-theme/theme-block.php
madhusudhand, line 1: this can be removed.

admin/create-theme/theme-media.php
```diff
@@ -18,105 +19,55 @@ public static function get_media_folder_path_from_url( $url ) {
...
 	}
 
-	public static function get_media_absolute_urls_from_blocks( $flatten_blocks ) {
+	/**
+	 * Get the absolute URLs of all media files for a template
+	 */
+	public static function get_media_absolute_urls_from_template( $template ) {
+
+		$template_blocks = parse_blocks( $template->content );
+		$blocks          = _flatten_blocks( $template_blocks );
```
madhusudhand, line 28: Just a note:
same parsing is being done while trying to remove site specific attributes [here](https://github.com/WordPress/create-block-theme/pull/524/files). All of it can be done in a single place to avoid redundancy of the heavy operations. I am going to iterate on the other PR once this is merged.

admin/create-theme/theme-templates.php
```diff
@@ -113,78 +161,127 @@ public static function add_templates_to_local( $export_type ) {
...
-				// Write the pattern
-				file_put_contents(
-					get_stylesheet_directory() . DIRECTORY_SEPARATOR . 'patterns' . DIRECTORY_SEPARATOR . $template_data->slug . '.php',
-					$pattern['content']
-				);
-			}
+			$template = Theme_Media::make_template_images_local( $template );
+			$template = self::escape_text_in_template( $template );
+			$template = self::eliminate_environment_specific_content( $template );
+			$template = self::paternize_template( $template );
```
madhusudhand: I believe `Save Changes` feature should not alter the template structure (probably not done right), specially converting the template to a pattern. It changes the editing experience. I think it should keep the content as is while saving, including escaping of text since the data is being saved to the current site.

All the escaping and patternization should happen with other export choices that take the theme out of the site.

pbking: I'm not sure that it does change the editing experience.  To the user the content being in a pattern or a template makes the content just as editable. Just as save-able.

And we wouldn't have the opportunity to escape the text, or use included media, if we don't 'save' that content to a pattern.

admin/create-theme/theme-media.php
pbking: I don't think this particular function modifies anything (site specific or not), it's just gathering up all of the links to media so that the location of the media can be transitioned.

I do agree though that we should be cleaning out all the site-specific stuff in a single swoop to keep the processing down.

admin/create-theme/theme-templates.php
```diff
@@ -113,78 +161,127 @@ public static function add_templates_to_local( $export_type ) {
...
-			// Write the media assets
-			Theme_Media::add_media_to_local( $template_data->media );
+	public static function escape_text( $text ) {
+		if ( ! $text ) {
+			return $text;
 		}
+		return "<?php echo __('" . $text . "', '" . wp_get_theme()->get( 'TextDomain' ) . "');?>";
+	}
+
+	public static function eliminate_environment_specific_content( $template ) {
```
pbking: I think the logic you are working on [here](https://github.com/WordPress/create-block-theme/pull/524/files) would drop in nicely here.

admin/create-theme/theme-templates.php
```diff
@@ -103,88 +124,219 @@ public static function clear_user_templates_customizations() {
...
+					$blocks[ $key ]['attrs']['className'] = preg_replace( '/wp-image-\d+/', '', $block['attrs']['className'] );
+				}
+			}
+
+			// set taxQuery to null for query blocks
+			if ( 'core/query' === $block['blockName'] ) {
+				if ( isset( $block['attrs']['taxQuery'] ) ) {
+					unset( $blocks[ $key ]['attrs']['taxQuery'] );
+				}
+				if ( isset( $block['attrs']['queryId'] ) ) {
```
madhusudhand: We would need to preserve `queryId` for [this reason](https://github.com/WordPress/wordpress-develop/pull/3899#discussion_r1104638014).

pbking: I removed that bit.

admin/create-theme/theme-templates.php
pbking: That logic has been added.  I'm going to close that PR since the logic is included here.  If anything is lacking as of now we can iterate on this branch or trunk.

mikachan (commented):
This is a big PR! But I know it's because it's a combination of multiple branches. Here's my test results from following the test instructions. All these actions are from the options within the Site Editor.

#### Make changes to Global Styles and save the theme:
Changes reflected in the theme.json file ✅

#### Make changes to templates and save the theme:
- Image assets included in the theme, referenced from the pattern, which is referenced from the template ❓

~~I noticed some characters in the image alt text are not correctly encoded:~~ (Sorry, I think this is correct as it's coming through as HTML character codes within the alt text, which is what I'd expect.)

image

They're also not localized, but I think this could happen in a follow-up (I'm happy to open a separate issue).

When testing with a Cover block, I noticed that there are some encoding issues with the start and end of the PHP tag for referencing an image: 

```
<!-- wp:cover {"url":"\u003c?php echo esc_url( get_stylesheet_directory_uri() ); ?\u003e/assets/images/flying-w.jpg","dimRatio":50,"customOverlayColor":"#88817d","isDark":false,"layout":{"type":"constrained"}} -->
```

- Paragraph blocks localized in a pattern and referenced from a template ❌

I noticed when the paragraph is localized, some characters are not encoded correctly:

```
<!-- wp:paragraph --><p><?php echo __('The page youâ&#128;&#153;re looking doesnâ&#128;&#153;t seem to exist. Maybe try a search?', 'pixl');?></p>
<!-- /wp:paragraph -->
```

image

Otherwise, this looks like it's working well.

- Templates with NO images or paragraphs not converted to a pattern ✅

#### Export a saved theme ✅
#### Clone a theme ✅
#### Create a Blank Theme ✅

I also tested based on the proposed changes from https://github.com/WordPress/create-block-theme/issues/162:

- Remove `ref` attributes from the Navigation block ✅
- Remove `id` attributes from Image and Cover blocks ✅
- Remove `wp-image-[id]` classes from Image and Cover blocks ✅
- Remove `category` IDs from query blocks ✅
- Remove `taxQuery` attribute from query blocks ❌ `taxQuery` is not removed from the block

In summary, looks like there are some small issues with encoding and the `taxQuery` attribute but overall this is testing really well. Thank you for working on this, the code is looking much cleaner and better organised. Looking forward to this landing, especially the changes related to removing the site-specific attributes ✨

tests/test-theme-media.php
```diff
@@ -0,0 +1,45 @@
...
+					<img class="wp-block-cover__image-background wp-image-628" alt="" src="http://example.com/image.jpg" data-object-fit="cover"/>
+					<div class="wp-block-cover__inner-container">
+					</div>
+				</div>
+			<!-- /wp:cover -->
+		';
+		$new_template      = Theme_Media::make_template_images_local( $template );
+
+		// The image should be replaced with a relative URL
+		$this->assertStringNotContainsString( 'http://example.com/image.jpg', $new_template->content );
```
mikachan, line 37: Maybe we could test the character encoding as part of this test.

tests/test-theme-templates.php
```diff
@@ -0,0 +1,114 @@
+<?php
+/**
+ * @package Create_Block_Theme
+ */
+class Test_Create_Block_Theme_Templates extends WP_UnitTestCase {
```
mikachan, line 5: Thanks for including so many tests 🤌

tests/test-theme-templates.php
```diff
@@ -0,0 +1,114 @@
...
+		$this->assertStringNotContainsString( '<!-- wp:image {"id":635} -->', $new_template->content );
+		$this->assertStringNotContainsString( 'wp-image-635', $new_template->content );
+	}
+
+	// TODO: I'm not sure of the proper way to format this property for testing or now to cause it to be
+	// added via the Global Styles Panel.
+	// public function test_eliminate_taxQuery_from_query_loop() {
+	// 	$template          = new stdClass();
+	// 	$template->content = '
+	// 	<!-- wp:query {"queryId":43,"query":{"perPage":10,"pages":0,"offset":0,"postType":"post","order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":true},"metadata":{"categories":["posts"]}} -->
```
mikachan: Here's an example with `taxQuery`:

```
<!-- wp:query {"queryId":20,"query":{"perPage":3,"pages":0,"offset":0,"postType":"post","order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":false,"taxQuery":{"post_tag":[9]}}} -->
```

mikachan: This attribute is added if you select a filter based on taxonomies in the block settings for a query block:

image

tests/test-theme-media.php
pbking: Ahh, good idea.  I shall.

pbking: Regarding: 
> I noticed some characters in image alt test are not correctly encoded:

I tried a few tests and attempted to reproduce, but nothing I did seemed to change what was already in the template I was copying from.  Can you share what alt text was getting transferred incorrectly?

Can you share

mikachan: Yeah sure, it was this alt text: Test test "this is a" ><<""""'''T3sT

image

The less than/greater than characters are less of a priority, but I think the double quotes are probably worth handling.

tests/test-theme-media.php
```diff
@@ -0,0 +1,45 @@
...
+		$template->content = '
+			<!-- wp:image -->
+			<figure class="wp-block-image"><img src="http://example.com/image.jpg" alt="Alternative Text" /></figure>
+			<!-- /wp:image -->
+		';
+		$new_template      = Theme_Media::make_template_images_local( $template );
+
+		// The image should be replaced with a relative URL
+		$this->assertStringNotContainsString( 'http://example.com/image.jpg', $new_template->content );
+		$this->assertStringContainsString( 'get_stylesheet_directory_uri', $new_template->content );
```
vcanales, line 18:
```suggestion
		$this->assertStringContainsString( '<?php echo esc_url( get_stylesheet_directory_uri() ); ?>', $new_template->content );
```

Test for the full string, to catch encoding issues at this stage. Making this modification helped me realize that at least this encoding issue, where `<` and `>` are turned into their unicode equivalents `\u003c` and `\u0033`, is caused somewhere down the line.

tests/test-theme-media.php
```diff
@@ -0,0 +1,45 @@
...
+					<div class="wp-block-cover__inner-container">
+					</div>
+				</div>
+			<!-- /wp:cover -->
+		';
+		$new_template      = Theme_Media::make_template_images_local( $template );
+
+		// The image should be replaced with a relative URL
+		$this->assertStringNotContainsString( 'http://example.com/image.jpg', $new_template->content );
+		$this->assertStringContainsString( 'get_stylesheet_directory_uri', $new_template->content );
```
vcanales, line 38:
```suggestion
		$this->assertStringContainsString( '<?php echo esc_url( get_stylesheet_directory_uri() ); ?>', $new_template->content );
```

Same suggestions as above; test for the full php line.

tests/test-theme-media.php
mikachan: Ah wait, this is correct for HTML, right?! You'd want the HTML character codes to be used in this case. Sorry, ignore! I think the only issue here is lack of localization, but that can happen in a follow-up.

vcanales: The issue with php tags being turned into unicode is fixed by [f97a8a8](https://github.com/WordPress/create-block-theme/pull/522/commits/f97a8a81dd3d857af1305f8c9b4449e9c8eb0cb4) — The issue was that by moving assets to the theme as a first step when preparing the template for export, we were passing php code to WP functions that don't work with it, such as `serialize_blocks`.

madhusudhand:
While testing, observed that patronizer copies content into new patterns instead of re-using existing patterns. 
created a separate issue #533, as this isn't a newly introduced behavior.

tests/test-theme-media.php
pbking: Added a test that runs the template through the whole process to verify encoding [here](https://github.com/WordPress/create-block-theme/pull/522/commits/649ba9d952b9d83ceed5e41261951ee2051ef6cd).

tests/test-theme-media.php
pbking: Similar test added [here](https://github.com/WordPress/create-block-theme/pull/522/commits/649ba9d952b9d83ceed5e41261951ee2051ef6cd).

tests/test-theme-templates.php
pbking: Thank you @mikachan 
I have added that to the test suite [here](https://github.com/WordPress/create-block-theme/pull/522/commits/1d48804ca0954de85df581ad1bb98f5519b0fd95) (which did what tests are supposed to do and showed me an error. :) )

tests/test-theme-templates.php
pbking: Writing code with tests is SO fun!

tests/test-theme-media.php
pbking: An encoding test has been added, and thank you hugely @vcanales for figuring out why it was busted!

madhusudhand (approved):
Verified the changes for all the exported options and changes to templates. It works as expected. 🚢

mikachan (approved):
This is working well for me too, I think this is good to bring in now 🎉

vcanales (approved):
+1 to previous comments; this is testing well for me.

PR #528, opened 2024-03-21
Remove/font management & wp-admin forms

pbking:
This change removes all of the wp-admin forms for Create Block Theme including font management and theme management.

The Sidebar tools should continue to build and work as expected.

A new "admin-landing" app has been stubbed out to replace it.

image


creativecoder:
I think we should get all of the options on the wp-admin form into the site editor before removing the form. It doesn't look like creating a theme variation or child theme is possible in the site editor right now. I recall seeing the child theme option used in the theme builder conversations from last year.

pbking:
> I think we should get all of the options on the wp-admin form into the site editor before removing the form. It doesn't look like creating a theme variation or child theme is possible in the site editor right now. 

Yes, both of those features need to be in the site editor before we release.  I'll get both of those features added to the list; they should both be pretty easy to add.  

I don't think we need to wait to remove this code though; the admin will need to be gone before v2 ships and the features need to be in before v2 ships.

vcanales (commented):
I'm don't think we should remove the current admin page without first creating the replacement. Am I missing something in terms of the benefit of doing this ahead of starting that work?

madhusudhand:
Now that #531 is merged, which  brings all the export options to editor, admin panel UI could be replaced.

pbking:
Closing this.  Half of it is already in with #595 and the rest should just be done again when a replacement exists.

PR #532, opened 2024-03-22
Add/create theme variation to site editor interface

### LINKED ISSUE #529: Create Theme Variation from Site Editor
Currently the only option to create a theme variation is from the wp-admin interface which is going away.

Users need an option to create a theme variation from the site editor CBT side panel.

Users should be prompted for the Name they want the variation to be, that should be used in the resulting JSON file and converted to a slug and used as the filename.

pbking:
Fixes #529

Adds ability to create a theme variation from the site editor interface:

image

To Test:

Make changes to Global Styles using the Global Styles interface.  (Do not persist those changes to your theme).

Save a theme variation using the new interface.  

Ensure that the name given the variation (using the form above it) is reflected in the JSON.
Ensure that the styles changed are present in the JSON.
Ensure that the style variation is available to the user in the Global Styles panel.

Should not be merged until https://github.com/WordPress/create-block-theme/pull/522 lands

### DIFF
--- a/admin/create-theme/theme-json.php
+++ b/admin/create-theme/theme-json.php
@@ -10,33 +10,27 @@ class Theme_Json {
 	}
 
 	public static function add_theme_json_variation_to_local( $export_type, $theme ) {
-		$variation_slug = sanitize_title( $theme['variation'] );
 		$variation_path = get_stylesheet_directory() . DIRECTORY_SEPARATOR . 'styles' . DIRECTORY_SEPARATOR;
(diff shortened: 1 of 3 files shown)


mikachan (approved):
These changes are working well for me following the test instructions 🎉 The code also looks good. Looks like we need to resolve some conflicts with the new child theme endpoint PR before we can bring this in.

madhusudhand (approved):
Resolved conflicts.
Changes LGTM!

PR #539, opened 2024-04-03
Replace font management with screen pointing to native font library

madhusudhand:
This change replaces the font management screen with links pointing to native font library.

image

This is an immediate change for 2.0
Clean up of the font management code is being done in #528

### DIFF
--- a/src/manage-fonts/index.js
+++ b/src/manage-fonts/index.js
@@ -1,5 +1,5 @@
 import { useState, useEffect } from '@wordpress/element';
-import { __ } from '@wordpress/i18n';
+import { __, _x } from '@wordpress/i18n';
 import FontFamily from './font-family';
 
 import DemoTextInput from '../demo-text-input';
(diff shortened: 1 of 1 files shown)


vcanales (approved):
Tinkered a bit with the copy, and made it translatable. LGTM

mikachan (approved):
LGTM, too! Great, simple design 👏 

I don't think we have a better link than the 6.5 release post, we can always update this later if needed.

t-hamano:
If I understand correctly, this looks like a breaking change.

- If the user has already installed fonts previously via this plugin's font management, will these fonts be inherited by the native font library?
- If fonts are installed through the font management of this plugin, is there a way to delete it from the UI?
- What happens if a user has already activated a font in this plugin's font management and then activates the same font in the native font library?

pbking:
> If I understand correctly, this looks like a breaking change.

I consider this a "breaking change" in that the functionality to manage fonts is now done differently. (Which is why we are releasing it with a major version change.) However, any theme previously managed fonts with CBT (or any other method) is completely compatible with this change and can continue to be managed with CBT.

> If the user has already installed fonts previously via this plugin's font management, will these fonts be inherited by the native font library?

Yes.  If a user installed fonts via CBT 1.x then those assets are included in the theme and configured via `theme.json`.  The native Font Library management system will understand and use that with or without CBT.  Those fonts can be activated/deactivated with that tool.  If CBT is used to further manage those fonts (combined with Core's Font Library management) then the theme exported will include those assets (for any activated font) and apply the correct configuration in `theme.json` just as it had before.

> If fonts are installed through the font management of this plugin, is there a way to delete it from the UI?

Yes, any font that is deactivated via the Font Library UI will be removed from the theme when the theme is saved via CBT. (See [instructions](https://github.com/WordPress/create-block-theme/pull/518) on this issue where the functionality was migrated).

> What happens if a user has already activated a font in this plugin's font management and then activates the same font in the native font library?

To clarify CBT *installs* fonts in a theme (copies the assets to the appropriate folder and includes the configuration to use those assets in `theme.json`).  The ability to "activate and deactivate" fonts is something that Core introduced in the Font Library.   Both 1.x and 2.x versions of CBT allow fonts to be installed and uninstalled.  To further answer your question, if a user has installed a font via the 1.x version of this plugin (or any other method) and a user attempts to install and activate the same font in the native font library they will be combined in the theme such that only one of each Font Face should be installed.

pbking:
I will add that additional instructions and documentation as to how to achieve all of the above is very necessary!

t-hamano:
@pbking Thank you for letting me know in detail. I didn't know that there was already an implementation for backward compatibility. If CBT's font management was fully compatible with the native font library, I think it would be perfectly fine.

PR #540, opened 2024-04-04
add deprecation notice in theme export admin screen

madhusudhand:
This PR introduces a notice about removal of theme export options from admin screen.

image


mikachan (approved):
This looks good to me and I like the idea of adding a deprecation notice to this page in the next release. This gives users a more gradual transition to the options in the editor.

I'd like to see what @vcanales thinks but otherwise I'm happy to bring this in.

vcanales (approved):
I like this, it's a good way to gently transition away from this page.

PR #541, opened 2024-04-05
Fix changelog creation script

vcanales:
Previous Changelogs have come up empty; this PR ensures that the changes since the last tag created are correctly added to the Changelog.

1. Don't specify `--merge`, since we're not using merge commits.
2. Prioritize printing the `message` instead of the body; GitHub automatically sets the squashed commit to the PR title and adds the number. This should always be available, but will still add `body` as a fallback just in case something is committed without a PR.
3. Use `git` itself to determine the last tag created, instead of using the package.json file.
4. Other minor fixes — capitalization, order of commits in the output, etc.

Sample output for v2.0.0 from the updated script:

```
= 2.0.0 =
* Add theme json inspector (#520)
* Add deprecation notice in theme export admin screen (#540)
* Replace font management with screen pointing to native font library (#539)
* Added creation of theme validation to site editor interface (#532)
* Add/child theme creation (#531)
* Add blueprint.json file to enable plugin previews (#511)
* Extracted any logic that may need to be tested from the api class (#522)
* Use CORE for Font Management (#518)
* Add integration tests (#393)
* Remove CODE_OF_CONDUCT.md from .distignore. (#515)
* Remove repo specific CoC. (#514)
* GitHub Actions: Add JavaScript Unit Test (#508)
* Add files and directories not needed for release to `.distignore` (#512)
* Replace dash icon with SVG icon (#506)
* Fix browser warning error when clicking the reset button (#505)
* Add markdown and package.json lint command (#504)
* Fix react warning error on font upload page (#502)
* Fix dynamic property deprecation (#501)
* Add text domain to translation target (#499)
```

### Testing:

Run `RELEASE_TYPE=major npm run update-version` — Your changelog file will include the commits from this branch, which is fine; when merged, it'll only use the history from `trunk`.


mikachan (approved):
Thank you for fixing this! It tests well for me. Running `RELEASE_TYPE=major npm run update-version`, I see the commits from this branch, formatted as expected:

image

(The screenshot cuts off the bottom, the last line is "* Add text domain to translation target (#499)".)

I see the screenshot changes are in line with what's described in the [plugin docs here](https://developer.wordpress.org/plugins/wordpress-org/plugin-developer-faq/#where-do-i-put-my-files), so these look fine too.

I think this is good to bring in 👍

creativecoder (commented):
This is also working well for me! I tested the `update-version` command with major and minor versions.

One question, why change screenshots from jpg to png? The png file size is a lot larger.

vcanales:
> One question, why change screenshots from jpg to png? The png file size is a lot larger.

No particular reason, that's just the default for when I took them. I've transformed them to jpeg on [ef6add0](https://github.com/WordPress/create-block-theme/pull/541/commits/ef6add0d9cc94b6ded96e14725625bbc993327de); thanks for spotting!

PR #546, opened 2024-04-09
Add missing build step to deploy workflow

vcanales:
Deploy workflow was missing the npm build steps.


pbking (approved):
Yup.  Need that step. :)

PR #554, opened 2024-04-10
Fix concatenation of translation strings

### LINKED ISSUE #553: [i18n] Translation strings with hardcoded concatenation
The below strings are incomplete parts of sentences, with hardcoded concatenation:

- `You can manage your theme fonts directly from the` + ` ` + `editor` (https://github.com/WordPress/create-block-theme/blob/ac153ae6e8cb13c26ab5a8c3cb6f9d1ab6cf3e84/src/manage-fonts/index.js#L231-L242)
- `For more information, please check out` + ` ` + `<a href="https://wordpress.org/news/2024/04/regina/">{ 'What's inside 6.5' }</a>` (https://github.com/WordPress/create-block-theme/blob/ac153ae6e8cb13c26ab5a8c3cb6f9d1ab6cf3e84/src/manage-fonts/index.js#L249-L254)

Both are like this because should include a link.
The first one is wrapped in gettext `_x()`, the second isn't.

Would be better to turn them into complete translatable strings like this:
- `You can manage your theme fonts directly from the %1$seditor%2$s.`
- `For more information, please check out %1$sWhat's inside 6.5%2$s.`

I've tried using `sprintf()` for that, but the links are getting escaped.
Example:

```
<p>
	{ sprintf(
			// translators: 1: Opening <a> tag. 2: Closing </a> tag.
			__(
				`For more information, please check out %1$sWhat's inside 6.5%2$s.`,
				'create-block-theme'
			),
			'<a href="https://wordpress.org/news/2024/04/regina/">',
			'</a>'
	) }
</p>
```

Output:
imagem

Solved: used `createInterpolateElement()` in PR #554

pedro-mendonca:
Use createInterpolateElement to make complete translation strings without concatenation.

Fixes #553

### DIFF
--- a/src/manage-fonts/index.js
+++ b/src/manage-fonts/index.js
@@ -1,4 +1,4 @@
-import { useState, useEffect } from '@wordpress/element';
+import { useState, useEffect, createInterpolateElement } from '@wordpress/element';
 import { __, _x } from '@wordpress/i18n';
 import FontFamily from './font-family';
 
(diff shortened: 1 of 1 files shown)


creativecoder (commented):
Thanks for addressing this translation bug!

My understanding is that the standard way to solve this is to use [`sprintf` from the `@wordpress/i18n` package], and use placeholders for any markup within the string. (https://github.com/WordPress/gutenberg/tree/trunk/packages/i18n#sprintf).

creativecoder:
> My understanding is that the standard way to solve this is to use [sprintf from the @wordpress/i18n package]

Now I see from https://github.com/WordPress/create-block-theme/issues/553 that you already tried that! Looks like using `createInterpolateElement` is the pattern that's used in Gutenberg for this situation.

src/manage-fonts/index.js
```diff
@@ -228,30 +232,36 @@ export default function () {
...
+							'You can manage your theme fonts directly from the <a>editor</a>.',
 							'create-block-theme'
-						) }
-					</a>
-					.
+						),
+						{
+							a: (
+								// eslint-disable-next-line jsx-a11y/anchor-has-content
+								<a href="/wp-admin/site-editor.php?canvas=edit" />
```
creativecoder, line 243: Using an absolute path like this will break when using non-standard hosting configurations, like putting core files in a subdirectory, or subdirectory multisite installations.

But that's not a new problem with this PR.

creativecoder (approved):
I can see the links correctly generated in wp-admin. Thanks for addressing this.

src/manage-fonts/index.js
pedro-mendonca: I've kept the text and link structure, only improved the i18n of the string.
Feel free to add any changes to the links in the PR.

creativecoder: Thanks, I think we should handle the link structure change separately from this PR.

pbking:
This view has been depreciated and likely to be removed on the next release.  However the work is already done and might be helpful as prior art somewhere.  Bringing this in now.

PR #557, opened 2024-04-10
Try moving the create and export options to the menu flow instead of the create panel

### LINKED ISSUE #538: Create theme UI is confusing with single form and many submit buttons
Create Theme form is bit confusing with single form with too many submit buttons which feels disconnected.

image
image

For example, to create a variation, there is no need of all the information from the form. When trying to create a duplicate variation, it simply throws a message "Variation already exists", there is no clue as to what went wrong.

Preset with a better user experience and ask only relevant information for the given choice.

pbking:
Experimenting with this:
image

image

image

Instead of a bunch of options on the 'Create Theme' panel.

Fixes #538


pbking:
image
image
image
image
image
image
image
image
image

pbking:
Closing this deferring to #563 which is a little cleaner

PR #564, opened 2024-04-16
Include activated Fonts on theme zip export functions

### LINKED ISSUE #558: Font library management no longer stores font in theme folder on theme export
Since version 2.0 exporting fonts no longer stores the font files inside the theme. Instead, the theme references the fonts in the `uploads` directory. 

This is an issue because the fonts are not managed in git / shared across environments.

It would be great to get back the previous handling where font files were moved to the theme folder when exporting :)

pbking:
Previously fonts were only "managed" with Create Block Theme on the "save" action.

This change includes the activated fonts in the exported zip action.

To test activate a font using the Font Manager.

Do NOT 'Save' the theme (that font will be persisted in the theme and this isn't the functionality we're adding).

Export the theme.  Note that any font that is active is included in the exported theme's `theme.json` and the assets it is referencing are included in `/assets/fonts/*`.

Fixes: #558

### DIFF
--- a/admin/class-create-theme.php
+++ b/admin/class-create-theme.php
@@ -109,7 +109,7 @@ class Create_Block_Theme_Admin {
 
 		$zip = Theme_Zip::copy_theme_to_zip( $zip, null, null );
 		$zip = Theme_Zip::add_templates_to_zip( $zip, 'current', $theme['slug'] );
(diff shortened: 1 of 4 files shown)


vcanales (approved):
Installed new fonts with the Font Library, activated them, and then exported the theme; assets are included with the export. 

LGTM

PR #567, opened 2024-04-17
Fix/un transposed patterns

### LINKED ISSUE #566: Cloning a theme causes critical error
Cloning a theme seems to cause a critical error:

`Parse error: syntax error, unexpected token "-", expecting "(" in /Users/sarahnorris/Source/Automattic/a8c-projects/themes/tt4-2/functions.php on line 22`

image

You do not need to make any changes to the theme - just attempting to clone seems to reproduce the error. I've tested this with Twenty Twenty-Four and Programme.

Originally mentioned on Slack here: https://wordpress.slack.com/archives/C015GUFFC00/p1712938209077399

pbking:
When a pattern is created during the cloning action (for example if the source theme has templates with text content and the template is processed to localize the text) then the patterns that are created have the SOURCE theme's namespace rather than the new theme.

This change passes that new slug value when creating the pattern so that it is used instead.

To test:
* Activate the Programme theme (which has un-localized content in the index template)
* Create a CLONE of Programme using CBT.  When the cloned theme is activated it should work as expected.
* Observe all of the created patterns and note that the namespace is correct (the slug based on the name given when cloning the theme)

Also the slug generated from the new theme name is no longer using -'s to fix the issue noted below. Fixes: #566

### DIFF
--- a/admin/create-theme/theme-templates.php
+++ b/admin/create-theme/theme-templates.php
@@ -131,10 +131,10 @@ class Theme_Templates {
 	 * @param object $template The template to extract content from.
 	 * @return object The template with the patternized content.
 	 */
-	public static function paternize_template( $template ) {
+	public static function paternize_template( $template, $slug = null ) {
 		// If there is any PHP in the template then paternize
 		if ( str_contains( $template->content, '<?php' ) ) {
-			$pattern                 = Theme_Patterns::pattern_from_template( $template );
+			$pattern                 = Theme_Patterns::pattern_from_template( $template, $slug );
 			$pattern_link_attributes = array(
 				'slug' => $pattern['slug'],
 			);
@@ -173,11 +173,11 @@ class Theme_Templates {
 			$template = Theme_Media::make_template_images_local( $template );
 		}
 
-		$template = self::paternize_template( $template );
-
 		if ( $slug ) {
 			$template = self::replace_template_namespace( $template, $slug );
 		}
+
+		$template = self::paternize_template( $template, $slug );
 
 		return $template;
 	}
--- a/admin/create-theme/theme-utils.php
+++ b/admin/create-theme/theme-utils.php
@@ -30,12 +30,9 @@ class Theme_Utils {
 		return $extension;
 	}
 
-	public static function replace_namespace( $content, $new_slug, $new_name ) {
-		$theme               = wp_get_theme();
-		$old_slug            = $theme->get( 'TextDomain' );
+	public static function replace_namespace( $content, $old_slug, $new_slug, $old_name, $new_name ) {
 		$new_slug_underscore = str_replace( '-', '_', $new_slug );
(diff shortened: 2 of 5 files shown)


mikachan (commented):
I've just given this a test using Programme, naming the cloned theme Programme Two, but I'm seeing a few issues:

- The cloned theme's slug used in patterns is `programme-two-two`, I think it should be `programme-two`
image

- I saw this error in the editor after the page reload: https://github.com/WordPress/create-block-theme/issues/566 - `Parse error: syntax error, unexpected token "-", expecting "(" in /themes/programme-two/functions.php on line 21`

Looks like `programme-two` is being used as the theme name in the functions.php file, when I guess it should be `programme_two`:

image

pbking:
This has been updated to address what you mentioned here:

> Looks like programme-two is being used as the theme name in the functions.php file, when I guess it should be programme_two

which was reported in #566 

This is ready for another test.

mikachan (commented):
Thanks! I've given this another test.

I'm now seeing `programmetwotwo` as the theme slug in the pattern block:
image

I was expecting this to be `programme-two` or `programmetwo`. I can see the new theme directory name is `programmetwo`, which looks correct.

The issue with the function names in functions.php has been fixed 🎉

I noticed that the hyphen has been removed rather than replaced with an underscore. I think this is fine, just wanted to check if this was the expected behaviour.

image

pbking:
> I noticed that the hyphen has been removed rather than replaced with an underscore. I think this is fine, just wanted to check if this was the expected behaviour.

Yes, though it's unfortunate.  It's impossible (difficult?  outside of my ability?) to determine which strings get the - and which get the _ so we just have to use neither. :( 

This was the original behavior of CBT.  It would be nice to improve it in the future though.

pbking:
> I'm now seeing programmetwotwo as the theme slug in the pattern block:

This was weird.  The problem was because the NEW slug (programmetwo) contained the OLD slug (programme).  And the "paternize" function happened BEFORE the namespace swap so when doing a string replace of programme>programmetwo the slug that was in the paternized template programmetwo became programmetwotwo.

I fixed it by making sure that the patternization of the template happens LAST (after the namespace swap).

vcanales (commented):
Tested: 
✅ Fix for the problem with themes such as "Program Two"
✅ Spaces don't create broken function names

Something to keep an eye out for is that if the theme is given a name that includes an actual underscore, those are not being replaced with dashes, creating Text Domains with underscores.

eg. this theme was given the name "Alter_2"
image

This is important because the text domain has to match the Theme's slug in the wp.org theme directory.

mikachan (approved):
Looks like this is working well for me now:

image

I used slightly different testing steps, as since the rebase I think most of the patternisation is done optionally. So instead of just cloning the theme, I saved the theme with all save options selected, and then cloned it. I think this is ready to bring in.

PR #572, opened 2024-04-18
Save only templates that have been changed

pbking:
This change ONLY persists CHANGED templates on the SAVE action.

This allows a user to make changes to the template OUTSIDE of the Editor (if desired).  Re-processing of these templates might not be desired.

Note: The downside of the way this currently works is that any processing actions the user has selected (such as localizing text or media assets) will ONLY be done on templates that have been changed by the user in some way.

To Test:

* Activate a theme.
* Change a template directly (not using the Site Editor) in a way that may be processed (such as adding a paragraph tag)
* Save the Theme using Create Block Theme with "localize text" selected.
* Note that the template remains unmodified.
* Change the template with the site editor (for example, changing the content of the paragraph tag)
* Save the Theme using Create Block Theme with "localize text" selected.
* Note that the template has been changed (it points to a pattern and the pattern contains localized text).

### DIFF
--- a/src/editor-sidebar/save-panel.js
+++ b/src/editor-sidebar/save-panel.js
@@ -18,6 +18,7 @@ export const SaveThemePanel = () => {
 	const [ saveOptions, setSaveOptions ] = useState( {
 		saveStyle: true,
 		saveTemplates: true,
+		processOnlySavedTemplates: true,
 		saveFonts: true,
 		removeNavRefs: false,
 		localizeText: false,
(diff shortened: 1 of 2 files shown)


pbking:
Per @mikachan suggestion I have added a control to make this optional which kind of negates the negatives.  It defaults to true (which will only process the templates with changes).

image

src/editor-sidebar/save-panel.js
```diff
@@ -96,6 +97,22 @@ export const SaveThemePanel = () => {
 						} );
 					} }
 				/>
+				<CheckboxControl
+					label="Process Only Saved Templates"
```
mikachan: We're using the word "save/saved" a lot in this panel, so I'm wondering if swapping "saved" for "modified" here would make things more clear.

```suggestion
					label="Process Only Modified Templates"
```

mikachan (approved):
This works as described 🎉

I've left one small suggestion but otherwise I think this is good to bring in.

src/editor-sidebar/save-panel.js
pbking: Yes, I agree.  Good call.

PR #575, opened 2024-04-21
i18n: Make modal titles translatable

t-hamano:
This PR makes the titles of metadata modals and theme.json modals translatable.

image
image


mikachan (approved):
Looks good, thanks @t-hamano!

PR #576, opened 2024-04-21
Fix hardcoded wp-admin URLs

### LINKED ISSUE #561: Hardcoded URLs to the Site Editor breaks links on some installs
On these admin pages, the links to the Site Editor are broken:

- **Appearance > Create Block Theme:** https://github.com/WordPress/create-block-theme/blob/v2.0.2/admin/create-theme/theme-form.php#L24
-  **Appearance > Manage Theme Fonts:** https://github.com/WordPress/create-block-theme/blob/v2.0.2/src/manage-fonts/index.js#L235

In both cases, the URLs are hardcoded and assume the WordPress installation is relative to the root of the site (`/wp-admin/site-editor.php?canvas=edit`).

Instead, they should use a function like `admin_url()` to reference the correct location of the pages in the admin (not sure of the JS equivalent offhand): https://developer.wordpress.org/reference/functions/admin_url/

t-hamano:
Fixes #561

This PR fixes the following two links so that they work correctly in an environment installed in a subdirectory.

http://localhost:8888/wp-admin/themes.php?page=manage-fonts

manage-fonts

http://localhost:8888/wp-admin/themes.php?page=create-block-theme

create-block-theme

## Testing Instructions

It's a bit complicated, but in order to actually test this PR, you'll need an environment with WordPress installed in a subdirectory. There are several ways to do this, but [this handbook page](https://developer.wordpress.org/advanced-administration/server/wordpress-in-directory/) will help you set up your environment.

### DIFF
--- a/src/manage-fonts/index.js
+++ b/src/manage-fonts/index.js
@@ -240,14 +240,17 @@ export default function () {
 						{
 							a: (
 								// eslint-disable-next-line jsx-a11y/anchor-has-content
-								<a href="/wp-admin/site-editor.php?canvas=edit" />
+								<a href="site-editor.php?canvas=edit" />
 							),
 						}
 					) }
 				</p>
 				<img
 					src="https://i0.wp.com/wordpress.org/news/files/2024/04/Font-Manager-2.png?w=620&ssl=1"
-					alt="WordPress 6.5 Font Library"
+					alt={ __(
+						'WordPress 6.5 Font Library',
+						'create-block-theme'
+					) }
 				/>
 				<p>
 					{ createInterpolateElement(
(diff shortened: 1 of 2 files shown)


src/manage-fonts/index.js
```diff
@@ -240,14 +240,17 @@ export default function () {
...
 						}
 					) }
 				</p>
 				<img
 					src="https://i0.wp.com/wordpress.org/news/files/2024/04/Font-Manager-2.png?w=620&ssl=1"
-					alt="WordPress 6.5 Font Library"
+					alt={ __(
+						'WordPress 6.5 Font Library',
+						'create-block-theme'
+					) }
```
t-hamano, line 253: Minor fix

mikachan (approved):
These changes test well for me and the code looks good. Thanks for the fix, @t-hamano!

PR #577, opened 2024-04-21
Code Quality: Remove `no-undef` eslint rule

t-hamano:
This PR removes the disabled `no-undef` rule in the eslint rules and fixes the lint errors that occurred with it.

Disabling the `no-undef` rule globally is dangerous. This is because you can use variables and functions that have not been previously defined or imported, making it easier to overlook critical errors.

In fact, there were already variables and functions that were not defined. The reason these didn't cause an error was probably because the variable or function happened to be defined in another JavaScript.

Instead of removing the `no-undef` rule, I added the following configuration to the eslint:

- `globals`: Suppresses warning errors that occur in the two constructors `FontFace` and `FileReader`.
- `overrides`: To prevent `it` and `describe` from causing warning errors in unit tests, add unit test rules to the javascript files in the `test` directory.

## Testing Instructions

This PR does not add or change any functionality, so it should work correctly as before, but please test the following:

- Write code that does not follow the eslint rules to ensure that the eslint itself works as before. For example, inject a component like `<Test />` that doesn't exist. When you run `npm run lint:js` you should get the following error:
  `error  'Test' is not defined  react/jsx-no-undef`
- Write a function like `test()` that is not predefined. trunk does not warn this statement as an error, but when you run `npm run :lint` on this PR, you should get the following error:
  `error  'test' is not defined  no-undef`

### DIFF
--- a/src/editor-sidebar/metadata-editor-modal.js
+++ b/src/editor-sidebar/metadata-editor-modal.js
@@ -1,6 +1,7 @@
 import { __ } from '@wordpress/i18n';
 import { useState } from '@wordpress/element';
-import { useSelect } from '@wordpress/data';
+import { useSelect, useDispatch } from '@wordpress/data';
+import { store as noticesStore } from '@wordpress/notices';
 import {
 	// eslint-disable-next-line
 	__experimentalVStack as VStack,
@@ -27,6 +28,8 @@ export const ThemeMetadataEditorModal = ( { onRequestClose } ) => {
 		tags_custom: '',
 		recommended_plugins: '',
 	} );
+
+	const { createErrorNotice } = useDispatch( noticesStore );
 
 	useSelect( async ( select ) => {
 		const themeData = select( 'core' ).getCurrentTheme();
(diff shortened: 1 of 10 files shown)


src/editor-sidebar/metadata-editor-modal.js
```diff
@@ -28,6 +29,8 @@ export const ThemeMetadataEditorModal = ( { onRequestClose } ) => {
 		recommended_plugins: '',
 	} );
 
+	const { createErrorNotice } = useDispatch( noticesStore );
```
t-hamano, line 32: This is one example of preventing a critical error. If this code is missing and the theme metadata update fails, an error like the one below will occur and no message will be displayed.

```
Uncaught (in promise) ReferenceError: createErrorNotice is not defined
```

pbking:
Thank you.  I have followed the reasoning back when [the linting rules were added](https://github.com/WordPress/create-block-theme/pull/196) and I see how this is a better (and important fix).

Everything continues to build and run as expected following this change.  

🚢

pbking (approved):

PR #582, opened 2024-04-23
Fix error when exporting theme with no fonts

pbking:
When there are no fonts to export an error is thrown (null ref).  
This change checks for fonts to copy to the theme zip before trying to.

Found the bug while testing [this reported issue](https://wordpress.org/support/topic/create-child-theme-bug/).

Fixes #585 

To test: 

* Activate Twenty Twenty Four
* Make a minor change with Global Styles (such as background color) and save Global Styles. (Do not add fonts.)
* Create a Child Theme (Create Theme > Create Child Theme)
* The theme should be successfully created.
* Export the Theme with CBT
* The theme should be successfully exported.

PR #586, opened 2024-04-24
Process group background image when saving theme

pbking:
When saving a theme with "localize images" selected the background image of a group block should be included.

To test:

* Edit a template in a theme and add a group block
* Add a background image to that group block
* Save your changes (to the user space)
* Using CBT save your changes (to the theme).  Make sure "Localize Images" is selected
* Observe our theme files.  Note that the template that was edited uses a pattern, the pattern has the group block with an image referenced from the /assets folder
* Export the theme.  Note that the template, pattern and assets are correct

### DIFF
--- a/admin/create-theme/theme-media.php
+++ b/admin/create-theme/theme-media.php
@@ -73,6 +73,13 @@ class Theme_Media {
 					}
 				}
 			}
+
+			// Gets the absolute URLs of background images in these blocks
+			if ( 'core/group' === $block['blockName'] ) {
+				if ( isset( $block['attrs']['style']['background']['backgroundImage']['url'] ) && Theme_Utils::is_absolute_url( $block['attrs']['style']['background']['backgroundImage']['url'] ) ) {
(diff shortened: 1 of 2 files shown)


vcanales:
After following the test steps provided, I somehow end up with an error when trying to export the zip file.

```
POST http://localhost:8888/index.php?rest_route=%2Fcreate-block-theme%2Fv1%2Fexport&_locale=user 500 (Internal Server Error)
```

As far as I can. tell, `rest_route` has no value assigned to it on that URL. Not sure how it's related, but going back to trunk everything works normally. Any ideas?


image

pbking:
I suspect the error was the same as was fixed [here](https://github.com/WordPress/create-block-theme/pull/582).

That fix was merged into trunk, but this work was branched from trunk before it was merged.  I have rebased trunk and expect it will work as expected now.

vcanales (approved):
Working as expected 👍

PR #587, opened 2024-04-24
Fix cover block nested content stripped

### LINKED ISSUE #552: Saving template with nested cover blocks to the theme wipes the contents of the inner cover block
~~If you create a template part, add a cover block to that template part and add blocks inside the cover block, saving the changes to the theme with Create Block Theme will wipe the contents of the cover block.~~

After running into this issue on a separate project, it seems to happen when you have nested cover blocks in a template. See [this comment](https://github.com/WordPress/create-block-theme/issues/552#issuecomment-2074712734) for updated steps to reproduce.

pbking:
Fixes #552

When a Cover Block has nested content (of a certain type... unsure exactly what the situation is that causes it) the content is sometimes removed when processing the template for save.

Test using steps outlined in #552.  Unit tests added to reproduce the issue (with content provided).

### DIFF
--- a/admin/create-theme/theme-templates.php
+++ b/admin/create-theme/theme-templates.php
@@ -389,6 +389,9 @@ class Theme_Templates {
 				unset( $block['attrs']['id'] );
 				// remove wp-image-[id] class from inner content
 				foreach ( $block['innerContent'] as $inner_key => $inner_content ) {
+					if ( is_null( $inner_content ) ) {
+						continue;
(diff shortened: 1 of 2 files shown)


mikachan (approved):
I followed the testing steps in [this comment](https://github.com/WordPress/create-block-theme/issues/552#issuecomment-2074712734), and I could successfully reproduce the issue. This PR fixes it 🎉

PR #595, opened 2024-04-25
Remove font management

vcanales:
Removes font management code, since this functionality has been replaced by the native Font Library in Core, and therefore been deprecated.

This decouples the removal of font management from the removal of other deprecated parts of the Plugin, which is proposed in https://github.com/WordPress/create-block-theme/pull/528 but is on hold pending feature parity with the features in the editor, and an admin page that points to the new workflow.

### Testing

Make sure that the following commands run without errors:

1. Lint
`npm run lint:php && npm run lint:js`

1. Build
`npm run build` and/or `npm run start`

1. Test
`npm run test:php && npm run test:unit`

After the above, load the Plugin into your dev environment and make sure that no errors are present in the Admin, editor, and in the browser's DevTools Console.


src/editor-sidebar/create-panel.js
```diff
@@ -34,19 +34,22 @@ export const CreateThemePanel = ( { createType } ) => {
...
-			subfolder:
-				themeData.stylesheet.lastIndexOf( '/' ) > 1
-					? themeData.stylesheet.substring(
-							0,
-							themeData.stylesheet.lastIndexOf( '/' )
-					  )
-					: '',
-		} );
-	}, [] );
+	useSelect(
```
vcanales: This is a lint autofix.

pbking: This is the only change to create-panel but now I'm seeing an error thrown.

pbking: I rolled it back to what's in trunk and the error is gone.

pbking: I rolled this back so that the branch was in a working state... 
This change broke the app and I'm not sure why.  I also attempted to remove the array rather than add `theme` to it with the same problem.

I'm not sure how to proceed but my 'druthers is to leave the warning for now.

pbking:
There are still SOME font things that CBT does to manage the fonts on save and export.  I've returned those to the codebase along with the tests.

I also removed some files not used as well as a few functions from the util.js.  The only unit test that was in place was for a function removed so rather than dismantling the unit test framework I just stuck a sample test in there for now.  (I still don't think there's anything valuable to test in the view right now, but there might be someday.)

Lastly, I returned the create-panel.js to how it was on /trunk.  THe change caused the create panel to break for me when I tried to use it with the change.  Comments about that above.

pbking (approved):
As this exists now I think it's good to come in.

PR #598, opened 2024-04-26
Move screenshots to display in wordpress.org plugin page

matiasbenedetto:
## What?
Move screenshots image files location.

## Why?
To display in wordpress.org plugin page.


vcanales (approved):
I had a hard time finding why this was necessary; for context, I believe it's because the GitHub action we use to deploy the plugin to dotorg handles the movement of assets, and it uses the `.wordpress-org` directory to do so. This is documented here: https://github.com/10up/action-wordpress-plugin-asset-update?tab=readme-ov-file#optional-environment-variables

PR #601, opened 2024-04-26
Fix custom fonts assets path

### LINKED ISSUE #600: Eported custom fonts path are wrong in the zip
## Steps to reproduce
1. Activate TT3 or TT4 theme.
2. Install a few custom fonts using the font library.
3. Export your theme

This is the resulting zip has the fonts in this path:
```
 twentytwentyfour//assets/fonts//zOL64pLDlL1D99S8g8PtiKchm-VsjOLhZBY.woff2
 ```
 ## Expected:

 The expected path is: 
 ```
 twentytwentyfour/assets/fonts/zOL64pLDlL1D99S8g8PtiKchm-VsjOLhZBY.woff2
 ```

 ## Screencast
 
[Screencast from 30-04-24 12:54:43.webm](https://github.com/WordPress/create-block-theme/assets/1310626/453f258f-6cf9-4a4e-931f-089d6e334e79)

### PR DESCRIPTION
**Pull request by matiasbenedetto**
## What?
Fix the custom fonts assets path in the exported zip.
Fixes: https://github.com/WordPress/create-block-theme/issues/600

## How?
The two extra slashes from the path were removed. This removed the two empty folders being added to the zip file.

## Testing instructions:
See https://github.com/WordPress/create-block-theme/issues/600

### DIFF
--- a/admin/create-theme/theme-zip.php
+++ b/admin/create-theme/theme-zip.php
@@ -51,7 +51,7 @@ class Theme_Zip {
 				$font_dir       = wp_get_font_dir();
 				$font_face_path = path_join( $theme_font_asset_location, $font_filename );
 				if ( str_contains( $font_face['src'], $font_dir['url'] ) ) {
-					$zip->addFileToTheme( $font_dir['path'] . '/' . $font_filename, $font_face_path );
+					$zip->addFileToTheme( path_join( $font_dir['path'], $font_filename ), $font_face_path );
 				} else {
 					// otherwise download it from wherever it is hosted
 					$tmp_file = download_url( $font_face['src'] );
--- a/admin/create-theme/theme-zip.php
+++ b/admin/create-theme/theme-zip.php
@@ -34,7 +34,7 @@ class Theme_Zip {
 		$theme_json = json_decode( $theme_json_string, true );
 
 		$font_families_to_copy     = Theme_Fonts::get_user_activated_fonts();
-		$theme_font_asset_location = '/assets/fonts/';
+		$theme_font_asset_location = 'assets/fonts';
 		$font_slugs_to_remove      = array();
 
 		if ( ! $font_families_to_copy ) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by vcanales** [comment]
I haven't been able to reproduce the issue; how do you suggest looking at the zip file in order to verify that it's including extra empty directories?

File: admin/create-theme/theme-zip.php [hunk 1]
```diff
@@ -47,14 +47,15 @@ public static function add_activated_fonts_to_zip( $zip, $theme_json_string ) {
...
 			$font_slugs_to_remove[] = $font_family['slug'];
 			foreach ( $font_family['fontFace'] as &$font_face ) {
-				$font_filename = basename( $font_face['src'] );
-				$font_dir      = wp_get_font_dir();
+				$font_filename  = basename( $font_face['src'] );
+				$font_dir       = wp_get_font_dir();
+				$font_face_path = path_join( $theme_font_asset_location, $font_filename );
 				if ( str_contains( $font_face['src'], $font_dir['url'] ) ) {
-					$zip->addFileToTheme( $font_dir['path'] . '/' . $font_filename, $theme_font_asset_location . '/' . $font_filename );
+					$zip->addFileToTheme( $font_dir['path'] . '/' . $font_filename, $font_face_path );
```
vcanales: Could we also use `path_join` here?

```suggestion
					$zip->addFileToTheme( path_join( $font_dir['path'], $font_filename ), $font_face_path );
```

vcanales (approved):
Looking at this again. I think it doesn't hurt to guard against extra slashes, even if I can't figure out how to reproduce the exact issue. LGTM, pending considering the suggestion [here](https://github.com/WordPress/create-block-theme/pull/601/files#r1582164072)

admin/create-theme/theme-zip.php
t-hamano: The `path_join()` function [only accepts two arguments](https://github.com/WordPress/wordpress-develop/blob/66b5d25be2b5e69833bacfe8f64de660ffc2bfa9/src/wp-includes/functions.php#L2155), so I think it's like this

```php
$zip->addFileToTheme( path_join( $font_dir['path'], $font_filename ), $font_face_path );
```

matiasbenedetto:
> I haven't been able to reproduce the issue; how do you suggest looking at the zip file in order to verify that it's including extra empty directories?

@vcanales, here is a screencast showing the problem:

[Screencast from 30-04-24 12:54:43.webm](https://github.com/WordPress/create-block-theme/assets/1310626/2cc62d12-6db8-4262-9fb8-4edf183ff6aa)

admin/create-theme/theme-zip.php
vcanales: Thanks for catching that, got the argument list wrong on the suggestion — edited

pbking: I tested with your suggestion @vcanales and that works as expected.  I'm going to commit that suggestion and bring this one in.

pbking (approved):
Works as expected.  Cleans up the code.

🚢

admin/create-theme/theme-zip.php
matiasbenedetto: :+1: added

PR #603, opened 2024-04-27
Editor Sidebar: Make save panel text translatable

t-hamano:
I noticed that the text in the panel below was not translatable, so I applied a translation function.

image


vcanales (approved):
LGTM

PR #606, opened 2024-04-28
Fix problem with zip file creation on Windows

### LINKED ISSUE #602: Problem with zip file creation on Windows
When creating zip files on Windows, e.g. using LocalWP, the plugin creates files called 

* twentytwentythree/partspost-meta.html
* twentytwentythree/templatesindex.html

rather than 

* twentytwentythree/parts/post-meta.html
* twentytwentythree/parts/index.html

I'm not quite sure what the odd character is but it appears as a bold dot in Windows file explorer. 

This only happens on Windows where DIRECTORY_SEPARATOR has been set to the backslash character

We've managed to create a workaround by changing the code on lines 175 and 198 in theme-zip.php from 

~~~
$template_folders['wp_template'] . DIRECTORY_SEPARATOR . $template->slug . '.html',
~~~  

and

~~~
$template_folders['wp_template_part'] . DIRECTORY_SEPARATOR . $template->slug . '.html',
~~~

to 

~~~
$template_folders['wp_template'] . '/' . $template->slug . '.html',
~~~  

and

~~~
$template_folders['wp_template_part'] . '/' . $template->slug . '.html',
~~~

The problem does not occur on a Linux host.

t-hamano:
Fixes #602

This PR fixes an issue where files are not created correctly when using certain file archivers on Windows OS.

Where the problem occurs, the file is generated by `ZipArchive::addFromString`. [The PHP documentation](https://www.php.net/manual/en/ziparchive.addfromstring.php) mentions the following:

> Note: For maximum portability, it is recommended to always use forward slashes (/) as directory separator in ZIP filenames.

I don't know the exact cause, but I think backslashes are causing problems with certain file archivers.

## Testing Instructions

- Check out this branch on Windows.
- Download & install [7-Zip.](https://www.7-zip.org/)
- Run "Export Zip" in the Site Editor.
- Unzip the file using 7-zip.

## Screenshot

| Before | After |
|--------|--------|
| image | image |


admin/create-theme/theme-zip.php
```diff
@@ -172,7 +172,7 @@ public static function add_templates_to_zip( $zip, $export_type, $new_slug ) {
 
 			// Write the template content
 			$zip->addFromStringToTheme(
-				$template_folders['wp_template'] . DIRECTORY_SEPARATOR . $template->slug . '.html',
+				$template_folders['wp_template'] . '/' . $template->slug . '.html',
```
vcanales: We could also use [`path_join`](https://developer.wordpress.org/reference/functions/path_join/). I find that it's easier to figure out what's going at a glance, and also provides protection against extra slashes that might show up.

```suggestion
				path_join( $template_folders['wp_template'], $template->slug . '.html' ),
```

admin/create-theme/theme-zip.php
```diff
@@ -195,7 +195,7 @@ public static function add_templates_to_zip( $zip, $export_type, $new_slug ) {
 
 			// Write the template content
 			$zip->addFromStringToTheme(
-				$template_folders['wp_template_part'] . DIRECTORY_SEPARATOR . $template->slug . '.html',
+				$template_folders['wp_template_part'] . '/' . $template->slug . '.html',
```
vcanales:
```suggestion
				path_join( $template_folders['wp_template_part'], $template->slug . '.html' ),
```

nicholasingham:
Just a thought. The plugin code doesn't use the path_join function at the moment. 

Syntax such as `$a. '/' . $b` or `'pattern/' . $c` is used throughout. 

For consistency, would it be better to continue to use the `. '/' .` syntax for this PR and decide whether to change to path_join everywhere later?

vcanales:
> Just a thought. The plugin code doesn't use the path_join function at the moment.
> 
> Syntax such as `$a. '/' . $b` or `'pattern/' . $c` is used throughout.
> 
> For consistency, would it be better to continue to use the `. '/' .` syntax for this PR and decide whether to change to path_join everywhere later?

We're introducing it as a part of another PR as well https://github.com/WordPress/create-block-theme/pull/601. 

There are also places where it might not be suitable to use it ie. where `DIRECTORY_SEPARATOR` is in use.

I don't see a drawback in introducing it here, and also gradually whenever we need to make changes to path handling logic.

nicholasingham:
Makes sense. Many thanks.

pbking (approved):
Took longer to get my windows environment working again then it did to test it (and for just a 2 line change at that 🤣 )

Works as expected.  The unpacked zip archive looks as expected.

Thanks for this fix.

PR #607, opened 2024-04-29
Editor Sidebar: Persist "Save Changes" panel settings

### LINKED ISSUE #597: Improve 'Save Changes' panel with persisted options or a "select all" option.
The new "Save Changes" options give the end user much more control over how a theme is saved and finally exported. To help serve most users in the best possible way, some items are not selected by default:

image

On the other hand, some users will want these options on every save. Should we consider adding a "select all" checkbox for these options to save some clicks here?

t-hamano:
Closes #597

This PR persists the checked state of the "Save Changes" panel.

To make settings persistent, use the [@wordpress/preferences](https://github.com/WordPress/gutenberg/blob/trunk/packages/preferences/README.md) package. This package is also used in Gutenberg to persist various settings, which are stored in a database. This ensures that the same settings remain the same for the same user even if the user reloads the browser or uses a different browser.

database-info

### Testing Instructions

- By default, the following four items should be checked.
  - Save Fonts
  - Save Style Changes
  - Save Template Changes
  - Process Only Modified Templates
- When the "Save template Changes" is unchecked, the four checkboxes below it will be disabled as before.
- Reload your browser and confirm that the settings persist.
- Confirm that when you click "Save Changes" the theme will update according to your settings as before.

### About Initializing Settings

To restart the test from the beginning, perform the following two operations.

Chrome Dev Tools > Apprication > Storage > Local storage > Right click "http://localhost:8888" > Clear

local-storage

Connect to database > `wp_usermeta` table > Delete `wp_persisted_preferences` record

database


pbking (approved):
Nicely done.  Works as expected.  Simple change.

It's a nice cleanup for the effected view code too.

A great addition!

bacoords:
Yes! Thanks @t-hamano (I came here to open an issue for this functionality and of course you already did it.)

PR #611, opened 2024-04-30
Proposal: Use non-default port for wp-env

vcanales:
I find myself having to stop my Gutenberg environment, which uses the default port, when working on both at around the same time. This change should make it easier to have both environments running in parallel.

I realize I could choose to run a .wp-env.override.json file as well, but I think this might benefit other folks as well.

### DIFF
--- a/.wp-env.json
+++ b/.wp-env.json
@@ -1,5 +1,7 @@
 {
 	"core": "WordPress/WordPress",
+	"port": 8988,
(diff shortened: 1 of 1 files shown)


matiasbenedetto (approved):
The containers run at the specified ports. It looks good to me.

pbking (approved):

PR #618, opened 2024-05-06
Tests: use tests-wordpress wp-env for phpunit

vcanales:
Running tests on the `cli` environment causes conflicts with the `dev` environment; the tests will log you out, modify the theme, etc.

Running tests on `tests-wordpress` will allow you to run tests along with the `dev` environment without interference.

## Testing:
1. Run `npm run wp-env start` to start a WordPress environment.
2. Log into the dev environment and see that everything is working normally.
3. Running `npm run test:php` will work as before, and will not mess with the dev environment.
Before applying this branch, running `npm run test:php` would log you out and activate an invalid theme on your dev environment.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
 		"lint:php:fix": "composer run-script format",
 		"lint:md-docs": "wp-scripts lint-md-docs",
 		"lint:pkg-json": "wp-scripts lint-pkg-json",
-		"test:php": "npm run test:php:setup && wp-env run cli --env-cwd='wp-content/plugins/create-block-theme' composer run-script test",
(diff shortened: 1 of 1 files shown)


matiasbenedetto (approved):
I could reproduce this on trunk:
> running npm run test:php would log you out and activate an invalid theme on your dev environment.

This branch doesn't affect the main container but breaks WordPress in the tests container, as it used to in the main dev container. I think that's unexpected and probably caused by problematic test cases we should fix.

Still, this change mitigates the problem so, I'm approving the PR, but doesn't solve it.
I think the follow-up of this PR is fixing the tests that break the WordPress instance in the tests container.

PR #619, opened 2024-05-06
Update wp-env package version

matiasbenedetto:
## What?
Update wp-env package version.

## Why?
To use th latest relase of the package.

I've been getting errors running unit tests for Create Block Theme on Linux. Using the latest version seems to fix the problem.


vcanales (approved):
LGTM.

Note that after installing the new version, I got database-related errors that went away by running `npm run wp-env destroy`. I think this should be fine; the environment is expected to be ephemeral anyway.

PR #620, opened 2024-05-07
Update package-lock.json

t-hamano:
I noticed that when I run `npm ci` on the latest trunk, `package-lock.json` gets updated. I think this is because only `@wordpress/env` in `package.json` was updated in #619.

### Testing Instructions

- Check out this branch and run `npm i`.
- Run `git diff`
- There should be no difference.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -79,114 +79,43 @@
 			}
 		},
 		"node_modules/@babel/code-frame": {
-			"version": "7.22.13",
-			"resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.22.13.tgz",
-			"integrity": "sha512-XktuhWlJ5g+3TJXc5upd9Ks1HutSArik6jf2eAjYFyIOf4ej3RN+184cZbzDvbPnuTJIUhPKKJE3cIsYTiAT3w==",
+			"version": "7.24.2",
+			"resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.24.2.tgz",
+			"integrity": "sha512-y5+tLQyV8pg3fsiln67BVLD1P13Eg4lh5RW9mF0zUuvLrv9uIQ4MCL+CRT+FTsBlBjcIan6PGsLcBN0m3ClUyQ==",
 			"dev": true,
 			"dependencies": {
-				"@babel/highlight": "^7.22.13",
-				"chalk": "^2.4.2"
+				"@babel/highlight": "^7.24.2",
+				"picocolors": "^1.0.0"
 			},
 			"engines": {
 				"node": ">=6.9.0"
 			}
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -11,7 +11,6 @@
 			"dependencies": {
 				"@codemirror/lang-json": "^6.0.1",
 				"@uiw/react-codemirror": "^4.21.24",
-				"@wordpress/env": "^9.9.0",
```
t-hamano, line 14: This package is installed as `devDependencies` in `package.json`. Therefore, it is unclear why it existed as `dependencies` in `package-lock.json`.

mikachan (commented):
I'm still seeing some differences in my package-lock.json after running `npm i`:

- It looks like `@playwright/test`, `playwright`, `playwright/node_modules/fsevents`, and `typescript` have been removed, but these changes haven't been committed to the package-lock. 

e.g.

image

t-hamano:
Thanks for the testing.

I tried regenerating `package-lock.json`, but does the difference still occur? By the way, if you run `npm i` on trunk, does a diff occur?

mikachan:
> I tried regenerating package-lock.json, but does the difference still occur? 

Yes, although not as many differences - the difference with `playwright/node_modules/fsevents` is not happening anymore, but the others still are (playwright and typescript). I can push my package-lock so you can see the differences if you like?

> By the way, if you run npm i on trunk, does a diff occur?

Yes, I see a lot of changes in my package-lock on trunk.

t-hamano:
>I can push my package-lock so you can see the differences if you like?

Yes, I would appreciate it if you could push it. Your Node version is 20, right?

mikachan:
I've just pushed it here: https://github.com/WordPress/create-block-theme/pull/620/commits/3a65a352ead8d81b42c8d241140ce167c78b723d

> Your Node version is 20, right?

Yes, I'm using v20.12.2.

t-hamano:
Check out the latest commits on this branch and I tried running `npm i` again, but the diff still occurs 😅 My Node version is also 20.12.0. Could you tell me the npm version? My npm version is 10.5.0.

mikachan:
My npm version is v10.5.0 too! I'm happy to go with your package-lock version, I don't think it's a big deal, but it's confusing why they're different 😅

t-hamano:
I don't understand why there is a difference since the npm version and node version are exactly the same 😅

I tried regenerating `package-lock.json` again. Could you running the following on this branch?

```
rm -rf node_modules/
npm cache clear --force
npm install
```

If it doesn't resolve the issue, it might be better to close this PR...

pbking (approved):
This tests fine for me.
Installed dependencies from scratch and the lock file remains unchanged.

I'm keen to bring this in.

PR #621, opened 2024-05-08
Metadata screenshot

### LINKED ISSUE #560: Add ability to change screenshot of theme from editor tools
The screenshot can be provided from the wp-admin interface but not from the editor interface.  This should be added.

pbking:
Add ability for user to manage the theme's screenshot from Editor on Metadata Modal.

Fixes: #560

image

To Test:

* From the Editor open the CBT Metadata Editor Modal
* Click "Replace" and choose another image to replace the screenshot with
* Click "Update"

The image should be replaced in the theme folder and sized to 1200 x 900 px.

If a screenshot existed prior it should be removed. (Especially if the previous screenshot was another file type)

### DIFF
--- a/admin/create-theme/theme-utils.php
+++ b/admin/create-theme/theme-utils.php
@@ -205,27 +205,34 @@ class Theme_Utils {
 		return null;
 	}
 
-	public static function process_screenshot( $file_path ) {
-		$new_screeenshot_id      = attachment_url_to_postid( $file_path );
+	public static function copy_screenshot( $file_path ) {
+
+		$new_screeenshot_id = attachment_url_to_postid( $file_path );
+
+		if ( ! $new_screeenshot_id ) {
+			return new \WP_Error( 'screenshot_not_found', __( 'Screenshot not found', 'create-block-theme' ) );
+		}
+
 		$new_screenshot_metadata = wp_get_attachment_metadata( $new_screeenshot_id );
 		$upload_dir              = wp_get_upload_dir();
 
 		$new_screenshot_location = path_join( $upload_dir['basedir'], $new_screenshot_metadata['file'] );
 
-		// TODO: We're not actually processing anything.  The image might be of the wrong dimensions, too large, etc.
-		// We're just getting the file location so that we can copy it.
-		// Should we process it?  Or just enforce that the screenshot is the correct size?
+		$new_screenshot_filetype = Theme_Utils::get_screenshot_file_extension( $file_path );
+		$new_location            = path_join( get_stylesheet_directory(), 'screenshot.' . $new_screenshot_filetype );
 
-		return $new_screenshot_location;
+		// copy and resize the image
+		$image_editor = wp_get_image_editor( $new_screenshot_location );
+		$image_editor->resize( 1200, 900, true );
+		$image_editor->save( $new_location );
+
+		return true;
 	}
 
 	public static function replace_screenshot( $new_screenshot_path ) {
 		if ( ! Theme_Utils::is_valid_screenshot_file( $new_screenshot_path ) ) {
 			return new \WP_Error( 'invalid_screenshot', __( 'Invalid screenshot file', 'create-block-theme' ) );
 		}
-
-		// Clean up, resize or compress the screenshot if necessary
-		$processed_screenshot_location = Theme_Utils::process_screenshot( $new_screenshot_path );
 
 		// Remove the old screenshot
 		$old_screenshot = wp_get_theme()->get_screenshot( 'relative' );
@@ -234,9 +241,7 @@ class Theme_Utils {
 		}
 
 		// Copy the new screenshot
-		$new_screenshot_filetype = Theme_Utils::get_screenshot_file_extension( $new_screenshot_path );
-		$new_location            = path_join( get_stylesheet_directory(), 'screenshot.' . $new_screenshot_filetype );
-		copy( $processed_screenshot_location, $new_location );
+		return Theme_Utils::copy_screenshot( $new_screenshot_path );
 	}
 
 }
--- a/admin/create-theme/theme-utils.php
+++ b/admin/create-theme/theme-utils.php
@@ -228,7 +228,10 @@ class Theme_Utils {
 		$processed_screenshot_location = Theme_Utils::process_screenshot( $new_screenshot_path );
 
 		// Remove the old screenshot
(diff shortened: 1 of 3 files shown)


admin/create-theme/theme-utils.php
```diff
@@ -185,4 +185,55 @@ public static function is_valid_screenshot( $file ) {
...
+	public static function process_screenshot( $file_path ) {
+		$new_screeenshot_id      = attachment_url_to_postid( $file_path );
+		$new_screenshot_metadata = wp_get_attachment_metadata( $new_screeenshot_id );
+		$upload_dir              = wp_get_upload_dir();
+
+		$new_screenshot_location = path_join( $upload_dir['basedir'], $new_screenshot_metadata['file'] );
+
+		// TODO: We're not actually processing anything.  The image might be of the wrong dimensions, too large, etc.
+		// We're just getting the file location so that we can copy it.
+		// Should we process it?  Or just enforce that the screenshot is the correct size?
```
pbking: Wondering if we should actually do any processing to the media selected (resize, compress, etc) or just use it as we get it and enforce sizes etc?

pbking:
I removed the `Remove` button as I don't think that the screenshot should ever not be there.

t-hamano:
@pbking Thank you for your quick implementation! I would like to test the backend a little more, but first I made some adjustments to the UI. What do you think? Are the buttons a little too noticeable?

| Before | After |
|--------|--------|
| no-screenshot-before | no-screenshot-after |
| has-screenshot-before | has-screenshot-after |

admin/create-theme/theme-utils.php
```diff
@@ -185,4 +185,55 @@ public static function is_valid_screenshot( $file ) {
...
+	public static function replace_screenshot( $new_screenshot_path ) {
+		if ( ! Theme_Utils::is_valid_screenshot_file( $new_screenshot_path ) ) {
+			return new \WP_Error( 'invalid_screenshot', __( 'Invalid screenshot file', 'create-block-theme' ) );
+		}
+
+		// Clean up, resize or compress the screenshot if necessary
+		$processed_screenshot_location = Theme_Utils::process_screenshot( $new_screenshot_path );
+
+		// Remove the old screenshot
+		unlink( path_join( get_stylesheet_directory(), wp_get_theme()->get_screenshot( 'relative' ) ) );
```
t-hamano: If there is no screenshot to begin with, a warning error will occur, so we may want to check if there is any old screenshot.

```
PHP Warning:  unlink(/var/www/html/wp-content/themes/twentytwentyfour/): Is a directory in /var/www/html/wp-content/plugins/create-block-theme/admin/create-theme/theme-utils.php on line 232
```

admin/create-theme/theme-utils.php
```diff
@@ -185,4 +185,55 @@ public static function is_valid_screenshot( $file ) {
...
+
+		// Copy the new screenshot
+		$new_screenshot_filetype = Theme_Utils::get_screenshot_file_extension( $new_screenshot_path );
+		$new_location            = path_join( get_stylesheet_directory(), 'screenshot.' . $new_screenshot_filetype );
+		copy( $processed_screenshot_location, $new_location );
```
t-hamano: [The theme handbook](https://developer.wordpress.org/themes/releasing-your-theme/required-theme-files/) states:

> In the WordPress.org theme directory, the screenshot acts as a visual indicator of what your theme looks like. It is visible both in the web view and in the admin dashboard. **The screenshot must not be bigger than 1200 x 900px**. 

Therefore, how about resizing the file to 1200 x 900px or less? It may be better to resize _after_ copying so as not to affect the original media.

I haven't tested it, but an approach like the one below might work?


```php
copy( $processed_screenshot_location, $new_location );

$image_editor = wp_get_image_editor( $new_location );
// Resize and crop image
$image_editor->resize( 1200, 900, true );
// Overwrite new screenshot
$image_editor->save();
```

admin/create-theme/theme-utils.php
```diff
@@ -185,4 +185,55 @@ public static function is_valid_screenshot( $file ) {
...
+		);
+		$filetype                 = wp_check_filetype( $file_path, $allowed_screenshot_types );
+		if ( in_array( $filetype['type'], $allowed_screenshot_types, true ) ) {
+			return $filetype['ext'];
+		}
+		return null;
+	}
+
+	public static function process_screenshot( $file_path ) {
+		$new_screeenshot_id      = attachment_url_to_postid( $file_path );
```
t-hamano: `attachment_url_to_postid()` function [returns zero when no post ID was found](https://github.com/WordPress/wordpress-develop/blob/08b4dd3b8b96ca388b939bb049f3177b90c2d883/src/wp-includes/media.php#L5316). Just in case, how about adding some error checking here?

admin/create-theme/theme-utils.php
pbking: Checked.

pbking:
> What do you think? Are the buttons a little too noticeable?

I think they are fine either way.  I like the option you just pushed slightly better though.

👍

admin/create-theme/theme-utils.php
pbking: I did something very similar; thanks for getting me started. 
The editor copied the image and I pointed it at a new location on the save action so it was all one action.

admin/create-theme/theme-utils.php
pbking: Defensive logic added.

t-hamano (approved):
LGTM!

- PHP error logs will not occur when adding a screenshot to a theme that does not originally have it.
- The screenshot is resized appropriately
- Also works on Windows host OS

madhusudhand:
It works great. I have observed one minor issue with image size. 
Tried uploading following image

image-alignment-1200x4002-1

It previews screenshot differently.

image

image

We should probably consider both ratio `4:3` and max size `1200x900` while resizing ?

pbking:
Hrm 🤨

Regarding the wide image...  The size restrictions are noted as a MAXIMUM of 1200 x 900.  Which the image fits in.  The original is less than 900 pixels tall.  I believe that the cropping/resizing is acting as I would expect.

I tried another image that was very wide, but over 900 px tall (2500 x 1200) and it cropped and resized to the expected 1200 x 900 size.

There might be a better way to handle this edge case (upscaling the image maybe?  then cropping it?) but I don't think it's worth the effort (when a user should really be providing an image of the appropriate dimensions anyway...)

If you believe that we should consider addressing that further would you mind opening an issue to explore additional solutions @madhusudhand ?

PR #622, opened 2024-05-09
Allow spaces in theme slugs

### LINKED ISSUE #228: Update `get_theme_slug` to allow theme names with hyphens/spaces
Any theme name that includes a space needs to either have the space removed or replaced with a hyphen for use in the directory name and the text domain.

Currently, the `get_theme_slug` function is removing hyphens from a new theme name based on the following logic:

When the source theme has a single-word slug but the new theme has a multi-word slug (e.g. cloning twentytwentythree and then naming the new theme tt3-clone would change the new theme name to 'tt3clone').

This can be confusing, as the .zip file has a different name to the chosen theme name, and it results in some discrepancies throughout the theme code where the slug is referenced, such as in pattern blocks and the Text Domain.

I believe this was to prevent hyphens from being used in theme support functions (e.g. `tt3-clone_support()`), but perhaps this logic could be more specific to the function name creation, rather than the theme slug itself. Could we replace the hyphens with underscores just for function names?

This issue came up while testing https://github.com/WordPress/create-block-theme/pull/213.

pbking:
This change allows spaces in theme slugs.

(I thought this would be a much harder problem, but it was actually pretty easy.)

I realized that any time a slug with _'s needed to be replaced it was ALSO followed by a _  ( i.e. slug_name_function and never just slug_name).  So replacing `old_slug_` with `new_slug_` was the answer.

The order of replacement (replacing slugs with _'s first) had to be changed to make sure that when cloning a theme without _'s in the slug that the functions were correctly changed.

Fixes #228 

To Test: 
* Using the CBT interface in the Editor
* Create a Blank theme with a name that contains spaces, dashes and/or underscores
* Clone a theme with a name spaces, dashes and/or underscores

Observe that the new theme has the correct textdomain and it is used correctly throughout the theme (especially in PHP files with functions that use the theme slug such as `functions.php`).

### DIFF
--- a/admin/create-theme/theme-utils.php
+++ b/admin/create-theme/theme-utils.php
@@ -6,23 +6,7 @@ class Theme_Utils {
 	}
 
 	public static function get_theme_slug( $new_theme_name ) {
-		$theme = wp_get_theme();
-
-		// If the source theme has a single-word slug but the new theme has a multi-word slug
-		// then function will look like: function apple-bumpkin_support() and that won't work.
-		// There are no issues if it is multi-word>single-word or multi>multi or single>single.
-		// Due to the complexity of this situation (compared to the simplicity of the others)
-		// this will enforce the usage of a singleword slug for those themes.
-
-		$old_slug = $theme->get( 'TextDomain' );
-		$new_slug = sanitize_title( $new_theme_name );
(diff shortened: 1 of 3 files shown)


madhusudhand (approved):
It works great! 🚢

Issue #523, opened 2024-03-19
Make sure patch version gets removed when theme is exported


iamtakashi:
> RECOMMENDED: Tested up to is recommended to have major versions only (e.g. 5.8). Patch version is not needed (e.g. 5.8.1).

When I run the theme check, this often comes up. Let's make sure it gets removed when the theme is exported. It seems to be low-hanging fruit that will save us time!

pbking:
I don't believe that we should be automatically removing the patch version.  It's recommended, though not enforced and the Automattic theme workflow still depends on patch versioning.

The Metadata editor allows for the version to be changed.  It's not automated, but you can remove the patch version in that way.

I do believe that the patch version can be excluded when creating/cloning a theme though.

mikachan:
I believe this issue is in reference to the "Tested up to" WordPress version number, not the theme version number.

pbking:
Ahhh

Silly. Of course. 

In that case, yrs of course!

PR #623, opened 2024-05-09
Reset changelog and version on clone and theme creation

### LINKED ISSUE #570: Clear Changelog when cloning a theme
When cloning a theme the changelog from the old theme should be cleared and started with a fresh start.

### LINKED ISSUE #613: Reset the theme version number to 1.0.0 when cloning a theme
Mentioned by @iamtakashi:

> shouldn’t the version number start with 1.0.0 instead of the same as the base theme?

image

pbking:
style.css and readme.txt should be bare (having only the values defined when creating the theme), changelog should be removed (with only the first "initial release" included) for cloned themes.

To test: clone a theme, create a child theme and create a blank theme

Observe the style.css and readme.txt files and ensure that the version is as expected.

NOTE: For CLONED themes the TAGS value in the style.css is copied over.  This is expected.

Fixes #613 
Fixes #570

### DIFF
--- a/admin/create-theme/theme-create.php
+++ b/admin/create-theme/theme-create.php
@@ -8,6 +8,9 @@ class Theme_Create {
 
 	public static function clone_current_theme( $theme ) {
 
+		$theme['version']     = '1.0';
+		$theme['tags_custom'] = implode( ', ', wp_get_theme()->get( 'Tags' ) );
+
 		// Create theme directory.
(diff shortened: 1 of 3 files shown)


madhusudhand (approved):
LGTM!

pbking:
I misunderstood the point of #523 so returned starting version to 1.0.0 (semantic versioning).

Issue #525, opened 2024-03-20
Font licenses are no longer automated with new Core Font Library functionality


mikachan:
Part of replacing the integrated font management with the new Font LIbrary in Core means that we're losing the automated font license functionality provided by this plugin.

This is currently handled with the `manage_font_license()` function, included [here](https://github.com/WordPress/create-block-theme/blob/trunk/admin/class-manage-fonts.php#L335).

Perhaps this could be added as a follow-up to https://github.com/WordPress/create-block-theme/pull/518 and https://github.com/WordPress/create-block-theme/pull/519.

matiasbenedetto:
I'm working on a fix for this. I will submit the PR once this work gets merged: https://github.com/WordPress/create-block-theme/pull/626

pbking:
Fixed with #649
Closing

PR #626, opened 2024-05-10
Refactor Theme_Readme (readme.txt) PHP class

### LINKED ISSUE #624: Child and cloned themes doesn't include a reference to the original theme in the readme.txt
## What ?
Themes created as child themes or with the 'clone' functionality don't include the string referencing the original theme in the readme.txt.

## Expected
The CBT plugin used to include a reference in the readme.txt file for those types of themes.
The related code is still there, but the functionality no longer works.

https://github.com/WordPress/create-block-theme/blob/05242557344499c9ee47e258d0012628b308b3a3/admin/create-theme/theme-readme.php#L57C1-L117C3

### LINKED ISSUE #631: Load data from readme.txt in the 'Edit metadata' modal
## What?
Currently, the 'Edit metadata' modal loads only some existing theme data items (such as 'Theme description', 'Author', 'Version', etc.) when you open it. Still, other existing data points coming from readme.txt (such as 'Recommended plugins' and 'Image Credtis') are not loaded, so they can not be edited but only overwritten with new content. Apart from that, it is not clear if you are editing or simply adding new data to the existing one.

Screenshot 2024-05-13 at 10-15-59 Blog Home ‹ Template ‹ WP1 ‹ Editor — WordPress

## Expected
Load data from readme.txt in the 'Edit metadata' modal so the theme creators can:
- Edit the existing data for 'Recommended plugins' and 'Image Credits'.
- Know the content of the readme.txt file without the need to inspect it in the filesystem as a text file.

matiasbenedetto:
## What?
- Refactor ThemeReadme (readme.txt) class to make it more modular, re-usable, and testable.
- Add tests for their public methods in WordPress core codebase fashion.

## Why?
- To simplify the addition or fix of plugin functionality related to the readme file.
- Fixes: https://github.com/WordPress/create-block-theme/issues/624
- Enable the fix of [this issue](https://github.com/WordPress/create-block-theme/issues/525) in an easier way.
- Partially fixes: https://github.com/WordPress/create-block-theme/issues/631


admin/create-theme/theme-readme.php
```diff
@@ -137,123 +152,154 @@ static function copyright_section( $new_copyright_section, $original_theme_credi
...
+		$wp_version          = $theme['wp_version'] ?? get_bloginfo( 'version' );
+		$image_credits       = $theme['image_credits'] ?? '';
+		$recommended_plugins = $theme['recommended_plugins'] ?? '';
 
-			// Get current theme readme.txt
-			$current_readme         = get_stylesheet_directory() . '/readme.txt' ?? '';
-			$current_readme_content = file_exists( $current_readme ) ? file_get_contents( $current_readme ) : '';
+		// Get current readme content.
+		$readme_content = self::get_content();
+		$updated_readme = $readme_content;
```
pbking, line 185: Can we just simplify this to 
```
$updated_readme = self::get_content();
if ( ! $updated_readme ) {
  return;
}
```

admin/create-theme/theme-readme.php
```diff
@@ -1,50 +1,103 @@
...
+		$recommended_plugins  = $theme['recommended_plugins'] ?? '';
+		$is_child_theme       = $theme['is_child_theme'] ?? false;
+
+		// Generates a reference to the original theme for child and cloned themes.
+		$original_theme_credits = $is_child_theme
+			? self::original_theme_credits( $name, $is_child_theme )
+			: null;
+		// Generates the copyright section text.
+		$copyright_section_content = self::copyright_text( $name, $copy_year, $author, $original_theme_credits );
+
```
pbking: I find this part very confusing.  How could we refactor this to be more clear what is happening and why?

pbking: Perhaps we could make this 
```
$copyright_section_content = self::generate_copyright_text( $theme );
```

And within that function determine what copyright is most appropriate.

pbking: Also the comment notes that it would generate a reference for child AND cloned themes, however the "cloned" part of this seems to have gone away.

admin/create-theme/theme-readme.php
```diff
@@ -1,50 +1,103 @@
...
+		if ( ! file_exists( $path ) ) {
+			return '';
+		}
+		return file_get_contents( $path );
+	}
 
-		return "=== {$name} ===
+	/**
+	* Build a readme.txt file for CHILD/GRANDCHILD themes.
+	*
```
pbking: This also builds one for BLANK themes, so probably drop the "for CHILD/GRANDCHILD themes" part

admin/create-theme/theme-readme.php
pbking: Can we pull the get_content() call out of here and just provide the string to be updated instead?

That would feel a lot cleaner, cleaner to test, pure function without dependencies.

admin/create-theme/theme-readme.php
```diff
@@ -137,123 +152,154 @@ static function copyright_section( $new_copyright_section, $original_theme_credi
...
+	/**
+	 * Write the readme.txt file.
+	 *
+	 * @param string $content The content to write.
+	 * @return bool True if the content was written, false otherwise.
+	 */
+	public static function write( $content ) {
+		$path = self::file_path();
+		return (bool) file_put_contents( $path, $content );
+	}
```
pbking: This function is used only in a single place (beyond the tests) and that could just be replace with a call to file_put_contents().  I don't believe this adds value, just complication.

admin/create-theme/theme-readme.php
matiasbenedetto: Thanks for flagging that. I simplified it in this commit: https://github.com/WordPress/create-block-theme/pull/626/commits/904f45e06bd46f7c424a0bc7bb1370cb24781d1a

matiasbenedetto: I fixed the cloned themes reference and added a test case for it here: https://github.com/WordPress/create-block-theme/pull/626/commits/b4cb121893f986f185de99044de2bf3308f55a03

admin/create-theme/theme-readme.php
matiasbenedetto: done here: https://github.com/WordPress/create-block-theme/pull/626/commits/ad9f54c832558fe5ac1dff9ffec54c6e8c810147

admin/create-theme/theme-readme.php
matiasbenedetto: Removed here: https://github.com/WordPress/create-block-theme/pull/626/commits/2d4cedbab2ab13e69a1d475b7879ebdce5b4e374

admin/create-theme/theme-readme.php
matiasbenedetto: removed here: https://github.com/WordPress/create-block-theme/pull/626/commits/3e1bd608bcd39aa248454d293896bafdbc4ed680

admin/create-theme/theme-readme.php
pbking: This is much easier to understand!

admin/create-theme/theme-readme.php
pbking: love it

admin/create-theme/theme-readme.php
pbking: 🧼

pbking:
I found that the "Recommend Plugins" wasn't being sent back to the client (though it was being persisted if edited from the client). 

I did a lil' refactoring, pulling the logic to get that and send it to the client from the "utils" class to the "readme" class and had it pull all the sections, not just look exclusively for the "recommended plugins" section.

pbking (approved):
Everything is working great!  Code looks great, huge improvement.

If you're OK with my additions then I think this is ready to come in!

matiasbenedetto:
> I found that the "Recommend Plugins" wasn't being sent back to the client (though it was being persisted if edited from the client).

Haha yep, I created an issue about the same https://github.com/WordPress/create-block-theme/issues/631

> I did a lil' refactoring, pulling the logic to get that and send it to the client from the "utils" class to the "readme" class and had it pull all the sections, not just look exclusively for the "recommended plugins" section.

Cool, good addition.

PR #629, opened 2024-05-10
Fix/metadata saving

### LINKED ISSUE #523: Make sure patch version gets removed when theme is exported

pbking:
Fixes the following issues:

* Retain copyright text in style.css when updating metadata
* Uses only major/minor version of wordpress for "tested up to" value
* Use existing license in style.css instead of overwriting. 
* Use existing Requires WP in style.css instead of overwriting.

Fixes #523 

To test:

Click "Update" on the Metadata modal of CBT.

Note that the "Tested Up to" value is only major.minor
Note that the license values remain unchanged (set it to something noticeable before saving to ensure this is true)
Note that the Requires WP in style css remains unchanged (set it to something noticeable before saving to ensure this is true)
Note that the license text remains after saving.

### DIFF
--- a/admin/create-theme/theme-readme.php
+++ b/admin/create-theme/theme-readme.php
@@ -12,7 +12,7 @@ class Theme_Readme {
 		$author              = $theme['author'];
 		$author_uri          = $theme['author_uri'];
 		$copy_year           = gmdate( 'Y' );
-		$wp_version          = get_bloginfo( 'version' );
+		$wp_version          = Theme_Utils::get_current_wordpress_version();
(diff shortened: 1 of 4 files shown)


pbking:
Closing in deference to #634 and #635

Issue #495, opened 2023-12-21
Theme exporting: improve tags interface


creativecoder:
When exporting a theme, the theme tags section inherits whatever tags are being used in the base theme.

<img width="1057" alt="image" src="https://github.com/WordPress/create-block-theme/assets/1699996/963d617a-205c-4cee-82f4-b838605f5da9">

These could use some improvements, for starters

- A "Deselect all" option, so you can easily remove the base theme tags
- A better description that tells these are important/useful

matiasbenedetto:
This issue references a deprecated UI, so I'm closing this issue as not planned.

PR #630, opened 2024-05-12
Improve tags UI

t-hamano:
Inspired by #495

This PR replaces the tag's input UI from a text field to a `FormTokenField` component.

https://github.com/WordPress/create-block-theme/assets/54422211/5fd13b2b-2162-490c-9e4d-3500f3af4467

I believe this UI will have the following benefits:

- Tags have a background color, improving visibility
- In addition to commas, you can use the enter key to complete tag entry.
- Can prevent registration of duplicate tags

Also, this PR is intended to change only the UI, so there are no internal changes. If this PR is considered the first step in improving the UI, I think the ideal UI logic is as follows:

- Show theme tag _label_ instead of theme tag _slug_
- Sort the existing tag list into theme tags and custom tags and separate input fields.
- Regarding theme tags, those that are equal to the array obtained from the `get_theme_feature_list()` function are displayed as candidates, and tags that are not included in that array are not allowed.

Below is a mockup suggestion.

image

## Testing Instructions

- In the trunk branch, register some tags.
- If you switch to this branch, your tags should remain fine.
- Verify that entering and saving tags work correctly in this branch.

### DIFF
--- a/src/editor-sidebar/metadata-editor-modal.js
+++ b/src/editor-sidebar/metadata-editor-modal.js
@@ -189,6 +189,16 @@ export const ThemeMetadataEditorModal = ( { onRequestClose } ) => {
 					}
 					onChange={ onChangeTags }
 				/>
+				<ExternalLink
+					href="https://make.wordpress.org/themes/handbook/review/required/theme-tags/"
+					style={ {
+						fontSize: '12px',
+						marginTop: '-20px',
+						marginBottom: '1rem',
+					} }
+				>
+					{ __( 'Read more.', 'create-block-theme' ) }
+				</ExternalLink>
 				<TextareaControl
 					label={ __( 'Recommended Plugins', 'create-block-theme' ) }
 					help={
--- a/src/editor-sidebar/metadata-editor-modal.js
+++ b/src/editor-sidebar/metadata-editor-modal.js
@@ -191,16 +191,19 @@ export const ThemeMetadataEditorModal = ( { onRequestClose } ) => {
 					}
 					onChange={ onChangeTags }
 				/>
-				<ExternalLink
-					href="https://make.wordpress.org/themes/handbook/review/required/theme-tags/"
+				<HStack
 					style={ {
-						fontSize: '12px',
 						marginTop: '-20px',
 						marginBottom: '1rem',
 					} }
 				>
-					{ __( 'Read more.', 'create-block-theme' ) }
-				</ExternalLink>
+					<ExternalLink
+						href="https://make.wordpress.org/themes/handbook/review/required/theme-tags/"
+						style={ { fontSize: '12px' } }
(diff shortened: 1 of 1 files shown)


pbking:
A great improvement!

Could we use the opportunity to point to more details about the tags?

https://make.wordpress.org/themes/handbook/review/required/theme-tags/

I tried something like the below (so that the style would match the Plugins bit below it) but I'm not sure that's the best way to do that. :? 

```
<ExternalLink href="https://make.wordpress.org/themes/handbook/review/required/theme-tags/" style={{fontSize:'12px', marginTop:'-20px'}}>
	{ __( 'Read more.', 'create-block-theme' ) }
</ExternalLink>
```

src/editor-sidebar/metadata-editor-modal.js
```diff
@@ -177,16 +182,12 @@ export const ThemeMetadataEditorModal = ( { onRequestClose } ) => {
...
-						setTheme( { ...theme, tags_custom: value } )
+					value={
+						theme.tags_custom ? theme.tags_custom.split( ', ' ) : []
 					}
-					placeholder={ __(
-						'A comma-separated collection of tags',
-						'create-block-theme'
-					) }
+					onChange={ onChangeTags }
 				/>
```
vcanales, line 193: I agree with @pbking in that we should include a link to the available tags documentation. The `FormTokenField` component unfortunately doesn't have a `help` prop like the other components used here, but we can add the link regardless:


```js
<>
	<FormTokenField
		label={ __( 'Theme tags', 'create-block-theme' ) }
		value={
			theme.tags_custom
				? theme.tags_custom.split( ', ' )
				: []
		}
		onChange={ onChangeTags }
	/>
	<ExternalLink
		href="https://make.wordpress.org/themes/handbook/review/required/theme-tags/"
		style={ {
			fontSize: '12px',
			marginTop: '-20px',
			marginBottom: '1rem',
		} }
	>
		{ __( 'Read more.', 'create-block-theme' ) }
	</ExternalLink>
</>
```

I've tried a bottom margin on the external link as well, in order to more or less match the spacing between the link and the next header in the other sections.

image

pbking: Thanks @vcanales.

I pulled in the changes you suggested.

pbking (approved):
This is a very good, simple change.  With the added link to tags info I think this is ready to bring in.

t-hamano:
I was also just working on updating this PR. I think we need to further align the focus outline to the text.

image

t-hamano:
How about this?

image

pbking:
👍 

Good change I like it.

🚢

vcanales (approved):
Looks good. @t-hamano feel free to merge when you're ready.

PR #634, opened 2024-05-14
Don't Clobber Metadata

pbking:
When saving metadata to style.css a number of default values are written without considering the existing values.

This change keeps those values, copying them from the old style.css file to what is generated.

Specifically:

Copyright information
License/License URI
WordPress minimum version

This also fixes a bug where the VERSION was being wiped when updating metadata.

To Test:
Activate a theme that has things that could be changed. (Ames is a good theme to test with, it has all of the above)
Using CBT 'Edit Theme Metadata' and 'Update' without changing anything.

Note how the style.css changed.  ONLY the "Tested up to" value should be different.

### DIFF
--- a/admin/create-theme/theme-styles.php
+++ b/admin/create-theme/theme-styles.php
@@ -21,11 +21,19 @@ class CBT_Theme_Styles {
 		$requires_php  = $current_theme->get( 'RequiresPHP' );
 		$template      = $current_theme->get( 'Template' );
 		$text_domain   = $theme['slug'];
+		$license       = 'GNU General Public License v2 or later';
+		$license_uri   = 'http://www.gnu.org/licenses/gpl-2.0.html';
+		$tags          = CBT_Theme_Tags::theme_tags_list( $theme );
 
-		//TODO: These items don't seem to be available via ->get('License') calls
-		$license      = 'GNU General Public License v2 or later';
-		$license_uri  = 'http://www.gnu.org/licenses/gpl-2.0.html';
-		$tags         = CBT_Theme_Tags::theme_tags_list( $theme );
+		preg_match( '/License: (.+)/', $style_css, $matches );
+		if ( isset( $matches[1] ) ) {
+			$license = $matches[1];
+		}
+		preg_match( '/License URI: (.+)/', $style_css, $matches );
+		if ( isset( $matches[1] ) ) {
+			$license_uri = $matches[1];
+		}
+
 		$css_metadata = "/*
 Theme Name: {$name}
 Theme URI: {$uri}
--- a/admin/create-theme/theme-styles.php
+++ b/admin/create-theme/theme-styles.php
@@ -28,19 +28,19 @@ class CBT_Theme_Styles {
 		$wp_min        = $current_theme->get( 'RequiresWP' );
 		$version       = $theme['version'];
 		$requires_php  = $current_theme->get( 'RequiresPHP' );
-		$template      = $current_theme->get( 'Template' );
 		$text_domain   = $theme['slug'];
+		$template      = $current_theme->get( 'Template' ) ? "\n" . 'Template: ' . $current_theme->get( 'Template' ) : '';
 		$license       = $style_data['License'] ? $style_data['License'] : 'GNU General Public License v2 or later';
 		$license_uri   = $style_data['LicenseURI'] ? $style_data['LicenseURI'] : 'http://www.gnu.org/licenses/gpl-2.0.html';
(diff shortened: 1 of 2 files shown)


admin/create-theme/theme-styles.php
```diff
@@ -46,7 +60,8 @@ public static function update_style_css( $style_css, $theme ) {
 
 		$css_metadata .= "Text Domain: {$text_domain}
 Tags: {$tags}
-*/
+
+{$copyright}*/
```
t-hamano: How about wrapping this in an if statement? If no more than default metadata is present, a blank line is inserted before the comment's closing tag.

Below is an example for TT4.

image

admin/create-theme/theme-styles.php
```diff
@@ -17,22 +17,36 @@ public static function update_style_css( $style_css, $theme ) {
...
+		$copyright     = '';
+
+		preg_match( '/License: (.+)/', $style_css, $matches );
+		if ( isset( $matches[1] ) ) {
+			$license = $matches[1];
+		}
+		preg_match( '/License URI: (.+)/', $style_css, $matches );
+		if ( isset( $matches[1] ) ) {
+			$license_uri = $matches[1];
+		}
```
t-hamano: Wouldn't it be more accurate and simpler to use `get_file_data()`? The `WP_Theme` class also seems to use this function [internally](https://github.com/WordPress/wordpress-develop/blob/13d5244c43d65da5fd3ff7abd390fa36bb4a87b5/src/wp-includes/class-wp-theme.php#L348).

```php
$theme_data = get_file_data(
	get_stylesheet_directory() . '/style.css',
	array(
		'License' => 'License',
		'License URI' => 'License URI',
	)
);
```

admin/create-theme/theme-styles.php
```diff
@@ -46,7 +60,8 @@ public static function update_style_css( $style_css, $theme ) {
 
 		$css_metadata .= "Text Domain: {$text_domain}
 Tags: {$tags}
-*/
+
+{$copyright}*/
```
t-hamano: I think we can delete this blank line (The line after `{$copyright}*/`). Otherwise, a double newline will be inserted at the end of the file.

double-line

t-hamano:
By the way, I submitted a core ticket proposing to be able to also retrieve the `License` and `License URI` fields via `wp_get_theme()`: https://core.trac.wordpress.org/ticket/61227#ticket

admin/create-theme/theme-styles.php
pbking: Nice.  I didn't even know about get_file_data, so yeah.  That does look simpler. :)

admin/create-theme/theme-styles.php
pbking: Refactored slightly to ensure the blank lines here are as expected.

admin/create-theme/theme-styles.php
pbking: Refactored slightly to ensure the blank lines here are as expected.

admin/create-theme/theme-styles.php
```diff
@@ -17,39 +25,37 @@ public static function update_style_css( $style_css, $theme ) {
...
 		$author_uri    = $theme['author_uri'];
 		$wp_version    = get_bloginfo( 'version' );
+		$wp_min        = $current_theme->get( 'RequiresWP' );
 		$version       = $theme['version'];
 		$requires_php  = $current_theme->get( 'RequiresPHP' );
-		$template      = $current_theme->get( 'Template' );
 		$text_domain   = $theme['slug'];
+		$template      = $current_theme->get( 'Template' ) ? "\n" . 'Template: ' . $current_theme->get( 'Template' ) : '';
+		$license       = $style_data['License'] ? $style_data['License'] : 'GNU General Public License v2 or later';
+		$license_uri   = $style_data['LicenseURI'] ? $style_data['LicenseURI'] : 'http://www.gnu.org/licenses/gpl-2.0.html';
```
t-hamano, line 34:
```suggestion
		$license       = $style_data['License'] ?? 'GNU General Public License v2 or later';
		$license_uri   = $style_data['LicenseURI'] ?? 'http://www.gnu.org/licenses/gpl-2.0.html';
```

Nit

t-hamano (approved):
LGTM! I feel that the refactoring has made it easier to imagine the output results.

admin/create-theme/theme-styles.php
pbking: That was my first swing but doesn't work because `$style_data['License']` would be `""` if unset, which is not null, so that value (an empty string) is used rather than the backup string provided.

PR #635, opened 2024-05-14
Use only major.minor version in 'Tested up to' field

### LINKED ISSUE #523: Make sure patch version gets removed when theme is exported

pbking:
Swap get_bloginfo('version') with custom function that supplies version with major.minor only.

Fixes #523 

To Test:
Using CBT update metadata.
Note that the "Tested Up to" shows the version of WordPress being used with only MAJOR.MINOR in style.css and readme.txt

### DIFF
--- a/admin/create-theme/theme-utils.php
+++ b/admin/create-theme/theme-utils.php
@@ -208,6 +208,12 @@ class CBT_Theme_Utils {
 		return CBT_Theme_Utils::copy_screenshot( $new_screenshot_path );
 	}
 
+	/**
+	 * Get the current WordPress version.
+	 *
+	 * @return string The current WordPress in the format x.x (major.minor)
+	 * Example: 6.5
+	 */
 	public static function get_current_wordpress_version() {
 		$wp_version = get_bloginfo( 'version' );
 		if ( preg_match( '/^\d+\.\d+/', $wp_version, $matches ) ) {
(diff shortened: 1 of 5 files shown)


matiasbenedetto (approved):
I added a docblock to the function added.
Apart from that it looks good to me.

admin/create-theme/theme-utils.php
```diff
@@ -208,4 +208,17 @@ public static function replace_screenshot( $new_screenshot_path ) {
...
 
+	/**
+	 * Get the current WordPress version.
+	 *
+	 * @return string The current WordPress in the format x.x (major.minor)
+	 * Example: 6.5
+	 */
+	public static function get_current_wordpress_version() {
+		$wp_version = get_bloginfo( 'version' );
+		if ( preg_match( '/^\d+\.\d+/', $wp_version, $matches ) ) {
```
vcanales: Why is this regex necessary? Is the version ever not in `major.minor` or `major.minor.patch` format? 

I think I'd favor using explode instead, to avoid directly using regex.

vcanales (commented):
I think this function is a great candidate for a unit test, rather than testing it by calling it from another class.

admin/create-theme/theme-utils.php
pbking: I'm impartial to how the version is snagged.  The regex fairy gave me that and it worked.  I wouldn't be sad if you refactored it.

pbking: 💣
exploded

PR #637, opened 2024-05-14
Replace/admin interface

pbking:
Replace the wp-admin interface with a React app that does the following:

* Export the theme zip
* Creates a new theme (Blank, clone child)
* Provides helpful information including details about the plugin, links to external resources and FAQs

image

image


beafialho:
I did a quick mockup of what this interface could look like, with a clearer content hierarchy and CTA's. The modal you shared above looks good to me when clicking each option.

CBT Interface

src/landing-page/landing-page.js
```diff
@@ -20,55 +20,55 @@ export default function LandingPage() {
...
 
-			<HStack alignment="left">
-				<Button variant="secondary" onClick={()=>setCreateModalType('blank')}>Create a new Blank Theme</Button>
+				<Button variant="link" style={{ fontSize:"22px"}} onClick={()=>setCreateModalType('blank')}>Create a new Blank Theme</Button>
 				<p>Start from scratch!  Create a blank theme to get started with your own design ideas.</p>
-			</HStack>
 
-			<HStack alignment="left">
-				<Button variant="secondary" onClick={()=>setCreateModalType('clone')}>Create a Clone of This Theme</Button><br/>
+				<Button variant="link" style={{ fontSize:"22px"}} onClick={()=>setCreateModalType('clone')}>Create a Clone of This Theme</Button><br/>
```
vcanales: Remove extra `br`, which separates the text from the title of the section.
```suggestion
				<Button variant="link" style={{ fontSize:"22px"}} onClick={()=>setCreateModalType('clone')}>Create a Clone of This Theme</Button>
```

pbking: Sorry, had to clobber your change.  Had quite a bit of styling work to push up.

vcanales: I don't know what clobber means 😂
But it's fine if it's gone :)

includes/class-create-block-theme-editor-tools.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+/**
+ * The Editor integration for the Create Block Theme plugin.
+ * @since      2.2.0
+ * @package    Create_Block_Theme
+ * @subpackage Create_Block_Theme/includes
+ * @author     WordPress.org
+ */
+class Create_Block_Theme_Editor_Tools {
```
vcanales: Let's keep the CBT class name prefix we've established.

includes/class-create-block-theme-admin-landing.php
```diff
@@ -0,0 +1,49 @@
+<?php
+
+/**
+ * The wp-admin landing page for the Create Block Theme plugin.
+ * @since      2.2.0
+ * @package    Create_Block_Theme
+ * @subpackage Create_Block_Theme/includes
+ * @author     WordPress.org
+ */
+class Create_Block_Theme_Admin_Landing {
```
vcanales: Let's keep the CBT class name prefix we've established.

vcanales (changes requested):
Took the liberty to:
1. Add missing translations
2. Change FAQ images to webp, which gives us a smaller filesize

src/landing-page/landing-page.js
```diff
@@ -0,0 +1,252 @@
...
+							{
+								a: (
+									// eslint-disable-next-line jsx-a11y/anchor-has-content
+									<a href="https://github.com/WordPress/create-block-theme" />
+								),
+							}
+						) }
+					</p>
+					<div className="cbt-lp-faq">
+						<h4>{ __( "FAQ's", 'create-block-theme' ) }</h4>
```
vcanales:
```suggestion
						<h4>{ __( "FAQ", 'create-block-theme' ) }</h4>
```

FAQ is already plural.

includes/class-create-block-theme-editor-tools.php
pbking: I actually wrote this part before we added the namespacing.  Thanks for pointing that out. :)

src/landing-page/create-modal.js
```diff
@@ -0,0 +1,169 @@
...
+				constructionFunction = createChildTheme;
+				break;
+		}
+
+		if ( ! constructionFunction ) {
+			return;
+		}
+		constructionFunction( theme )
+			.then( () => {
+				// eslint-disable-next-line
```
vcanales: Specify which rules are being disabled, so that it's easy to tell why at a glance.

```suggestion
				// eslint-disable-next-line no-undef, no-alert
```

src/landing-page/create-modal.js
```diff
@@ -0,0 +1,169 @@
...
+	__experimentalVStack as VStack,
+	// eslint-disable-next-line
+	__experimentalSpacer as Spacer,
+	// eslint-disable-next-line
+	__experimentalText as Text,
+	Modal,
+	Button,
+	TextControl,
+	TextareaControl,
+} from '@wordpress/components';
```
vcanales:
```suggestion
import {
	// eslint-disable-next-line @wordpress/no-unsafe-wp-apis
	__experimentalHStack as HStack,
	// eslint-disable-next-line @wordpress/no-unsafe-wp-apis
	__experimentalVStack as VStack,
	// eslint-disable-next-line @wordpress/no-unsafe-wp-apis
	__experimentalSpacer as Spacer,
	// eslint-disable-next-line @wordpress/no-unsafe-wp-apis
	__experimentalText as Text,
	Modal,
	Button,
	TextControl,
	TextareaControl,
} from '@wordpress/components';
```

src/landing-page/landing-page.js
```diff
@@ -0,0 +1,252 @@
...
+ */
+import { sprintf, __ } from '@wordpress/i18n';
+import { useState, createInterpolateElement } from '@wordpress/element';
+import { useSelect } from '@wordpress/data';
+import {
+	Button,
+	// eslint-disable-next-line
+	__experimentalVStack as VStack,
+	// eslint-disable-next-line
+	__experimentalHStack as HStack,
```
vcanales:
```suggestion
	// eslint-disable-next-line @wordpress/no-unsafe-wp-apis
	__experimentalVStack as VStack,
	// eslint-disable-next-line @wordpress/no-unsafe-wp-apis
	__experimentalHStack as HStack,
```

vcanales (commented):
Looking good! Just a few nitpicks regarding `eslint` rule deactivation

src/resolvers.js
```diff
@@ -29,6 +29,93 @@ export async function fetchThemeJson() {
...
+		if ( ! response?.data || 'SUCCESS' !== response?.status ) {
+			throw new Error(
+				`Failed to fetch style data: ${
+					response?.message || response?.status
+				}`
+			);
+		}
+		return response?.data;
+	} catch ( e ) {
+		// @todo: handle error
```
vcanales: Could we handle this error?

madhusudhand:
@beafialho may be we should make it a link pointing to the editor?

image

src/resolvers.js
pbking: Not there, in the resolvers.  The things calling the resolvers are handling the error.  But that todo would prevent the errors from being caught so I removed it (and standardized all of the functions while I was at it).

pbking:
@madhusudhand I added the link. Good idea.


image

src/admin-landing-page.scss
```diff
@@ -0,0 +1,58 @@
...
+	background-color: #fff;
+	margin-left: -20px;
+	a,
+	button {
+		color: #3858e9;
+	}
+	.cbt-lp-header {
+		width: 100%;
+		background-color: #2d59f2;
+	}
```
t-hamano:
```suggestion
	.cbt-lp-header {
		width: 100%;
		background-color: #2d59f2;

		img {
			vertical-align: top;
		}
	}
```

This is to remove the space below the image and make the header height exactly 270ox.

image

src/admin-landing-page.js
```diff
@@ -0,0 +1,31 @@
...
+		const domNode = document.getElementById( 'create-block-theme-app' );
+
+		// If version is less than 18 use `render` to render the app
+		// otherwise use `createRoot` to render the app
+		if ( createRoot === undefined ) {
+			render( <App />, domNode );
+		} else {
+			const root = createRoot( domNode );
+			root.render( <App /> );
+		}
```
t-hamano: Although I haven't tested it, it looks like just `createRoot` might work if the core version supported by CBT is 6.5 or later.

src/admin-landing-page.scss
```diff
@@ -0,0 +1,58 @@
...
+		}
+		h1,
+		h2,
+		h3,
+		h4,
+		h5,
+		h6 {
+			margin-bottom: 0.3em;
+		}
+		.cbt-lp-left-column {
```
t-hamano: For the mobile viewport, we will need to stack the columns vertically.

image

src/landing-page/landing-page.js
```diff
@@ -0,0 +1,260 @@
...
+			</div>
+
+			<HStack alignment="topLeft" className="cbt-lp-body">
+				<VStack alignment="left" className="cbt-lp-left-column">
+					<h1>
+						{ __(
+							'What would you like to do?',
+							'create-block-theme'
+						) }
+					</h1>
```
t-hamano: Should this be h1? Wouldn't a header image be more appropriate?

src/landing-page/create-modal.js
```diff
@@ -0,0 +1,169 @@
...
+				<br />
+				<HStack>
+					<Button
+						variant="primary"
+						disabled={ ! theme.name }
+						onClick={ () => onRequestClose() }
+					>
+						Close
+					</Button>
+				</HStack>
```
t-hamano: Personally, I don't think the close button is necessary. If necessary, I think at least a secondary variation would be good. `Close` also requires a translation function.

image

src/landing-page/create-modal.js
```diff
@@ -0,0 +1,169 @@
...
+			onRequestClose={ onRequestClose }
+		>
+			<VStack>
+				<Text>
+					{ __(
+						"Let's get started creating a new Block Theme.",
+						'create-block-theme'
+					) }
+				</Text>
+				<Spacer />
```
t-hamano: I don't recommend using `Spacer` component in this way. Because the height is zero. Can't we achieve the same with just `VStack` components?

src/landing-page/create-modal.js
```diff
@@ -0,0 +1,169 @@
...
+					) }
+				/>
+				<br />
+				<HStack>
+					<Button
+						variant="primary"
+						disabled={ ! theme.name }
+						onClick={ () => createBlockTheme() }
+					>
+						{ __( 'Create Block Theme', 'create-block-theme' ) }
```
t-hamano: It might be a good idea to rename this button to something more appropriate for the modal that is actually opened.

- Create Blank Theme
- Create Cloned Theme
- Create Child Theme

src/landing-page/landing-page.js
```diff
@@ -0,0 +1,260 @@
...
+				<CreateThemeModal
+					creationType={ createModalType }
+					onRequestClose={ () => setCreateModalType( false ) }
+				/>
+			) }
+
+			<div className="cbt-lp-header">
+				<img
+					src="/wp-content/plugins/create-block-theme/assets/header_logo.webp"
+					alt="Create Block Theme Logo"
```
t-hamano:
```suggestion
					alt={ __(
						'Create Block Theme Logo',
						'create-block-theme'
					) }
```

src/landing-page/landing-page.js
```diff
@@ -0,0 +1,260 @@
...
+			{ createModalType && (
+				<CreateThemeModal
+					creationType={ createModalType }
+					onRequestClose={ () => setCreateModalType( false ) }
+				/>
+			) }
+
+			<div className="cbt-lp-header">
+				<img
+					src="/wp-content/plugins/create-block-theme/assets/header_logo.webp"
```
t-hamano: Will this path still work when the WordPress site is installed in a subdirectory?

src/landing-page/landing-page.js
```diff
@@ -0,0 +1,260 @@
...
+								'Have a question? Ask for some help in the <a>forums</a>.',
+								'create-block-theme'
+							),
+							{
+								a: (
+									// eslint-disable-next-line jsx-a11y/anchor-has-content
+									<a href="https://wordpress.org/support/plugin/create-block-theme/" />
+								),
+							}
+						) }
```
t-hamano: Aren't the three links in this section external links? We might be able to use the `ExternalLink` component.

src/landing-page/landing-page.js
```diff
@@ -0,0 +1,260 @@
...
+					</Button>
+					<p>
+						{ __(
+							'Make a theme that uses the currently activated theme as a parent.',
+							'create-block-theme'
+						) }
+					</p>
+				</VStack>
+				<VStack className="cbt-lp-right-column">
+					<h4>{ __( 'About the Plugin', 'create-block-theme' ) }</h4>
```
t-hamano: Isn't this h2?

includes/class-create-block-theme-admin-landing.php
```diff
@@ -0,0 +1,49 @@
...
+
+		$asset_file = include plugin_dir_path( __DIR__ ) . 'build/admin-landing-page.asset.php';
+
+		// Enqueue CSS dependencies of the scripts included in the build.
+		foreach ( $asset_file['dependencies'] as $style ) {
+			wp_enqueue_style( $style );
+		}
+
+		// Enqueue CSS of the app
+		wp_enqueue_style( 'create-block-theme-app', plugins_url( 'build/admin-landing-page.css', __DIR__ ), array(), $asset_file['version'] );
```
t-hamano:
```suggestion
		// Enqueue CSS of the app
		wp_enqueue_style( 'create-block-theme-app', plugins_url( 'build/admin-landing-page.css', __DIR__ ), array( 'wp-components' ), $asset_file['version'] );
```

`$asset_file['dependencies']` also includes unnecessary items such as `react` and `wp-api-fetch`. The only required dependency should be `wp-components`.

Additionally, wouldn't `admin_enqueue_scripts` hook be more appropriate for enqueuing CSS and JS?

src/landing-page/create-modal.js
```diff
@@ -0,0 +1,169 @@
...
+				break;
+		}
+
+		if ( ! constructionFunction ) {
+			return;
+		}
+		constructionFunction( theme )
+			.then( () => {
+				// eslint-disable-next-line no-undef, no-alert
+				alert(
```
t-hamano:
```suggestion
				// eslint-disable-next-line no-alert
				window.alert(
```

src/landing-page/create-modal.js
```diff
@@ -0,0 +1,169 @@
...
+
+	if ( errorMessage ) {
+		return (
+			<Modal
+				title={ __( 'Create Block Theme', 'create-block-theme' ) }
+				onRequestClose={ onRequestClose }
+			>
+				<p>
+					<Text>{ errorMessage }</Text>
+				</p>
```
t-hamano:
```suggestion
				<Text as="p">{ errorMessage }</Text>
```

src/landing-page/landing-page.js
```diff
@@ -0,0 +1,260 @@
...
+							</summary>
+							<p>
+								{ __(
+									'There is a new panel accessible from the WordPress Editor which you can open by clicking on a new icon to the right of the “Save” button, at the top of the Editor.',
+									'create-block-theme'
+								) }
+							</p>
+							<img
+								src="/wp-content/plugins/create-block-theme/assets/faq_icon.webp"
+								alt="A screenshot of the Create Block Theme icon in the editor"
```
t-hamano:
```suggestion
								alt={ __(
									'A screenshot of the Create Block Theme icon in the editor',
									'create-block-theme'
								) }
```

src/landing-page/landing-page.js
```diff
@@ -0,0 +1,260 @@
...
+							</summary>
+							<p>
+								{ __(
+									'In the Create Block Theme Panel click "Save Changes to Theme". You will be presented with a number of options of which things you want to be saved to your theme. Make your choices and then click "Save Changes".',
+									'create-block-theme'
+								) }
+							</p>
+							<img
+								src="/wp-content/plugins/create-block-theme/assets/faq_save.webp"
+								alt="A screenshot of the Create Block Theme save changes panel"
```
t-hamano:
```suggestion
								alt={ __(
									'A screenshot of the Create Block Theme save changes panel',
									'create-block-theme'
								) }
```

src/landing-page/landing-page.js
```diff
@@ -0,0 +1,260 @@
...
+							</summary>
+							<p>
+								{ __(
+									'First Install and activate a font from any source using the WordPress Font Library. Then, using the Create Block Theme Panel select “Save Changes To Theme” and select “Save Fonts” before saving the theme. All of the active fonts will be activated in the theme and deactivated in the system (and may be safely deleted from the system). Any fonts that are installed in the theme that have been deactivated with the WordPress Font Library will be removed from the theme.',
+									'create-block-theme'
+								) }
+							</p>
+							<img
+								src="/wp-content/plugins/create-block-theme/assets/faq_fonts.webp"
+								alt="A screenshot of the WordPress Font Library modal window"
```
t-hamano:
```suggestion
								alt={ __(
									'A screenshot of the WordPress Font Library modal window',
									'create-block-theme'
								) }
```

t-hamano (commented):
I think it's a great UI. I may not fully understand the background to this, but I'd like to comment on something that caught my attention in the codebase.

src/admin-landing-page.scss
```diff
@@ -0,0 +1,58 @@
...
+@include wordpress-admin-schemes();
+
+.create-block-theme.landing-page {
+	background-color: #fff;
+	margin-left: -20px;
+	a,
+	button {
+		color: #3858e9;
+	}
+	.cbt-lp-header {
```
t-hamano: It may be a good idea to unify the selector rules added in #638. Probably `create-block-theme__{element-name}`?

src/resolvers.js
vcanales: In that case, I think the catch should either throw, or not be wrapped in a try catch statement at all.

src/admin-landing-page.js
pbking: You are correct sir.  6.5 is the minimum version for the plugin anyway.  Simplified.

src/landing-page/create-modal.js
pbking: I'm not sure that the <Text> component is helpful, however it's rendered.  

I've simplified this to just a <p> element (and removed the 'close' button per another suggestion)

image

src/landing-page/create-modal.js
pbking: I agree.  Close button removed.

image

src/landing-page/create-modal.js
pbking: Spacers removed, using `gap` instead.

I was surprised not to find that was a property of the VStack component (or did I simply not find it?)

src/landing-page/create-modal.js
pbking: Button named dynamically.  Easy enough.

src/landing-page/landing-page.js
pbking: No, probably not.  I'm not sure how to best accomplish this.

This is probably also true for the image assets that are referenced (the header, the FAQ images, etc).

includes/class-create-block-theme-admin-landing.php
pbking: Why wouldn't react and wp-api-fetch (and anything else in the file) be required?

src/landing-page/landing-page.js
pbking: Maybe, but I'm not sure how to use that component and also localize the bits that need localizing.  Happy for you to change it if you've a suggestion.  Happy for it to stay as-is; I'm not sure what using the component adds.

src/landing-page/landing-page.js
pbking: h3 I believe; adjusted and re-styled to be appropriate.

src/landing-page/landing-page.js
pbking: I believe h2 is more appropriate.  Adjusted it (and the other h4s) and restyled.

src/admin-landing-page.scss
pbking: I agree.  I have refactored to standardize on that.

src/landing-page/landing-page.js
vcanales: It's almost the same as if it were an anchor

https://developer.wordpress.org/block-editor/reference-guides/components/external-link/#usage

And for createInterpolateElement:

https://make.wordpress.org/core/2020/07/17/introducing-createinterpolateelement/

src/admin-landing-page.scss
pbking: Added some mobile styling:

image

src/landing-page/landing-page.js
pbking: Thank you both.

I have used that component:

image

src/landing-page/landing-page.js
pbking: Easily fixed by localizing the script: https://github.com/WordPress/create-block-theme/pull/637/commits/02adc55035b83f71dcb4b6e6228b333312363d99

pbking:
I believe all of the feedback has been addressed with one exception (that I'm not sure I understand).

There was some great reviewing in here!  

Maybe ready for another look?

src/landing-page/landing-page.js
```diff
@@ -0,0 +1,293 @@
...
+								'Want to contribute? Check out the <a>project on GitHub</a>.',
+								'create-block-theme'
+							),
+							{
+								a: (
+									<ExternalLink href="https://github.com/WordPress/create-block-theme" />
+								),
+							}
+						) }
+					</p>
```
vcanales: It's interesting that it works by replacing the anchor with the component, but strictly we should use the same element:

```suggestion
						{ createInterpolateElement(
							__(
								'Want to contribute? Check out the <ExternalLink>project on GitHub</ExternalLink>.',
								'create-block-theme'
							),
							{
								ExternalLink: (
									<ExternalLink href="https://github.com/WordPress/create-block-theme" />
								),
							}
						) }
					</p>
```

pbking: done

includes/class-create-block-theme-admin-landing.php
t-hamano: There was a slight error in my explanation. `$asset_file['dependencies']` contains the handle names of dependent _script_ libraries. It is not a style handle name. So I don't think this array is suitable for `wp_enqueue_style` dependencies.

t-hamano (commented):
`h1` no longer exists in the content 😅 I suggest making the title logo an h1, what do you think? In that case, I think we need to change the image alt from `Create Block Theme Logo` to `Create Block Theme`.

image

src/landing-page/create-modal.js
```diff
@@ -0,0 +1,165 @@
...
+			</Modal>
+		);
+	}
+
+	return (
+		<Modal
+			title={ __( 'Create Block Theme', 'create-block-theme' ) }
+			onRequestClose={ onRequestClose }
+		>
+			<VStack style={ { gap: '20px' } }>
```
t-hamano:
```suggestion
			<VStack spacing="5">
```

I think `spacing` prop is better than inline styles. This is equivalent to 20px.

includes/class-create-block-theme-admin-landing.php
```diff
@@ -0,0 +1,58 @@
...
+
+	public static function admin_menu_page() {
+
+		$asset_file = include plugin_dir_path( __DIR__ ) . 'build/admin-landing-page.asset.php';
+
+		// Enqueue CSS dependencies of the scripts included in the build.
+		foreach ( $asset_file['dependencies'] as $style ) {
+			wp_enqueue_style( $style );
+		}
+
```
t-hamano: I think this is no longer necessary.

includes/class-create-block-theme-admin-landing.php
```diff
@@ -0,0 +1,58 @@
...
+		// Enqueue CSS of the app
+		wp_enqueue_style( 'create-block-theme-app', plugins_url( 'build/admin-landing-page.css', __DIR__ ), array( 'wp-components' ), $asset_file['version'] );
+
+		// Load our app.js.
+		array_push( $asset_file['dependencies'], 'wp-i18n' );
+		wp_enqueue_script( 'create-block-theme-app', plugins_url( 'build/admin-landing-page.js', __DIR__ ), $asset_file['dependencies'], $asset_file['version'] );
```
t-hamano:
```suggestion
		// Load our app.js.
		wp_enqueue_script( 'create-block-theme-app', plugins_url( 'build/admin-landing-page.js', __DIR__ ), $asset_file['dependencies'], $asset_file['version'] );
```

I think `wp-i18n` is already included in `$asset_file['dependencies']`.

src/landing-page/landing-page.js
```diff
@@ -0,0 +1,293 @@
...
+import { fetchThemeStyleData, downloadExportedTheme } from '../resolvers';
+import { downloadFile } from '../utils';
+import { CreateThemeModal } from './create-modal';
+
+export default function LandingPage() {
+	const [ themeStyleData, setThemeStyleData ] = useState( '' );
+	const [ createModalType, setCreateModalType ] = useState( false );
+
+	useSelect( async () => {
+		setThemeStyleData( await fetchThemeStyleData() );
```
t-hamano: I don't think we need to add a custom endpoint since all we need is the theme name. How about just the code below?

```javascript
import { store as coreStore } from '@wordpress/core-data';

export default function LandingPage() {
	// ...

	const themeName = useSelect( ( select ) =>
		select( coreStore ).getCurrentTheme()
	)?.name?.raw;

	// ...
}

File: src/admin-landing-page.scss [hunk 32]
```diff
@@ -0,0 +1,80 @@
...
+				button {
+					font-size: 1.75em;
+					@media screen and (max-width: 775px) {
+						font-size: 1.25em;
+					}
+				}
+			}
+			&__right-column {
+				max-width: 330px;
+				margin: 0 60px;
```
t-hamano, line 57:
```suggestion
				max-width: 330px;
				margin: 0 60px;
				@media screen and (max-width: 775px) {
					max-width: 100%;
				}
```

| Before | After |
|--------|--------|
| image| image|

t-hamano (commented):
Thanks for the update! I have made additional suggestions.

Another thing that bothers me is that the translation files don't seem to be loading.

This plugin is [already translated into several locales](https://translate.wordpress.org/projects/wp-plugins/create-block-theme/stable/). And if I switch the site language to that locale, the script with the handle `create-block-theme-app-js-translations` should be enqueued, but it isn't.

For example, it is [100% translated in the Japanese locale](https://translate.wordpress.org/projects/wp-plugins/create-block-theme/stable/ja/default/), and the editor sidebar is fully translated. I was hoping that this new admin page would translate at least some of it, but it doesn't translate at all.

Will this not be translated unless the plugin is released and reflected in GlotPress?

src/landing-page/create-modal.js
pbking: Ah, that's what I was looking for.  Thank you.

src/landing-page/landing-page.js
pbking: I don't believe that we have access to the coreStore outside of the Editor... I might be wrong, but I think that this was giving some issues last time I tried to build an interface in this way.

src/landing-page/landing-page.js
```diff
@@ -0,0 +1,293 @@
...
+	}, [] );
```
vcanales: Referencing @t-hamano's comment, I've tried this and it works fine:

```suggestion
	const [ themeName, setThemeName ] = useState( '' );
	const [ createModalType, setCreateModalType ] = useState( false );

	const currentTheme = useSelect( ( select ) =>
		select( coreStore ).getCurrentTheme()
	);

	useEffect( () => {
		const name = currentTheme?.name?.raw || '';
		setThemeName( name );
	}, [ currentTheme ] );
```

Replacing references to `themeStyleData?.name` with `themeName`, and of course importing `useEffect` and the core data store:

```js
import {
	useState,
	createInterpolateElement,
	useEffect,
} from '@wordpress/element';
import { store as coreStore } from '@wordpress/core-data';
```

So I agree that we probably don't need the endpoint after all, at least not for the theme name.

t-hamano: The core API should work outside of the editor.

However, if we implement it like this, it won't work because the dependencies don't include `wp-core-data`.

```javascript
const themeName = useSelect( ( select ) =>
	select( 'core' ).getCurrentTheme()
)?.name?.raw;
```

If we want to use the core API outside of the editor, we need to explicitly import the core library.

```javascript
import { store as coreStore } from '@wordpress/core-data';

const themeName = useSelect( ( select ) =>
	select( coreStore ).getCurrentTheme()
)?.name?.raw;
```

Additionally, the theme name is only used in the button text, so there should be no need to manage it as a state.

```javascript
export default function LandingPage() {
	// ...
	const themeName = useSelect( ( select ) =>
		select( coreStore ).getCurrentTheme()
	)?.name?.raw;
	// ...

	return (
		// ...
		<Button>
			{ sprintf(
				__( 'Export "%s" as a Zip File', 'create-block-theme' ),
				themeName
			) }
		</Button>
		// ...
	);
}
```

pbking:
Latest round of adjustments are in.  Ready for another review I think!

pbking:
> Another thing that bothers me is that the translation files don't seem to be loading.
...
> Will this not be translated unless the plugin is released and reflected in GlotPress?

I've no idea why the translations wouldn't be working as expected.  That might be related...  Should we release this change and see?

vcanales (approved):
👍

src/resolvers.js
```diff
@@ -4,42 +4,92 @@
...
 				`Failed to fetch theme data: ${
 					response?.message || response?.status
 				}`
 			);
 		}
-
 		return JSON.stringify( response?.data, null, 2 );
-	} catch ( e ) {
-		// @todo: handle error
-	}
```
vcanales, line 29: Nitpick I guess, but prefer async/await instead of `.then`.

src/landing-page/landing-page.js
```diff
@@ -0,0 +1,300 @@
...
+	const [ createModalType, setCreateModalType ] = useState( false );
+
+	const currentTheme = useSelect( ( select ) =>
+		select( coreStore ).getCurrentTheme()
+	);
+
+	useEffect( () => {
+		const name = currentTheme?.name?.raw || '';
+		setThemeName( name );
+	}, [ currentTheme ] );
```
vcanales: As @t-hamano mentioned, it's not necessary to manage the `themeName` as a state variable.

```suggestion
export default function LandingPage() {
	const [ createModalType, setCreateModalType ] = useState( false );

	const themeName = useSelect( ( select ) =>
		select( coreStore ).getCurrentTheme()
	)?.name?.raw;
```

t-hamano (approved):
LGTM 🚀

I think if we address the comment left by @vcanales, we can ship this PR.

> Another thing that bothers me is that the translation files don't seem to be loading.

I checked the return value of `wp_set_script_translations()` and it is `true`, so there should be no problem with the implementation. Let's see if the translations will be reflected when this version is released and localized.

src/landing-page/landing-page.js
pbking: Ah, yes of course.

src/resolvers.js
pbking: That's the PHP programmer in you.  .then .catch can be a lovely thing. :)

vcanales: > That's the PHP programmer in you. .then .catch can be a lovely thing. :)

It's mostly about keeping a consistent codebase, since we use async/await on most other places.

It's also a bit pointless to have an `async function` that does not use `await` within; in this case, `apiFetch` is already returning a Promise.

PR #638, opened 2024-05-15
Update modal width to 65vw

madhusudhand:
This change brings the following UI changes.

* Update the modal width of metadata editor and theme.json viewer to `65vw` (similar to font library in core)
* Makes the modal footer absolutely position in the bottom

image

In addition to these UI changes, it also introduces a style-sheet for adding class definitions.


admin/class-create-theme.php
```diff
@@ -50,8 +50,16 @@ function create_block_theme_enqueue() {
...
+		wp_enqueue_style(
+			'create-block-theme-styles',
+			plugins_url( 'build/plugin-sidebar.css', dirname( __FILE__ ) ),
+			false,
+			$asset_file['version'],
+			'all'
+		);
 		wp_enqueue_script(
 			'create-block-theme-slot-fill',
+			'create-block-theme-styles'
```
t-hamano: Isn't this unnecessary? What is newly enqueued in this PR is a stylesheet, not a script.

src/plugin-styles.scss
```diff
@@ -0,0 +1,30 @@
+$plugin-prefix: "create-block-theme";
+$modal-footer-height: 70px;
+
+.#{$plugin-prefix} {
+	&__metadata-editor-modal {
+		width: 65vw !important;
```
t-hamano: If you define the width with CSS, It would be better to add a media query (`@media (min-width: 782px)`). See: https://github.com/WordPress/gutenberg/pull/54518

t-hamano (commented):
The `Modal` component can now define its width via [the `size` prop](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/modal/README.md#size-small--medium--large--fill). What do you think about using this prop instead of defining width in CSS? However, when using this props, the modal's height will not be full.

```javascript
<Modal
	size="large"
>
</Modal>
```

image

src/plugin-styles.scss
madhusudhand: `size="large"` seems to be fitting well.

admin/class-create-theme.php
madhusudhand: Nice catch. Thank you.

pbking:
What if additionally we shifted the 'screenshot' and 'subfolder' (two items that are pretty narrow) to a second, smaller column on the right?

image

admin/class-create-theme.php
```diff
@@ -51,6 +51,13 @@ function create_block_theme_enqueue() {
 			$asset_file['dependencies'],
 			$asset_file['version']
 		);
+		wp_enqueue_style(
+			'create-block-theme-styles',
+			plugins_url( 'build/plugin-sidebar.css', dirname( __FILE__ ) ),
+			false,
+			$asset_file['version'],
+			'all'
```
t-hamano:
```suggestion
			'create-block-theme-styles',
			plugins_url( 'build/plugin-sidebar.css', dirname( __FILE__ ) ),
			array(),
			$asset_file['version']
```

Nit:

- The type of the third argument `$deps` is `string[]`.
- The default value of the fifth argument `$media` is `all`, so I don't think it's necessary.

https://developer.wordpress.org/reference/functions/wp_enqueue_style/

t-hamano (approved):
LGTM!

t-hamano:
> What if additionally we shifted the 'screenshot' and 'subfolder' (two items that are pretty narrow) to a second, smaller column on the right?

I think this is an interesting idea and may be considered in another PR.

pbking (approved):
I merged /trunk and fixed a conflict and I think this is ready to come in too.  It looks nice; thank you for doing this. :)

matiasbenedetto:
I fixed the conflicts to merge this and include it in the next release.

PR #639, opened 2024-05-15
Remove unused `@wordpress/base-styles`

t-hamano:
This package was once added to introduce color schemes to the Manage Theme Fonts page (#419), but is no longer used anywhere.


pbking:
I'm planning to use it in the React app that will exist in the wp-admin space.

https://github.com/WordPress/create-block-theme/pull/637

I'm ... not sure what exactly that app will be... but I'm pretty sure it will use those styles. :)

t-hamano:
Oh, I had overlooked that such a development was taking place. Well then, I am happy to close this PR.

PR #640, opened 2024-05-15
Fix: Version field not updated

t-hamano:
I noticed that the version field was not taken into account in the `sanitize_theme_data` method, so we couldn't update this field.

image

This PR should also resolve the following PHP error that is logged when updating metadata.

```
PHP Warning:  Undefined array key "version" in /var/www/html/wp-content/plugins/create-block-theme/admin/create-theme/theme-styles.php on line 20
```

## Testing Instructions

- Opne the Edit Theme Metadata modal.
- Update "Version" field.
- Open style.css in your active theme and check the `Version:` header data.


pbking:
I've fixed the same in [this PR](https://github.com/WordPress/create-block-theme/pull/634) which also fixes some other values that get clobbered.

PR #641, opened 2024-05-15
Fix template texts localizing/escaping

### LINKED ISSUE #632: Localize option tries to localize already localized strings breaking the patters.
## What
The localize option tries to localize already localized strings, breaking the patterns.

Example output of saving TT4 theme using 'Localize Text' option enabled:
```
<!-- wp:paragraph {"align":"left"} -->
<p class="has-text-align-left"><?php echo __('Experience the fusion of imagination and expertise with Études <?php echo __('Architectural Solutions', 'twentytwentyfour');?>.', 'twentytwentyfour');?></p>
<!-- /wp:paragraph --></div>
``` 
Please observe that an echo statement is inside another echo statement, which will result in a parse error. This breaks the pattern where this is included.

## Expected
The localize functionality should not try to localize already localized text.

matiasbenedetto:
## What?
Fix template texts localizing/escaping.

Fixes: https://github.com/WordPress/create-block-theme/issues/632

## Why?
The current implementation consisting of:
1. Collect all the strings from the templates markup using regex.
2. Search and replace the strings collected.

This approach is flawed because it fails, for example, when there are substrings. The problem is explained in this comment: https://github.com/WordPress/create-block-theme/issues/632#issuecomment-2112602045

☣️ The severity of the problem is high because the resulting templates/patterns are broken because the plugin is producing invalid markup.

## How?
This PR improves the approach around escaping block strings for translation. It avoids the substring problem by just considering the HTML content block by block.
The new implementation consist of:
1. Parse the template markup with the core's function `parse_blocks().`
2. In the blocks that can contain text, use regex to update the HTML content of the block 'innetContent' property with the escaped text.
3.  Serialize the template blocks to markup again using core's `serialize_blocks()` function.

### DIFF
--- a/admin/create-theme/theme-locale.php
+++ b/admin/create-theme/theme-locale.php
@@ -29,9 +29,9 @@ class CBT_Theme_Locale {
 	 * Get a replacement pattern for escaping the text from the html content of a block.
 	 *
 	 * @param string $block_name The block name.
-	 * @return string|array|null The regex pattern to match the content that needs to be escaped.
+	 * @return array|null The regex patterns to match the content that needs to be escaped.
 	 *      Returns null if the block is not supported.
-	 *      Returns an array if the block has multiple html elements that need to be escaped.
+	 *      Returns an array of regex patterns if the block has html elements that need to be escaped.
 	 */
 	private static function get_text_replacement_patterns_for_html( $block_name ) {
 		switch ( $block_name ) {
--- a/admin/create-theme/theme-locale.php
+++ b/admin/create-theme/theme-locale.php
@@ -19,6 +19,47 @@ class CBT_Theme_Locale {
 		return "<?php echo __('" . $string . "', '" . wp_get_theme()->get( 'TextDomain' ) . "');?>";
 	}
 
+	/**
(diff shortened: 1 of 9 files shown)


admin/create-theme/theme-locale.php
```diff
@@ -0,0 +1,158 @@
...
+
+		$string = addcslashes( $string, "'" );
+		return "<?php echo __('" . $string . "', '" . wp_get_theme()->get( 'TextDomain' ) . "');?>";
+	}
+
+	/**
+	 * Get a replacement pattern for escaping the text from the html content of a block.
+	 *
+	 * @param string $block_name The block name.
+	 * @return string|array|null The regex pattern to match the content that needs to be escaped.
```
pbking: Looks like it always returns an array (which is my 'druthers), so just `array|null` right?

pbking:
I had one teeny tiny inconsequential nit.

This is great to bring in and fixes the problem at hand and enhances the tests quite a bit.

pbking (approved):

admin/create-theme/theme-locale.php
pbking: updated

PR #645, opened 2024-05-22
Fixed font utilities to work with font sources as an (optional) array.

### LINKED ISSUE #599: Critical error when removing fonts added using old naming convention in theme.json
When using the Save Changes option with Save Fonts ticked, we are getting the following error

~~~
PHP Fatal error:  Uncaught TypeError: basename(): Argument #1 ($path) must be of type string, array given in ... wp-content\plugins\create-block-theme\admin\create-theme\theme-fonts.php:117
~~~

We think this error occurs when create block theme is trying to remove fonts that were created in theme.json with the old naming convention and where the src is stored in theme.json as an array e.g.

~~~
	"fontFace": [
		{
			"fontFamily": "Ysabeau Office",
			"fontStyle": "normal",
			"fontWeight": "400",
			"src": [
				"file:./assets/fonts/ysabeau-office_normal_400.ttf"
			]
		},

~~~

We do not get the problem, when trying to remove fonts with the new naming convention where the src is stored as a string e.g. 

~~~
"src": "file:./assets/fonts/S6uyw4BMUTPHvxw6XweuBCY.woff2"
~~~

We have managed to work around this by temporarily changing the line of code 

~~~
$font_filename = basename( $font_face['src']);
~~~

to

~~~
$font_filename = basename( $font_face['src'][0]);
~~~

and removing fonts created the old way and reinstalling them the new way.

pbking:
When saving theme.json the `src` of font faces CAN be an array.

This change always writes that value as an array and can now operate on theme.json configurations that use arrays OR strings.

Fixes #599

### DIFF
--- a/includes/create-theme/theme-fonts.php
+++ b/includes/create-theme/theme-fonts.php
@@ -47,19 +47,25 @@ class CBT_Theme_Fonts {
 				continue;
 			}
 			foreach ( $font_family['fontFace'] as &$font_face ) {
-				$font_filename = basename( $font_face['src'] );
-				$font_dir      = wp_get_font_dir();
-				if ( str_contains( $font_face['src'], $font_dir['url'] ) ) {
-					// If the file is hosted on this server then copy it to the theme
-					copy( $font_dir['path'] . '/' . $font_filename, $theme_font_asset_location . '/' . $font_filename );
-				} else {
-					// otherwise download it from wherever it is hosted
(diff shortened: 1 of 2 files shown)


matiasbenedetto (approved):
This is working OK. I found a bug while testing this PR: https://github.com/WordPress/create-block-theme/issues/656. As it is somewhat related but isn't caused by this PR, I'm approving it.

PR #646, opened 2024-05-22
Move lib-font and add GPL license text to source files

matiasbenedetto:
## What?
Move lib-font and add GPL license text to source files.

## Why?
- We need to add the license text of the libs bundled in source code.
- This is part of the effort to re-introduce the font license theme automation (https://github.com/WordPress/create-block-theme/issues/525).


pbking (approved):
Looks fine to me

🚢

vcanales:
@matiasbenedetto Are we actually using lib-font? What about unbrotli? Actually... are we using any of these?

matiasbenedetto:
Yes, lib-font is used to read fonts metadata to extract font license info. Unbrotli nad inflate files contain the functions used by Lib-font to uncompress the fonts in compressed formats as .woff and .woff2.

vcanales:
I see. It's part of something that is not merged yet (https://github.com/WordPress/create-block-theme/pull/649)

PR #650, opened 2024-05-22
Enforce specifying which eslint rule is disabled when using eslint disable comments

vcanales:
Applying and enforcing this new rule reveals a lot of, for instance, no-undef rules being broken.


pbking (approved):
LGTM

🚢

PR #651, opened 2024-05-23
Fix the jslint warning from useSelect usage

pbking:
Fix the jslint warning (or infinate loop error when fixed) from useSelect usage.

Pulled the subfolder value (once) and set it in the initial call to `useState` rather than using that state in the `useSelect` when fetching the current theme's subfolder.

Fixes jslint warning and everything works as expected.

(To test, create a new theme using the Editor interface.  It should be created in the same subfolder as the currently activated theme)

### DIFF
--- a/src/editor-sidebar/create-panel.js
+++ b/src/editor-sidebar/create-panel.js
@@ -32,6 +32,14 @@ import ScreenHeader from './screen-header';
 export const CreateThemePanel = ( { createType } ) => {
 	const { createErrorNotice } = useDispatch( noticesStore );
 
+	const subfolder = useSelect( ( select ) => {
+		const stylesheet = select( 'core' ).getCurrentTheme().stylesheet;
+		if ( stylesheet.lastIndexOf( '/' ) > 1 ) {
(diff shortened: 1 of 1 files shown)


src/editor-sidebar/create-panel.js
```diff
@@ -32,30 +32,24 @@ import ScreenHeader from './screen-header';
...
 	const { createErrorNotice } = useDispatch( noticesStore );
 
+	const subfolder = useSelect( ( select ) => {
+		const stylesheet = select( 'core' ).getCurrentTheme().stylesheet;
+		if ( stylesheet.lastIndexOf( '/' ) > 1 ) {
+			return stylesheet.substring( 0, stylesheet.lastIndexOf( '/' ) );
+		}
+		return '';
+	}, [] );
+
```
vcanales, line 42: I've never tried setting initial state based on a value provided by a function called within the same component, but it doesn't seem to be a problem... so I'm inclined to approve this, albeit hesitantly

vcanales (approved):

PR #652, opened 2024-05-23
Added subfolder to initial theme state to eliminate render error

pbking:
To test:

Load the Metadata Modal window with the browser console open.

Note that no errors are shown.  (There should be no other changes.)

image

### DIFF
--- a/src/editor-sidebar/metadata-editor-modal.js
+++ b/src/editor-sidebar/metadata-editor-modal.js
@@ -48,6 +48,7 @@ export const ThemeMetadataEditorModal = ( { onRequestClose } ) => {
 		author_uri: '',
 		tags_custom: '',
 		recommended_plugins: '',
+		subfolder: '',
 	} );
 
 	const { createErrorNotice } = useDispatch( noticesStore );


vcanales (approved):
This change is OK to fix the error, but I'm now noticing that each panel is pulling theme data individually.

I think we should consider getting the data on the parent and share the data down, so that we may avoid repeatedly fetching the theme data, what do you think? (For another time, not now)

pbking:
> I think we should consider getting the data on the parent and share the data down, so that we may avoid repeatedly fetching the theme data, what do you think? (For another time, not now)

I'm unconcerned about the performance of repeated calls.  I would only want to refactor if it makes the code simpler or better since it ultimately doesn't effect the user's experience.  That said, I do think that consolidating all of the data fetching logic (into resolvers) and cacheing that so that it's not fetched every time would both improve the code and minimize the data.

vcanales:
> I'm unconcerned about the performance of repeated calls. I would only want to refactor if it makes the code simpler or better since it ultimately doesn't effect the user's experience.

Yeah, as you say, my concern is not about performance, but about the repetition of code in multiple children of a component that could do the fetching for them.

PR #654, opened 2024-05-29
Flush cache after saving/creating/exporting themes

### LINKED ISSUE #655: Cache is preventing that edited themes render as expected
## What?
When the user makes changes to a theme and save the changes using CBT in platforms using WP_Cache, the editor and the site frontend could fail to render the site because of the cache.

## How to reproduce
1. Create a WordPress site in a platform [using Object cache](https://pressable.com/knowledgebase/object-caching-memcached/) (example: Pressable).
2. Install Create Block Theme plugin.
3. Open the site editor, change the title of the default theme, and save the changes.
4. Go to CBT sidebar and click 'Save theme changes'.
5. The editor page will refresh, and you will see a blank page.

## Screencast
This screencast reproduces the issue and also shows how to fix it manually flushing the sites cache.

[Screencast from 29-05-24 21:22:38.webm](https://github.com/WordPress/create-block-theme/assets/1310626/d91602b2-3461-4a8f-8f5d-be2019942a60)

matiasbenedetto:
## What?
This pull request includes changes to flush the cache after creating new themes. 

Fixes: https://github.com/WordPress/create-block-theme/issues/655

## Why?
When the cache is not flushed, in plataforms using the WordPress cache utils, it could result in the editor loading a blank page. 
This ensures that the cache is updated with the latest theme data.

## How?
By calling `wp_cache_flush();` after the creation of new themes and variations.

## How to test.
- Create a site in pressable.
-  Install and activate CBT plugin with this patch.
-  Make a change to a template text and save the changes in the editor.
- Save the changes to the theme using the CBT  sidebar 'Save theme changes' button.
- Check that the changes were saved successfully and loaded in the editor as expected.

### DIFF
--- a/includes/class-create-block-theme-api.php
+++ b/includes/class-create-block-theme-api.php
@@ -382,7 +382,6 @@ class CBT_Theme_API {
 		}
 
 		$zip->close();
-		wp_cache_flush();
 
 		wp_cache_flush();
 
--- a/includes/class-create-block-theme-api.php
+++ b/includes/class-create-block-theme-api.php
@@ -203,6 +203,8 @@ class CBT_Theme_API {
 
 		$response = CBT_Theme_Create::clone_current_theme( $this->sanitize_theme_data( $request->get_params() ) );
 
+		wp_cache_flush();
+
 		if ( is_wp_error( $response ) ) {
 			return $response;
 		}
@@ -224,6 +226,8 @@ class CBT_Theme_API {
 
 		$response = CBT_Theme_Create::create_child_theme( $theme, $screenshot );
 
+		wp_cache_flush();
+
 		if ( is_wp_error( $response ) ) {
 			return $response;
 		}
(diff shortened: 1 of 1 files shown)


includes/class-create-block-theme-api.php
```diff
@@ -372,6 +382,9 @@ function rest_export_child_cloned_theme( $request ) {
 		}
 
 		$zip->close();
+		wp_cache_flush();
+
+		wp_cache_flush();
```
mikachan, line 386: I think this is a typo - I'm guessing we don't need to flush the cache twice here 👀

matiasbenedetto: haha, true. Fixed.

mikachan (approved):
This looks like it's working well, good detective work 👏

I tested by making sure I could reproduce the error on a new Pressable site, and then with the changes in this PR I updated templates in both TT4 and TT3 via CBT. The editor reloaded correctly both times 🎉

PR #660, opened 2024-05-31
Check if theme fonts present before removing

vcanales:
Fixes: https://github.com/WordPress/create-block-theme/issues/659

## What?
This PR implements a check to verify the presence of theme fonts before attempting to remove them. Additionally, it restructures the handling of deactivated font removal to ensure safety when theme fonts are not present.

## Why?
This PR addresses a critical error that occurs when saving changes to a theme after modifying font settings in a child theme, as detailed in GitHub issue #659. The issue resulted from an unhandled scenario where `array_filter()` was being called on a `null` value, leading to a fatal error.

## How?
The main changes include:
- Adding a check to ensure that theme font families are present before attempting their removal.
- Splitting the deactivation and asset removal process into separate functions for better clarity and error handling.
- Adjusting the logic to bail out early if theme font families are not defined, to prevent errors when `array_filter` operates on `null`.

## Testing Instructions
1. Create a child theme from TT4.
3. Click the "Save Changes to Theme" button.
4. Verify that no errors occur, and the changes are saved successfully.

PR #661, opened 2024-05-31
Refactor theme fonts class for readability

vcanales:
## What?
This Pull Request refactors the them fonts class to improve code readability and maintainability. The main focus is on cleaning up the code.

I felt inclined to do this because when debugging for https://github.com/WordPress/create-block-theme/pull/660, I found some parts of the code hard to follow.

## Why?
This PR addresses the need to make the code more readable and maintainable by:
1. Adding documentation and comments.
2. Simplifying functions and improving their readability.

### Key Changes:
1. **Class documentation added**: Added descriptions and the purpose of the `CBT_Theme_Fonts` class.

2. **Function improvements**:
   - *Inline Descriptions/Justifications*:
     - Refactored `make_theme_font_src_absolute()` to use `array_map` for cleaner array handling.
     - Replaced double newlines and improved spacing for better readability.
   - Changed `foreach` to `array_map`: This reduces the complexity and makes the handling of arrays cleaner.

3. **Code clean-up**:
   - Removed some redundant comments in favor of "self documenting" practices.
   - Used shorthand assignment (`?? null`) for checking array keys.
   - Introduced variables for commonly used expressions.

4. **Function calls**:
   - Replaced `CBT_Theme_Fonts::` with `self::` for static method calls within the class, enhancing readability and context awareness.

5. **Added test cases** for parts of the code that could break on further refactors.

### DIFF
--- a/includes/create-theme/theme-fonts.php
+++ b/includes/create-theme/theme-fonts.php
@@ -1,8 +1,10 @@
 <?php
-
-
+/**
+ * Theme Fonts
+ *
+ * @package Create_Block_Theme
+ */
 class CBT_Theme_Fonts {
-
 
 	/**
 	 * Make the font face theme src urls absolute.
@@ -16,7 +18,7 @@ class CBT_Theme_Fonts {
 	 * @return array|string
 	 */
 	private static function make_theme_font_src_absolute( $src ) {
-		$make_absolute = function ( $url ) {
+		$make_absolute = function( $url ) {
 			if ( str_starts_with( $url, 'file:./' ) ) {
 				return str_replace( 'file:./', get_stylesheet_directory_uri() . '/', $url );
 			}
@@ -24,14 +26,10 @@ class CBT_Theme_Fonts {
 		};
 
 		if ( is_array( $src ) ) {
-			foreach ( $src as &$url ) {
-				$url = $make_absolute( $url );
-			}
-		} else {
-			$src = $make_absolute( $src );
+			return array_map( $make_absolute, $src );
 		}
 
-		return $src;
+		return $make_absolute( $src );
 	}
 
 	/**
@@ -64,7 +62,7 @@ class CBT_Theme_Fonts {
 		foreach ( $font_families as &$font_family ) {
 			if ( isset( $font_family['fontFace'] ) ) {
 				foreach ( $font_family['fontFace'] as &$font_face ) {
-					$font_face['src'] = CBT_Theme_Fonts::make_theme_font_src_absolute( $font_face['src'] );
+					$font_face['src'] = self::make_theme_font_src_absolute( $font_face['src'] );
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


matiasbenedetto (commented):
Probably we should use this refactor to improve the tests of this class. They currently relay on setting users, creating new themes dynamically and after removing it without restoring the previous state.

Example:

https://github.com/WordPress/create-block-theme/blob/eabd38201d6a7cbb1631be4cf9eae820fe838028/tests/test-theme-fonts.php#L26-L60

 I think this kind of test is the cause of the destructive tests we found some time ago and we tried to workaround it here: https://github.com/WordPress/create-block-theme/pull/618

Could we follow the approach we are using in the [base case for the Readme class](https://github.com/WordPress/create-block-theme/blob/trunk/tests/CbtThemeReadme/base.php) ?. It can improve 3 aspects of these tests:

- Fine tune control and always predictable active theme for test.
- No destructive test env.
- Avoid realying on CBT complex functions.

tests/test-theme-fonts.php
```diff
@@ -214,6 +264,117 @@ public function test_get_all_fonts_from_theme_and_variation() {
...
+		// Verify that the src is correctly handled
+		$this->assertCount( 1, $fonts );
+		$this->assertEquals( 'single-src-font', $fonts[0]['slug'] );
+		$this->assertEquals( get_stylesheet_directory_uri() . '/assets/fonts/single-src-font.ttf', $fonts[0]['fontFace'][0]['src'] );
+
+		$this->uninstall_theme( $test_theme_slug );
+	}
+
+	public function test_array_font_src() {
+		wp_set_current_user( self::$admin_id );
```
matiasbenedetto, line 304: This is probably not needed.

vcanales: Without it, the tests break:

```php
Time: 00:00.357, Memory: 42.50 MB

There was 1 error:

1) Test_Create_Block_Theme_Fonts::test_copy_activated_fonts_to_theme
Undefined array key "typography"

/var/www/html/wp-content/plugins/create-block-theme/tests/test-theme-fonts.php:43

ERRORS!
Tests: 67, Assertions: 163, Errors: 1.
Script phpunit handling the test event returned with error code 2
✖ Command failed with exit code 2
Command failed with exit code 2
```

vcanales: > Probably we should use this refactor to improve the tests of this class. They currently relay on setting users, creating new themes dynamically and after removing it without restoring the previous state.

I agree that we should work on that, but since the tests currently cover the already-refactored code, I'd prefer refactoring them in another PR.

includes/create-theme/theme-fonts.php
```diff
@@ -74,53 +72,42 @@ public static function get_all_fonts() {
...
 
 	public static function copy_activated_fonts_to_theme() {
-		$user_settings = CBT_Theme_JSON_Resolver::get_user_data()->get_settings();
-		if ( ! isset( $user_settings['typography']['fontFamilies']['custom'] ) ) {
-			return null;
-		}
-
-		$font_families_to_copy = $user_settings['typography']['fontFamilies']['custom'];
+		$user_settings         = CBT_Theme_JSON_Resolver::get_user_data()->get_settings();
+		$font_families_to_copy = $user_settings['typography']['fontFamilies']['custom'] ?? null;
```
pbking:
```suggestion
		$font_families_to_copy = self::get_user_activated_fonts();
```

pbking (approved):
Seems fine to me.  Quality improvements.

PR #663, opened 2024-06-01
Make external links translatable

t-hamano:
This PR applies a translation function to some external links, since they may be locale-specific.

However, the following links remain unchanged:

---

https://github.com/WordPress/create-block-theme/blob/eabd38201d6a7cbb1631be4cf9eae820fe838028/src/landing-page/landing-page.js#L193
https://github.com/WordPress/create-block-theme/blob/eabd38201d6a7cbb1631be4cf9eae820fe838028/src/landing-page/landing-page.js#L206

GitHub URLs are the same regardless of locale.

---

https://github.com/WordPress/create-block-theme/blob/eabd38201d6a7cbb1631be4cf9eae820fe838028/assets/boilerplate/parts/footer.html#L6

Currently it is not possible to make strings in HTML templates translatable. See the following core ticket for more information: https://core.trac.wordpress.org/ticket/60298

### DIFF
--- a/src/editor-sidebar/metadata-editor-modal.js
+++ b/src/editor-sidebar/metadata-editor-modal.js
@@ -224,7 +224,10 @@ export const ThemeMetadataEditorModal = ( { onRequestClose } ) => {
 					} }
 				>
 					<ExternalLink
-						href="https://make.wordpress.org/themes/handbook/review/required/theme-tags/"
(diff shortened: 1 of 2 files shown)


madhusudhand (approved):
LGTM!

t-hamano:
Thanks for the review!

PR #664, opened 2024-06-01
Quality: Remove unused PHP classes

t-hamano:
Since the settings page has been rebuilt using React, the following two PHP classes are no longer used anywhere:

- `CBT_Form_Messages`
- `CBT_Theme_Form `


madhusudhand (approved):
LGTM!

PR #665, opened 2024-06-03
Update escaping function

### LINKED ISSUE #590: Escape translated strings
If you save changes made to a PHP file (like a pattern) with text strings, CBT will output the text strings with the `__()` localize function, like so:

```
<!-- wp:heading {"level":1} -->
<h1 class="wp-block-heading"><?php echo __( 'Posts', 'cozy-grove' ); ?></h1>
<!-- /wp:heading -->
```

But [according to](https://developer.wordpress.org/apis/security/escaping/) the WordPress Developer Handbooks, all output should be run through an escaping function. If you run the code generated by CBT through PHP_CodeSniffer with the WordPress Coding Standard enabled, it gives you an error for the lines with the unescaped strings.

Maybe [`wp_kses_post()`](https://developer.wordpress.org/reference/functions/wp_kses_post/), like so:

```
<!-- wp:heading {"level":1} -->
<h1 class="wp-block-heading"><?php echo wp_kses_post( __( 'Posts', 'cozy-grove' ) ); ?></h1>
<!-- /wp:heading -->
```

matiasbenedetto:
## What?
Replace a simple translate function call with an escaping and translating function when escaping is missing.

## Why?
Fixes: https://github.com/WordPress/create-block-theme/issues/590

### DIFF
--- a/includes/create-theme/theme-locale.php
+++ b/includes/create-theme/theme-locale.php
@@ -17,12 +17,16 @@ class CBT_Theme_Locale {
 		}
 
 		// Check if the text is already escaped.
-		if ( str_starts_with( $string, '<?php echo' ) ) {
+		if (
+			str_starts_with( $string, '<?php echo' ) ||
+			str_starts_with( $string, '<?php esc_html_e' ) ||
+			str_starts_with( $string, '<?php esc_html' )
+		) {
 			return $string;
 		}
 
 		$string = addcslashes( $string, "'" );
-		return "<?php echo __('" . $string . "', '" . wp_get_theme()->get( 'TextDomain' ) . "');?>";
+		return "<?php esc_html_e('" . $string . "', '" . wp_get_theme()->get( 'TextDomain' ) . "');?>";
 	}
 
 	/**
--- a/includes/create-theme/theme-locale.php
+++ b/includes/create-theme/theme-locale.php
@@ -17,11 +17,7 @@ class CBT_Theme_Locale {
 		}
 
 		// Check if the text is already escaped.
-		if (
-			str_starts_with( $string, '<?php echo' ) ||
-			str_starts_with( $string, '<?php esc_html_e' ) ||
-			str_starts_with( $string, '<?php esc_html' )
-		) {
+		if ( str_starts_with( $string, '<?php' ) ) {
 			return $string;
 		}
 
--- a/tests/CbtThemeLocale/escapeString.php
+++ b/tests/CbtThemeLocale/escapeString.php
@@ -13,29 +13,29 @@ class CBT_Theme_Locale_EscapeString extends CBT_Theme_Locale_UnitTestCase {
 	public function test_escape_string() {
 		$string         = 'This is a test text.';
 		$escaped_string = CBT_Theme_Locale::escape_string( $string );
-		$this->assertEquals( "<?php echo __('This is a test text.', 'test-locale-theme');?>", $escaped_string );
+		$this->assertEquals( "<?php esc_html_e('This is a test text.', 'test-locale-theme');?>", $escaped_string );
 	}
 
 	public function test_escape_string_with_single_quote() {
 		$string         = "This is a test text with a single quote '";
 		$escaped_string = CBT_Theme_Locale::escape_string( $string );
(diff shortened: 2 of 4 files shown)


includes/create-theme/theme-locale.php
```diff
@@ -17,12 +17,16 @@ public static function escape_string( $string ) {
...
 
 		// Check if the text is already escaped.
-		if ( str_starts_with( $string, '<?php echo' ) ) {
+		if (
+			str_starts_with( $string, '<?php echo' ) ||
+			str_starts_with( $string, '<?php esc_html_e' ) ||
+			str_starts_with( $string, '<?php esc_html' )
+		) {
 			return $string;
 		}
```
vcanales: There are plenty [other escaping functions](https://developer.wordpress.org/apis/security/escaping/#toc_1) that might show up. I'm wondering if we could instead try to catch them all with regex:

```suggestion
		if (
			preg_match(
				'/^<\?php\s+(echo|esc_)/',
				$string
			)
		) {
			return $string;
		}
```

All current escaping functions match that pattern, but we could also just add them all to avoid catching unintended functions:

```php
		if (
			preg_match(
				'/^<\?php\s+(echo|esc_html_e|esc_html__|esc_html_x|esc_attr_e|esc_attr__|esc_attr_x|esc_url|esc_js|esc_textarea)/',
				$string
			)
		) {
			return $string;
		}
```

matiasbenedetto: Good point but I think we could simplify it by just checking if the string starts with `<?php`. That will cover all the cases in a simpler way. Implemented it here: https://github.com/WordPress/create-block-theme/pull/665/commits/9b223b79813936882f943f1e2910e636e05d32ce

tests/CbtThemeLocale/escapeString.php
```diff
@@ -13,29 +13,29 @@ class CBT_Theme_Locale_EscapeString extends CBT_Theme_Locale_UnitTestCase {
 	public function test_escape_string() {
 		$string         = 'This is a test text.';
 		$escaped_string = CBT_Theme_Locale::escape_string( $string );
-		$this->assertEquals( "<?php echo __('This is a test text.', 'test-locale-theme');?>", $escaped_string );
+		$this->assertEquals( "<?php esc_html_e('This is a test text.', 'test-locale-theme');?>", $escaped_string );
```
vcanales, line 16: Unrelated to the changes, but why are we using CamelCase for these test files? It'd be better if we kept the convention we're already using for the other files. Same for the directory name.

matiasbenedetto: That's the WordPress core convention I think we should follow it. Example: https://github.com/WordPress/wordpress-develop/tree/trunk/tests/phpunit/tests/blocks

matiasbenedetto: > if we kept the convention we're already using for the other files. Same for the directory name.

Those should be progressively updated when we have the chance.

matiasbenedetto: I submitted an issue for standardizing the convention used on tests: https://github.com/WordPress/create-block-theme/issues/666

vcanales: I don't think those tests follow the actual convention either:

https://make.wordpress.org/docs/style-guide/formatting/filenames/#naming-files

> Use lowercase file, folder, and directory names. In general, separate words in filenames with hyphens, not underscores. Use standard [ASCII alphanumeric characters](https://wikipedia.org/wiki/ASCII#Character_set) in file, folder, and directory names.

vcanales (approved):
LGTM

tests/CbtThemeLocale/escapeString.php
matiasbenedetto: > Use lowercase file, folder, and directory names.

I'm not sure because, as far as I saw, the are numerous core test files using camel case.

Example 2: https://github.com/WordPress/wordpress-develop/tree/trunk/tests/phpunit/tests/blocks
Example 1: https://github.com/WordPress/wordpress-develop/tree/trunk/tests/phpunit/tests/fonts/font-face

PR #667, opened 2024-06-03
Add an about section in the editor

matiasbenedetto:
## What?
Add an 'About' section in the editor following the example from the admin page that will be deprecated in the next versions.

## Why?
- To provide some help to new users of the plugin.
- The admin page could be deprecated in the next versions.

## Screencast:
[Screencast from 03-06-24 13:33:21.webm](https://github.com/WordPress/create-block-theme/assets/1310626/a41467a4-908f-4b68-94a7-d66e62321532)

### DIFF
--- a/src/editor-sidebar/about.js
+++ b/src/editor-sidebar/about.js
@@ -7,6 +7,8 @@ import {
 	__experimentalVStack as VStack,
 	// eslint-disable-next-line @wordpress/no-unsafe-wp-apis
 	__experimentalText as Text,
+	// eslint-disable-next-line @wordpress/no-unsafe-wp-apis
+	__experimentalDivider as Divider,
 	PanelBody,
 	ExternalLink,
 } from '@wordpress/components';
@@ -37,7 +39,7 @@ function AboutPlugin() {
 					) }
 				</Text>
 
-				<hr />
+				<Divider />
 
 				<Text weight="bold">
 					{ __( 'Help', 'create-block-theme' ) }
--- a/src/editor-sidebar/about.js
+++ b/src/editor-sidebar/about.js
@@ -1,0 +1,86 @@
+/**
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import {
+	// eslint-disable-next-line @wordpress/no-unsafe-wp-apis
+	__experimentalVStack as VStack,
+	// eslint-disable-next-line @wordpress/no-unsafe-wp-apis
+	__experimentalText as Text,
+	PanelBody,
+	ExternalLink,
+} from '@wordpress/components';
+
+/**
+ * Internal dependencies
+ */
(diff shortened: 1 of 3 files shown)


src/editor-sidebar/about.js
```diff
@@ -0,0 +1,86 @@
...
+				</Text>
+
+				<Text>
+					{ __(
+						"Themes created with Create Block Theme don't require Create Block Theme to be installed on the site where the theme is used.",
+						'create-block-theme'
+					) }
+				</Text>
+
+				<hr />
```
mikachan: It looks like we use a `CardDivider` rather than a `hr` in Gutenberg, I think we should use them here as well.

src/plugin-sidebar.js
```diff
@@ -183,6 +185,26 @@ const CreateBlockThemePlugin = () => {
 										<Icon icon={ chevronRight } />
 									</HStack>
 								</NavigatorButton>
+
+								<hr></hr>
```
mikachan: I think this should be a `CardDivider` too, rather than two `hr`s.

src/plugin-styles.scss
```diff
@@ -22,4 +22,15 @@ $modal-footer-height: 70px;
 			object-fit: cover;
 		}
 	}
+
+	&__plugin-sidebar {
+		&__about-button {
+			color: #949494;
```
mikachan: Should we use variables from`@wordpress/base-styles` here instead of hard-coded values? Looks like this would be `$gray-600`.

mikachan (commented):
Looks good! I've left some comments related to keeping the markup and styling more in line with GB.

src/editor-sidebar/about.js
matiasbenedetto: It seems like `CardDivider` is intended to be used inside a `Card` component. I replaced the `<hr/>` with `Divider` components instead: https://github.com/WordPress/create-block-theme/pull/667/commits/93ff377ee2cf54492b8e1e474a33c3c60ff9ff78

src/plugin-sidebar.js
matiasbenedetto: It seems like CardDivider is intended to be used inside a Card component. I replaced the `<hr/>`s with Divider components instead: https://github.com/WordPress/create-block-theme/commit/93ff377ee2cf54492b8e1e474a33c3c60ff9ff78

src/editor-sidebar/about.js
mikachan: Oh I see, looks good!

src/plugin-styles.scss
matiasbenedetto: Done here: https://github.com/WordPress/create-block-theme/pull/667/commits/38c85125eb76b142a95c9081bbf7315f971eca4a

mikachan (approved):
LGTM!

PR #670, opened 2024-06-10
Remove unused REST API endpoint

t-hamano:
This PR removes the following two REST API endpoints:

- `/create-block-theme/v1/export-clone`
- `/create-block-theme/v1/export-child-clone`

As far as I can tell, these two endpoints are not being called anywhere. Perhaps they are no longer in use due to a redesign of the settings page and the plugin sidebar, etc.


matiasbenedetto (approved):
Looks good to me.

Issue #612, opened 2024-05-01
Unexpected results in exported ZIP


erikjoling:
Hi, 

**Prequisite**: I'm using Windows 11 and LocalWP.

When I export the theme using the Site Editor the name of the themefolder inside the ZIP is not the same as my installed theme. It seems like it uses the textdomain as a foldername. When using the legacy settings page (Appearance > Create Block Theme) the themefolder has the correct name.

Furthermore, the themefolder inside the ZIP file contains an empty folder (root?) with type "local disk".  This file contains some of my image assets.

![image](https://github.com/WordPress/create-block-theme/assets/20614110/5a73f33f-3261-406a-bfce-554a87807a99)

pbking:
> When using the legacy settings page (Appearance > Create Block Theme) the themefolder has the correct name.

What is the name of the folder you are expecting if not the theme's slug?  

Could you share the .zip file that was created?  

That "Local Disk" you're showing is certainly interesting.

erikjoling:
>> When using the legacy settings page (Appearance > Create Block Theme) the themefolder has the correct name.
>
> What is the name of the folder you are expecting if not the theme's slug?

I would expect the same folder name as the current folder name of the theme. In my example the theme's folder is named `hetgroenedoel` but the themefolder inside the exported .zip file is named `hgd`, which is my theme's slug or textdomain.

> That "Local Disk" you're showing is certainly interesting.

After changing some of the theme files the "Local Disk" (root) folder was gone. I'll let you know if it happens again.

t-hamano:
> Furthermore, the themefolder inside the ZIP file contains an empty folder (root?) with type "local disk". This file contains some of my image assets.

I was able to reproduce this issue with the default theme. The results of the test may vary depending on the type of file archiver, but I was able to reproduce it with two archivers: [7-zip](https://www.7-zip.org/) and [WinRar](https://www.win-rar.com/start.html?&L=13).

- Initialize your local environment once.
- Export the theme from the plugin sidebar of the site editor. If you check the zip file, you will see the following contents.
  ![image](https://github.com/WordPress/create-block-theme/assets/54422211/610f57b1-7dce-41be-ab72-4de06276396b)
- Make changes to the Blog Home template and save it.
- Export the theme from the plugin sidebar of the site editor. If you check the zip file with the two programs, you will see the following contents.
  - 7-zip:
[...]

PR #671, opened 2024-06-10
Quality: Fix warning error when exporting theme

t-hamano:
I discovered this while investigating issue #612.

https://github.com/WordPress/create-block-theme/assets/54422211/8da13a84-c74f-4698-8222-01b824c75e83

I don't know the exact cause, but it seems that this problem can be solved by adding an `exit;` to the end of the callback function.

The `exit;` also exists in [the core export function](https://github.com/WordPress/wordpress-develop/blob/bca94a5fe45f9414f0d11a16372895a4817b1716/src/wp-includes/rest-api/endpoints/class-wp-rest-edit-site-export-controller.php#L92).

### Testing Instructions

- Add `define( 'WP_DEBUG_LOG', true );` to the `wp-config.php`.
- Confirm that no PHP errors are logged when you export the theme.

### DIFF
--- a/src/utils/download-file.js
+++ b/src/utils/download-file.js
@@ -1,3 +1,8 @@
+/**
+ * WordPress dependencies
+ */
+import { downloadBlob } from '@wordpress/blob';
+
 /*
  * Download a file from in a browser.
  *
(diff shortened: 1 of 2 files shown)


src/utils/download-file.js
```diff
@@ -9,25 +14,5 @@ export default async function downloadFile( response ) {
...
-		a.download = filename;
-		document.body.appendChild( a );
-		a.click();
-		document.body.removeChild( a );
-
-		setTimeout( () => {
-			URL.revokeObjectURL( url );
-		}, 100 );
-	}
+	downloadBlob( filename, blob, 'application/zip' );
```
t-hamano, line 17: This is a minor refactoring. With this change, we no longer check if the browser is IE, but this should not be a problem since WordPress itself does not support IE.

pbking:
> I don't know the exact cause, but it seems that this problem can be solved by adding an exit; to the end of the callback function.

Yes, I believe that in other "export" logic we have had in CBT we likewise called `exit` for that reason.  It seems that wasn't carried over back when that transitioned to the logic we have now.

Thank you for cleaning that up.

src/utils/download-file.js
pbking: I was actually unfamiliar with `msSaveBlob` until looking into this change.  Likewise WordPress' implementation of the anchor downloading technique so I'm happy to replace that.  It works as expected.

Thank's for cleaning that up.

pbking (approved):

PR #672, opened 2024-06-15
a11y: Improve color contrast for help button

t-hamano:
The text color for the help button is [$gray-600](https://github.com/WordPress/gutenberg/blob/4d1aa40f02c6a46e08b1496f74c66a059611b0a2/packages/base-styles/_colors.scss#L12), which meets accessibility requirements for large text, but not for normal text. For normal text, we should use a color at least darker than [$gray-700](https://github.com/WordPress/gutenberg/blob/4d1aa40f02c6a46e08b1496f74c66a059611b0a2/packages/base-styles/_colors.scss#L11).

Or, we could just leave it as the normal text color.

| Before ($gray-600, #949494) | After ($gray-700, #757575) |
|--------|--------|
| image| image|
| image| image|

### DIFF
--- a/src/plugin-styles.scss
+++ b/src/plugin-styles.scss
@@ -27,12 +27,8 @@ $modal-footer-height: 70px;
 
 	&__plugin-sidebar {
 		&__about-button {
-			color: $gray-600;
+			color: $gray-700;
 			margin-top: 3rem;
-		}
-
-		&__about-button svg {
-			fill: $gray-600;
 		}
 	}
 }


src/plugin-styles.scss
```diff
@@ -27,12 +27,8 @@ $modal-footer-height: 70px;
...
 	&__plugin-sidebar {
 		&__about-button {
-			color: $gray-600;
+			color: $gray-700;
 			margin-top: 3rem;
 		}
-
-		&__about-button svg {
-			fill: $gray-600;
-		}
```
t-hamano, line 36: The `svg` element in the button component is applied with `currentColor` by default.

https://github.com/WordPress/gutenberg/blob/4d1aa40f02c6a46e08b1496f74c66a059611b0a2/packages/components/src/button/style.scss#L345

mikachan (approved):
Thanks @t-hamano! I think using `$gray-700` is a good change and seems like a suitable contrast for this font size.

Issue #287, opened 2023-03-23
Add block pattern management


justintadlock:
@annezazu reached out to see what I thought of integrating my [Block Pattern Builder](https://github.com/justintadlock/block-pattern-builder) plugin into Create Block Theme. One of the pain points being that the pattern creation process was a bit too manual for non-technical users. 

BPB was originally built as a proof of concept of a pattern manager that I'd envisioned core WP eventually handling. The current release is not 100% up to date with the current Patterns API features, but I've been working locally to update it.

I'm happy to start on the process of a PR if this is desirable for the plugin.

If so, I imagine one thing that would need to change is prefixes. Names/prefixes:

- Patterns: `bpb/pattern-name`
- CPT: `bpb_pattern`
- Taxonomy: `bpb_pattern_category`
- Role capabilities: `bpb_cap_name`

Any thoughts on naming/prefixing for those things?

----

Potential features:

- [x] Allow user-created patterns to be exported with a theme
- [ ] Allow deleting theme patterns via the Editor; or allow theme patterns to be excluded from the theme export

matiasbenedetto:
I haven't used [this plugin before](https://github.com/justintadlock/block-pattern-builder), and I'm curious about the differences with [this recently launched plugin](https://wpengine.com/builders/pattern-manager-download/) for managing patterns.

Will integrating your plugin add any new features different from the other plugin?

justintadlock:
Block Pattern Builder was primarily built for users and hasn't been updated in a bit for newer Pattern API features. In and of itself, it wouldn't add any new features.

Pattern Manager was built primarily for developers as part of their workflow. It'd be an obvious choice to integrate into Create Block Theme.

annezazu:
I'll ask the folks behind the other plugin too about possible integration in case there are a few ways forward here. Ultimately, just want to see what we can do in the short term to make patterns easier and easier to create and manage.

ndiego:
Now that 6.3 includes the ability for users to create both synced and standard patterns from within the Editor, it would be great to revisit this issue. Creating standard patterns in the Editor and then exporting them directly to the theme, much like templates and template parts, would be amazing. 

cc @mikachan 😉

annezazu:
[...]

PR #675, opened 2024-06-27
Try: Add synced patterns to theme on save

mikachan:
This PR explores the most minimal way we can introduce some pattern management into the plugin by allowing custom patterns to be saved to the theme filesystem on save, allowing them to be exported with the theme.

Part of https://github.com/WordPress/create-block-theme/issues/287.

image

## Testing Instructions
1. Create a new synced pattern in the Editor (also try adding categories)
2. Go to the Editor, and in the Create Block Theme menu, save the theme changes with the "Save My Patterns" option checked

### DIFF
--- a/includes/create-theme/theme-patterns.php
+++ b/includes/create-theme/theme-patterns.php
@@ -24,13 +24,13 @@ class CBT_Theme_Patterns {
 	public static function pattern_from_wp_block( $pattern_post ) {
 		$pattern          = new stdClass();
 		$pattern->id      = $pattern_post->ID;
+		$pattern->title   = $pattern_post->post_title;
 		$pattern->name    = sanitize_title_with_dashes( $pattern_post->post_title );
-		$theme_slug       = wp_get_theme()->get( 'TextDomain' );
-		$pattern->slug    = $theme_slug . '/' . $pattern->name;
+		$pattern->slug    = wp_get_theme()->get( 'TextDomain' ) . '/' . $pattern->name;
 		$pattern->content = (
 		'<?php
 /**
- * Title: ' . $pattern_post->post_title . '
+ * Title: ' . $pattern->title . '
  * Slug: ' . $pattern->slug . '
  * Categories: hidden
  * Inserter: no
--- a/includes/create-theme/theme-patterns.php
+++ b/includes/create-theme/theme-patterns.php
@@ -29,7 +29,8 @@ class CBT_Theme_Patterns {
 		$pattern->slug         = wp_get_theme()->get( 'TextDomain' ) . '/' . $pattern->name;
 		$pattern_category_list = get_the_terms( $pattern->id, 'wp_pattern_category' );
 		$pattern->categories   = ! empty( $pattern_category_list ) ? join( ', ', wp_list_pluck( $pattern_category_list, 'name' ) ) : '';
-		$pattern->content      = (
+		// $pattern->sync_status  = get_post_meta( $pattern->id, 'wp_pattern_sync_status', true );
+		$pattern->content = (
 		'<?php
 /**
  * Title: ' . $pattern->title . '
--- a/includes/create-theme/theme-patterns.php
+++ b/includes/create-theme/theme-patterns.php
@@ -144,6 +144,12 @@ class CBT_Theme_Patterns {
 				);
 
 				self::replace_local_pattern_references( $pattern );
+
+				// If it's a synced pattern then remove it from the database.
+				// This is to ensure that these patterns are loaded from the theme.
+				if ( 'unsynced' !== $pattern->sync_status ) {
+					wp_delete_post( $pattern->id, true );
+				}
 			}
 		}
 	}
--- a/includes/create-theme/theme-patterns.php
+++ b/includes/create-theme/theme-patterns.php
@@ -21,6 +21,28 @@ class CBT_Theme_Patterns {
 		);
 	}
 
+	public static function pattern_from_wp_block( $post ) {
+		$pattern_name    = sanitize_title_with_dashes( $post->post_title );
+		$theme_slug      = $new_slug ? $new_slug : wp_get_theme()->get( 'TextDomain' );
+		$pattern_slug    = $theme_slug . '/' . $pattern_name;
+		$pattern_content = (
+		'<?php
+/**
+ * Title: ' . $post->post_title . '
+ * Slug: ' . $pattern_slug . '
+ * Categories: hidden
+ * Inserter: no
+ */
+?>
+' . $post->post_content
+		);
+		return array(
+			'name'    => $pattern_name,
+			'slug'    => $pattern_slug,
+			'content' => $pattern_content,
+		);
+	}
+
 	public static function escape_alt_for_pattern( $html ) {
 		if ( empty( $html ) ) {
 			return $html;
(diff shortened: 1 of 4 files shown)


eirichmond:
@mikachan Great job on this! Initial testing works well. I know it's a work in progress, but I wondered why the Inserter was set to 'no' in the pattern header. Once saved and exported, the pattern still doesn't come from the theme/patterns folder; it remains in the database. I guess there is logic behind that. For me, if I'm developing a theme, I wouldn't want anything left in the database after saving and exporting the patterns. I think there might be some thinking behind synced patterns, but as a developer creating patterns as part of the theme, I'd want them not to be restricted and synced. I'd prefer the theme to be distributed to users who could then decide to generate their own synced patterns from the ones provided by the theme, if that makes sense.

Also, I think the heredoc approach might be more readable for generating the header. This allows you to include the variables directly within the string without needing to concatenate multiple parts.

Again, I know this is a work in progress, so I guess any image-related details (e.g., downloaded to the theme assets/images and changing the reference to get_theme_file_uri('xxxx'), etc.) haven't been addressed yet. I'd love to contribute to this with my own pull request if it helps. Great job 🥳

eirichmond:
In hindsight, I guess the reasoning for not removing them from the database is to preserve synced patterns that have already been placed throughout the site. It would be really useful if there was a flag saved within postmeta indicating whether a pattern was already synced. This could then be used to avoid removing the pattern when the theme is saved and the patterns are exported.

mikachan:
> It would be really useful if there was a flag saved within postmeta indicating whether a pattern was already synced.

It looks like the sync status is saved in postmeta under `wp_pattern_sync_status`, and you're able to access it using:

`get_post_meta( $pattern->id, 'wp_pattern_sync_status', true );`

includes/create-theme/theme-patterns.php
```diff
@@ -21,6 +21,30 @@ public static function pattern_from_template( $template, $new_slug = null ) {
...
 
+	public static function pattern_from_wp_block( $pattern_post ) {
+		$pattern               = new stdClass();
+		$pattern->id           = $pattern_post->ID;
+		$pattern->title        = $pattern_post->post_title;
+		$pattern->name         = sanitize_title_with_dashes( $pattern_post->post_title );
+		$pattern->slug         = wp_get_theme()->get( 'TextDomain' ) . '/' . $pattern->name;
+		$pattern_category_list = get_the_terms( $pattern->id, 'wp_pattern_category' );
+		$pattern->categories   = ! empty( $pattern_category_list ) ? join( ', ', wp_list_pluck( $pattern_category_list, 'name' ) ) : '';
+		// $pattern->sync_status  = get_post_meta( $pattern->id, 'wp_pattern_sync_status', true );
```
mikachan: I'm not sure this belongs here but I've left it here commented out as a reference for now.

mikachan:
I've updated this PR with the changes from @eirichmond's PR (https://github.com/WordPress/create-block-theme/pull/693), refactored a couple of things and added some error handling. I think this is ready for review 🤞

includes/create-theme/theme-patterns.php
```diff
@@ -47,4 +71,111 @@ public static function create_pattern_link( $attributes ) {
...
+								),
+								$pattern->name
+							)
+						);
+					}
+
+					// Create the pattern file.
+					$pattern_file = $patterns_dir . 'pattern-' . $pattern->name . '.php';
+					file_put_contents(
+						$patterns_dir . DIRECTORY_SEPARATOR . 'pattern-' . $pattern->name . '.php',
```
jffng: Do you think the `pattern-` prefix is necessary?

```suggestion
						$patterns_dir . DIRECTORY_SEPARATOR . 'pattern-' . $pattern->name . '.php',
```

includes/create-theme/theme-patterns.php
```diff
@@ -47,4 +71,111 @@ public static function create_pattern_link( $attributes ) {
...
+						$pattern->content
+					);
+
+					self::replace_local_pattern_references( $pattern );
+
+					// Remove it from the database to ensure that these patterns are loaded from the theme.
+					wp_delete_post( $pattern->id, true );
```
jffng, line 198: If I create a new pattern and save it using CBT, it reloads the editor but the post has been deleted. With Gutenberg active I at least get a better error response: 

**Without Gutenberg**
https://github.com/user-attachments/assets/79a8cd6a-6f0a-4a5e-9339-228b3b8da0e3

**Gutenberg 19.0.0**
 Screenshot 2024-08-14 at 2 49 49 PM

I wonder if we can handle this redirection better, maybe return the user back to the patterns view?

jffng (commented):
This is testing pretty well for me, nice work! I was able to save custom patterns and it exported with the theme.

includes/create-theme/theme-patterns.php
colorful-tones: This threw me off. I could not find my patterns and thought it was not working ...until I saw this hint:

> Do you think the pattern- prefix is necessary?

includes/create-theme/theme-patterns.php
colorful-tones: I received same error with latest Gutenberg trunk. When I disabled Gutenberg I got WSOD with HTML `<title>null ‹ Template ‹ Gutenberg Trunk ‹ Editor — WordPress</title>` 🤔

includes/create-theme/theme-patterns.php
jffng: Yeah same! Especially if the theme already has lots of patterns it was hard to spot.

jffng: I think I expected the file name to be the kebab-cased pattern name.

mikachan: I'm not sure why I added the prefix, haha 🙈 I'll remove it for now, I can't see what value it's adding. If I remember I'll let you know!

mikachan: Removed the prefix here: https://github.com/WordPress/create-block-theme/pull/675/commits/f017b9485cfddac0bc0f0f5cf2f5312304f42c4e.

includes/create-theme/theme-patterns.php
mikachan: > I wonder if we can handle this redirection better, maybe return the user back to the patterns view?

Yes, nice idea, I've done this here: https://github.com/WordPress/create-block-theme/pull/675/commits/6de813f38bcce5951cf35ff6ac33d36cda4a6318.

Thanks both!

mikachan:
Thanks for the feedback @jffng and @colorful-tones! This should be ready for another review.

colorful-tones:
I'm confused about something in my testing. I created a synced pattern, then assigned it to a template, and then went into CBT and checked _only_ the 'Save Synced Patterns' thinking that the template would also be saved, because of this overall statement: "Any synced patterns created in the Editor will be moved to the theme. Note that this will delete all synced patterns from the Editor and any references in templates will be made relative to the theme." 

However, after hitting the 'Save Changes' from CBT I only see my new pattern in the patterns directory and the template has no changes at all, which makes me think the statement is a bit misleading. I imagine I would want to check the "Save Template Changes" box too, but the "any references in templates will be made relative to the theme." next to the "Save Synced Patterns" implied to me that my template changes would be saved too. 🤔 

Also, after saving patterns to the theme when I go back to the template it states: "Block has been deleted or is unavailable."

Screenshot 2024-08-15 at 10 44 34 AM

Screenshot 2024-08-15 at 10 49 30 AM

mikachan:
Thanks @colorful-tones! I've pushed an update that will update any templates that reference the pattern and save those changes to the theme, regardless of whether the "Save Template Changes" option is checked. I think that matches up with the expectations in the "Save Sycned Patterns" description. This should also fix the "Block has been deleted or is unavailable." issue.

colorful-tones:
@mikachan I did a bunch of testing and everything is working. Nice work! 

I think this functionality will help folks with creating and saving their patterns.

colorful-tones (approved):
LGTM! I think this functionality will be super helpful.

mikachan:
Thanks so much for reviewing, @colorful-tones!

PR #678, opened 2024-07-01
Persist font settings when cloning a theme

### LINKED ISSUE #677: Installed font families are not downloaded to the theme folder when the option 'Clone theme' is used.
## What?
Installed font families are not downloaded to the theme folder when the option 'Clone theme'  and after that changes are saved to the theme using Save Changes to the Theme` button.

This issue was found working on this: https://github.com/Automattic/themes/pull/7865

**Step by step:**
1. Install a font face
2. Clone the theme using these buttons: `Create Theme` -> `Clone Theme`.
3. Save using `Save Changes to the Theme` button.

### Problem explained technically:
When `Clone Theme` is used and there are custom fonts activated in the global styles, these fonts are added using an absolute link in the new `theme.json` created. When the changes are saved using `Save Changes to the Theme` those fonts are not processed (they are not downloaded and the  link is not transformed to be relative to theme).

## Expected result
The `Clone Theme` option probably should have the same options so the user can control if the fonts are processed when the theme is cloned.

image

mikachan:
This PR should apply the "Save Fonts" save option when cloning a theme via Create Theme -> Clone Theme. This means that any font settings and files are copied over to the new cloned theme, including updating the file asset paths.

Fixes https://github.com/WordPress/create-block-theme/issues/677.

To test:

1. Install and activate some custom fonts using the Font Library.
2. Go to the CBT menu, Create Theme -> Clone Theme.
3. Name your theme.
4. Check the theme.json of the newly cloned theme: ensure that the `fontFace.src` paths are relative to the cloned theme. i.e. the paths start with `file:.` and not an absolute URL.

### DIFF
--- a/includes/create-theme/theme-create.php
+++ b/includes/create-theme/theme-create.php
@@ -25,6 +25,9 @@ class CBT_Theme_Create {
 
 		wp_mkdir_p( $new_theme_path );
 
+		// Persist font settings for cloned theme.
+		CBT_Theme_Fonts::persist_font_settings();
+
 		// Copy theme files.
 		$template_options = array(
 			'localizeText'   => false,


matiasbenedetto (commented):
I think this could work but we would be removing the possibility of having linked fonts in a cloned theme. All the fonts will be downloaded and made local as we don't have that possibility in WordPress core yet I think it's OK and nobody will expect the opposite. I will try in action to see if it solves the problem.

matiasbenedetto (approved):
This works well fixing https://github.com/WordPress/create-block-theme/issues/677
Let's merge it.

mikachan:
Thanks @matiasbenedetto! 

> I think this could work but we would be removing the possibility of having linked fonts in a cloned theme.

Ah yes, we can address this later once Core supports it, like you say 👍

PR #681, opened 2024-07-03
Sanitize DOS (Windows) new line style on readme.txt

### LINKED ISSUE #680: The Theme Check plugin sometimes returns "Both DOS and UNIX style line endings" error
In version 2.3.0, sometimes the Theme Check plugin returns the following error:

> REQUIRED: Both DOS and UNIX style line endings were found in the file readme.txt. This causes a problem with SVN repositories and must be corrected before the theme can be accepted. Please change the file to use only one style of line endings.

This might be related to https://github.com/WordPress/create-block-theme/pull/662.

We might be able to fix this by using regex to normalise all new lines in the open text fields for the theme meta data.

To reproduce:

- Update the font credits and manually add image credits using the CBT UI
- Export the theme zip, install it on a local site, and run the Theme Check plugin

matiasbenedetto:
## What?
Replaces DOS new line style on readme.txt

## Why? 
This type of character causes problems on WordPress.org theme directory.

Reference from a theme review:
> Both DOS and UNIX style line endings were found in the file readme.txt. This causes a problem with SVN repositories and must be corrected before the theme can be accepted.

Fixes https://github.com/WordPress/create-block-theme/issues/680

## How?
The "\r\n" new line characters are replaced by the unix style new line character "\n" when creating or updating a readme.txt text.

### DIFF
--- a/includes/create-theme/theme-readme.php
+++ b/includes/create-theme/theme-readme.php
@@ -97,6 +97,9 @@ License URI: {$license_uri}
 		// Adds the Images section
 		$readme_content = self::add_or_update_section( 'Images', $image_credits, $readme_content );
 
+		// Sanitize the readme content
+		$readme_content = self::sanitize( $readme_content );
+
 		return $readme_content;
 	}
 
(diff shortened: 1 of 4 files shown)


mikachan (approved):
Thanks for addressing this! Code looks good to me and it's testing well. I think we can bring this in 🎉

PR #693, opened 2024-07-27
Export patterns

eirichmond:
This is a hybrid of @mikachan work, just removed inserter from the header of patterns the they are loaded from the theme, adjusted to heredoc for readability and removed unsynced patterns from the database after saving to the theme.

### DIFF
--- a/includes/create-theme/theme-patterns.php
+++ b/includes/create-theme/theme-patterns.php
@@ -4,21 +4,42 @@ class CBT_Theme_Patterns {
 	public static function pattern_from_template( $template, $new_slug = null ) {
 		$theme_slug      = $new_slug ? $new_slug : wp_get_theme()->get( 'TextDomain' );
 		$pattern_slug    = $theme_slug . '/' . $template->slug;
-		$pattern_content = (
-		'<?php
-/**
- * Title: ' . $template->slug . '
- * Slug: ' . $pattern_slug . '
- * Categories: hidden
- * Inserter: no
- */
-?>
-' . $template->content
-		);
+		$pattern_content = <<<PHP
+		<?php
+		/**
+		 * Title: {$pattern->title}
+		 * Slug: {$pattern->slug}
+		 * Categories: {$pattern->categories}
+		 */
+		?>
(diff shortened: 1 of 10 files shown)


includes/create-theme/theme-patterns.php
```diff
@@ -47,4 +68,91 @@ public static function create_pattern_link( $attributes ) {
...
+
+				self::replace_local_pattern_references( $pattern );
+
+				/**
+				 * if the pattern is not synced then remove if from
+				 * database so patterns are loaded from the theme
+				 * and not the database
+				 */
+
+				if ( stripos( $pattern->content, 'pattern-overrides' ) !== true ) {
```
mikachan, line 151: This looks like a neat way to check if the pattern is not synced. I already mentioned [here](https://github.com/WordPress/create-block-theme/pull/675#issuecomment-2260911432), but I also found `wp_pattern_sync_status` in postmeta that we could use as well.

includes/create-theme/theme-patterns.php
```diff
@@ -4,23 +4,44 @@ class CBT_Theme_Patterns {
...
-		);
+		$pattern_content = <<<PHP
+		<?php
+		/**
+		 * Title: {$pattern->title}
+		 * Slug: {$pattern->slug}
+		 * Categories: {$pattern->categories}
+		 */
+		?>
+		{$pattern_post->post_content}
```
mikachan, line 15:
```suggestion
		 * Title: {$template->title}
		 * Slug: {$pattern_slug}
		 * Categories: {$template->categories}
		 */
		?>
		{$template->post_content}
```

I think this should use `$template` rather than `$pattern_post` - I'm guessing this was accidentally copied over from `pattern_from_wp_block`. `$template->categories` might need to be created, too.

mikachan (commented):
Thanks so much for working on this, @eirichmond!

It looks like this may need a rebase or a fresh merge from https://github.com/WordPress/create-block-theme/pull/675, I've just merged trunk into that PR so it's up to date. At the moment, it's hard to see which changes are being brought in. However, this is looking great so far 🎉

PR #701, opened 2024-08-15
Update broken link in CONTRIBUTOR.md

colorful-tones:
I was getting this when clicking it
Screenshot 2024-08-15 at 2 22 02 PM


t-hamano (approved):
The currently available view id is 4 (https://github.com/orgs/WordPress/projects/188/views/4), but maybe a view created in the past was deleted and is no longer accessible.

In any case, I don't think we need to rely on a specific view id.

PR #704, opened 2024-08-16
Bump minimum required WordPress version to 6.6

t-hamano:
Bump the minimum required WordPress version to 6.6.

There is a convention in the Gutenberg project to support one version older than the latest WordPress release. Regarding the Create Block Theme plugin, I don't think there is a clear backward compatibility policy yet, but if it's been a few weeks since the latest WP version was released, I think it's okay to bump up the minimum WordPress version.

Please let me know your thoughts.

In this PR, l also moved the following APIs from the `@wordpress/edit-site` package to the `@wordpress/editor` package to prevent browser warning errors:

- `PluginSidebar`
- `PluginSidebarMoreMenuItem`


matiasbenedetto (approved):
LGTM

PR #705, opened 2024-08-16
Don't enable sidebar UI in classic theme

t-hamano:
In WordPress 6.6, Patterns was added as a submenu in the Appearance menu for classic themes:

image

This means that classic themes have access to some screens in the site editor.

I noticed that when I opened the Pattern Editor in a classic theme, the Plugins sidebar is available:

image

This plugin is for block themes and should not be used in classic themes.


matiasbenedetto (approved):
LGTM, thanks.

Issue #355, opened 2023-05-09
Create blank theme: Update the minimum WordPress version


carolinan:
When you select the Create blank theme option, the minimum required WordPress version is set to 5.8.
The theme cannot be activated on WordPress 5.8 because it is missing an index.php file.

The requirement to include the index.php file was removed in WordPress 6.0, so I recommend increasing it to at least that version.

matiasbenedetto:
thanks for reporting this. This should fix it: https://github.com/WordPress/create-block-theme/pull/356

PR #715, opened 2024-09-04
Adds minimum WordPress version to theme metadata

jffng:
**Description**
This PR adds a field to the metadata to allow users to update the required minimum WordPress version in the theme's metadata.

**Background**
In reviewing some themes recently (https://github.com/Automattic/themes/pull/8058) (https://github.com/Automattic/themes/pull/8067), I've noticed the required minimum version is empty. 

I think we should allow the user to be able to edit this via the Metadata editor. 

Since this metadata is a requirement, the PR also supplies a minimum version (5.9, when block themes were introduced) if none exists. I'm not sure about this part.


includes/create-theme/theme-styles.php
```diff
@@ -26,15 +26,18 @@ public static function update_style_css( $style_css, $theme ) {
...
-		$requires_php  = $current_theme->get( 'RequiresPHP' );
-		$text_domain   = $theme['slug'];
-		$template      = $current_theme->get( 'Template' ) ? "\n" . 'Template: ' . $current_theme->get( 'Template' ) : '';
-		$license       = $style_data['License'] ? $style_data['License'] : 'GNU General Public License v2 or later';
-		$license_uri   = $style_data['LicenseURI'] ? $style_data['LicenseURI'] : 'http://www.gnu.org/licenses/gpl-2.0.html';
-		$tags          = CBT_Theme_Tags::theme_tags_list( $theme );
-		$css_contents  = $css_contents ? "\n\n" . $css_contents : '';
-		$copyright     = '';
+		if ( ! $wp_min ) {
+			$wp_min = '5.9';
```
mikachan: It looks like we've hard-coded `6.0` elsewhere, maybe we should go with that rather than `5.9`?

includes/create-theme/theme-styles.php
```diff
@@ -26,15 +26,18 @@ public static function update_style_css( $style_css, $theme ) {
...
-		$version       = $theme['version'];
-		$requires_php  = $current_theme->get( 'RequiresPHP' );
-		$text_domain   = $theme['slug'];
-		$template      = $current_theme->get( 'Template' ) ? "\n" . 'Template: ' . $current_theme->get( 'Template' ) : '';
-		$license       = $style_data['License'] ? $style_data['License'] : 'GNU General Public License v2 or later';
-		$license_uri   = $style_data['LicenseURI'] ? $style_data['LicenseURI'] : 'http://www.gnu.org/licenses/gpl-2.0.html';
-		$tags          = CBT_Theme_Tags::theme_tags_list( $theme );
-		$css_contents  = $css_contents ? "\n\n" . $css_contents : '';
-		$copyright     = '';
+		if ( ! $wp_min ) {
```
mikachan: We should probably use this same logic in `build_style_css` below, which builds the style.css for child themes.

We're also hard-coding `6.0` for the "Requires at least" value in the readme files on [line 70 of theme-readme.php](https://github.com/WordPress/create-block-theme/blob/trunk/includes/create-theme/theme-readme.php#L70), so maybe we could use the same logic here too.

mikachan (commented):
Nice, I think this is a good idea to help prevent missing minimum versions.

matiasbenedetto (commented):
I think this field could be a good candidate for a 'select' type component instead of a plain 'input' one. The select could feature the version '5.9', '6.0', '6.1'.

image

As a probably not-needed addition, each option could have some comment/documentation like "Minimum version for block themes" or something like that.

richtabor:
Is there a case where the minimum version would be the version the theme was exported with?

matiasbenedetto (commented):
It doesn't seem to be updating the data in the theme as expected. In this screencast I tried updating theme description (working as expected) and WordPress minimum version (not updating).

https://github.com/user-attachments/assets/7daa4c0b-c9ed-45ca-bd45-91954847eb35

matiasbenedetto:
> Is there a case where the minimum version would be the version the theme was exported with?

That's not always true. A theme could be a very basic template that doesn't require the latest version of Wordpress which is commonly what a theme creator would use.

jffng:
Thanks for the reviews and feedback.

> I think this field could be a good candidate for a 'select' type component instead of a plain 'input' one.

I updated the control to be a Select control instead of a text input. An issue with this is that if the theme sets a version not provided by the predetermined options, it won't be reflected in the metadata editor. 

> Is there a case where the minimum version would be the version the theme was exported with?

It's possible, but do we think that's a better default if no required version is supplied?

> It doesn't seem to be updating the data in the theme as expected.

Whoops, I wasn't properly sanitizing the field in the REST API request. Fixed in [b1d1558](https://github.com/WordPress/create-block-theme/pull/715/commits/b1d1558edea9a222483ba08bb7953c9565ab723c)

includes/create-theme/theme-styles.php
jffng: I updated to `6.0`, but it made me wonder why we're going with `6.0` vs `5.9` since that's when block themes were introduced.

richtabor:
> That's not always true. A theme could be a very basic template that doesn't require the latest version of Wordpress which is commonly what a theme creator would use.

Problem is, you don't really know. Picking a version is basically guessing. :)

includes/create-theme/theme-styles.php
mikachan: I wasn't sure either - looking at when we added `6.0` to the readme class, I found this [related issue](https://github.com/WordPress/create-block-theme/issues/355) where it looks like we decided based on this:

> The requirement to include the index.php file was removed in WordPress 6.0, so I recommend increasing it to at least that version.

In that case, I think we're fine to stick with `6.0`.

includes/create-theme/theme-styles.php
```diff
@@ -25,7 +25,8 @@ public static function update_style_css( $style_css, $theme ) {
 		$author        = stripslashes( $theme['author'] );
 		$author_uri    = $theme['author_uri'];
 		$wp_version    = CBT_Theme_Utils::get_current_wordpress_version();
-		$wp_min        = $current_theme->get( 'RequiresWP' );
+		$wp_min        = $theme['requires_wp'];
+		$wp_min        = $wp_min ? $wp_min : '6.0';
```
creativecoder: Simplify?
```suggestion
		$wp_min        = isset( $theme['requires_wp'] ) && $theme['requires_wp'] ? $theme['requires_wp'] : '6.0';
```

src/editor-sidebar/metadata-editor-modal.js
```diff
@@ -39,12 +40,25 @@ const ALLOWED_SCREENSHOT_MEDIA_TYPES = [
...
+	'5.9',
+	'6.0',
+	'6.1',
+	'6.2',
+	'6.3',
+	'6.4',
+	'6.5',
+	'6.6',
+	'6.7',
+];
```
creativecoder: This seems tedious to maintain. How about adding `global $wp_version` from PHP into a JS global and using that to generate a range of versions?

jffng: @creativecoder do you mean something like this? https://github.com/WordPress/create-block-theme/pull/715/commits/f3d352a0d0fd72fa9502368fda8655c0b3d94f30

creativecoder: Yes, that's in line with what I was thinking!

vcanales:
> do we think that's a better default if no required version is supplied?

I think it would make sense because it would reflect the features available to the theme builder at the time of export.

includes/create-theme/theme-styles.php
```diff
@@ -25,7 +25,7 @@ public static function update_style_css( $style_css, $theme ) {
 		$author        = stripslashes( $theme['author'] );
 		$author_uri    = $theme['author_uri'];
 		$wp_version    = CBT_Theme_Utils::get_current_wordpress_version();
-		$wp_min        = $current_theme->get( 'RequiresWP' );
+		$wp_min        = isset( $theme['requires_wp'] ) && $theme['requires_wp'] ? $theme['requires_wp'] : '6.0';
```
matiasbenedetto: If  $theme['requires_wp'] is not defined we could defaut to the wordpress version currently in use instead to an arbitrary one:


```suggestion
		$wp_min        = $theme['requires_wp'] ?: CBT_Theme_Utils::get_current_wordpress_version();
```

matiasbenedetto (commented):
When you update the minimum version the `style.css` file is updated but the `readme.txt` file not.

The minimum version of the blank themes created is hardcoded. Probably we should set something dynamic.

includes/create-theme/theme-styles.php
jffng: Done here https://github.com/WordPress/create-block-theme/pull/715/commits/f983c4e2dc630fe81a7b36fb25eee6a506b8533d

jffng:
Thanks @matiasbenedetto for the review.

> When you update the minimum version the style.css file is updated but the readme.txt file not.

Fixed here: https://github.com/WordPress/create-block-theme/pull/715/commits/416b149e4c243ebc2cf4b7653b86b5a12c3e7a03

> The minimum version of the blank themes created is hardcoded. Probably we should set something dynamic.

Done here: https://github.com/WordPress/create-block-theme/pull/715/commits/8e25b5b87778f2a6747078534abb533fb1ad5702

matiasbenedetto:
> >The minimum version of the blank themes created is hardcoded. Probably we should set something dynamic.
> Done here: https://github.com/WordPress/create-block-theme/commit/8e25b5b87778f2a6747078534abb533fb1ad5702

The default version is empty now when creating a blank theme.

https://github.com/user-attachments/assets/e94ee839-0e95-4342-bdea-4eae1cae5ae5

jffng:
> The default version is empty now when creating a blank theme.

Sorry for not testing that before, I fixed it here: [b3b3386](https://github.com/WordPress/create-block-theme/pull/715/commits/b3b33861b5a1621dde5bb278cca3ba99e62f804b). I assumed I could check for the existence of the parameter, but the sanitization function ensures that `requires_wp` will always exist and falls back to an empty string.

I also added the minimum version selector to the additional metadata panel when creating a new theme: 2f9225f097fd0cdf1ed3dc6159fd6048d6d3d257

Here's what I tested:
- Updating current theme's metadata
- Creating a blank theme
- Create a clone theme
- Creating a child theme

Am I missing any other flows?

mikachan (approved):
This is working well in my testing 🎉 

I've tested creating themes from the Appearance page, which defaults to using the active WP version in each case:
- Blank theme ✅ 
- Clone ✅ 
- Child ✅ 

I also tested the options available from the Editor, where I was able to set a custom WP version or it would default to the active version:
- Editing active theme's metadata ✅ 
- Blank theme ✅ 
- Clone ✅ 
- Child ✅ 

I don't believe there are any other flows and I think this is ready to bring in, so I'm leaving a ✔️ , but it'd be great if @matiasbenedetto has a chance to review before landing too.

PR #717, opened 2024-09-09
Update node dependencies

mikachan:
Updates node dependencies to their latest versions.


creativecoder (approved):
This seems to be working fine for me.

I ran through all of the package.json scripts without error, and the plugin build in this branch looks to be working as expected.

PR #718, opened 2024-09-09
Remove "Categories: hidden" from default pattern header

mikachan:
This removes "Categories: hidden" from the `pattern_from_template` function, as the `hidden` category does not exist and “categories” is not a required property for registering blocks.

Props to @carolinan for mentioning this [here](https://wordpress.slack.com/archives/C015GUFFC00/p1725864890894449).


jffng (approved):

carolinan:
Thank you

PR #719, opened 2024-09-09
Process inner html of blocks when escaping text content

### LINKED ISSUE #573: The HTML within a paragraph gets converted into a translation string on theme save
I ended up with:

```
<!-- wp:paragraph {"style":{"typography":{"fontStyle":"normal","fontWeight":"500"}}} -->
<p style="font-style:normal;font-weight:500">
<?php echo __('<a href="mailto:email@example.com" data-type="mailto" data-id="mailto:email@example.com">email@example.com</a><br>Tel.: +48 000 000 000', 'mytheme');?>
</p>
<!-- /wp:paragraph -->
```

while the more desirable would be:

```
<!-- wp:paragraph {"style":{"typography":{"fontStyle":"normal","fontWeight":"500"}}} -->
<p style="font-style:normal;font-weight:500">
<a href="mailto:email@example.com" data-type="mailto" data-id="mailto:email@example.com"><?php echo __('email@example.com', 'mytheme'); ?></a><br><?php echo __('Tel.: +48 000 000 000', 'mytheme');?>
</p>
<!-- /wp:paragraph -->
```

I am aware it may be difficult to implement, but I'm just reporting the situations when I was forced to edit the generated template or pattern manually to make it deliverable.

_I'm not even sure that every paragraph should be converted into a translation, but I have no idea how we could distinguish between the ones we want and the ones we don't._

### LINKED ISSUE #682: HTML inside text blocks renders as text.

### LINKED ISSUE #691: Bug: Something isn't right when selecting all options in Save Changes to Theme
When I select all the options in Save Changes to theme, this is happening:

Captura de ecrã 2024-07-19, às 10 29 21

The links that have a `#` are also having an issue with weird text showing up on the front end.

There's also this warning from Theme Check plugin:

```
WARNING: Found ="<?php esc_html_e in patterns/front-page.php. Use esc_attr_e() inside HTML attributes, and esc_url() for link attributes. A manual review is needed.

Line 36: <figure class='wp-block-image size-full'><img src='<?php echo esc_url( get_stylesheet_directory_uri() ); ?>/assets/images/mood_1-min.jpg' alt='<?php esc_html_e('', 'cottage');?>'/></figure>
```

jffng:
**What**
This PR enables HTML markup inside block content to be processed while still allowing the text content to be safely escaped.

Before | After
--- | ---
<video src="https://github.com/user-attachments/assets/a9c1a925-5622-43cc-ace3-5a2942f5b936"> | <video src="https://github.com/user-attachments/assets/367f866e-3f9e-4cc5-b781-0b336e19dc5f">

**How**
Updates the `CBT_Theme_Locale::escape_text_content` function to handle nested HTML markup. 

Uses `WP_HTML_Tag_Processor` to process every token of the blocks content, and generates a string that can be formatted and then translated. 

**Why**
Fixes #573 #682 and #691.  

**To test**
- Add some markup to the inside of a block in a template.
- Save the changes to the theme.
- Verify the template has been patternized correctly, and the template appears as expected.


includes/create-theme/theme-locale.php
```diff
@@ -28,6 +28,72 @@ private static function escape_text_content( $string ) {
...
+			$is_tag_closer = $p->is_tag_closer();
+
+			if ( '#tag' === $token_type ) {
+				// Add a placeholder for the token.
+				$text .= '%s';
+				if ( $is_tag_closer ) {
+					$tokens[] = "</{$token_name}>";
+				} else {
+					// Depending on the HTML tag, we may need to process attributes so they are correctly added to the placeholder.
+					switch ( $token_name ) {
```
jffng: One thing I would like feedback on is whether this is too fragile, complicated, or tedious an approach to maintain.

I plan to add a few more unit tests.

madhusudhand (commented):
It works! But I encountered following scenario where multiple individually translated strings now becomes the single translation.

I guess this is because the template in the editor comes with a single string (after translations) and now it is seen as a single string. 

I don't think this should be blocking this change. It is to be addressed separately by reading original pattern and current template together may be?

**Original**
```
<h2 class="wp-block-heading" id="botswana-new-zealand-south-korea-japan-madagascar" style="font-size:80px;font-style:normal;font-weight:900;line-height:1.1;text-transform:uppercase"><a href=""><?php echo esc_html__( 'ITALY', 'adventurer' ); ?></a><br><a href=""><?php echo esc_html__( 'COSTA RICA', 'adventurer' ); ?></a><br><a href=""><?php echo esc_html__( 'CANADA', 'adventurer' ); ?></a><br><a href=""><?php echo esc_html__( 'LAOS', 'adventurer' ); ?></a><br><a href=""><?php echo esc_html__( 'TURKEY', 'adventurer' ); ?></a></h2>
```

**After CBT localization**
```
<h2 class="wp-block-heading" id="botswana-new-zealand-south-korea-japan-madagascar" style="font-size:80px;font-style:normal;font-weight:900;line-height:1.1;text-transform:uppercase"><?php /* Translators: %s are html tags */ echo sprintf( esc_html__( '%sITALY%s%s%sCOSTA RICA%s%s%sCANADA%s%s%sLAOS%s%s%sTURKEY%s', 'adventurer' ), '<a href="">', '</a>', '<br>', '<a href="">', '</a>', '<br>', '<a href="">', '</a>', '<br>', '<a href="">', '</a>', '<br>', '<a href="">', '</a>' ); ?></h2>
```

includes/create-theme/theme-locale.php
mikachan: I think we're good to rely on the output of `get_token_name()`, so based on that, I believe this is a fine way to parse these tags. We have a limited scope since we're currently only dealing with `a`, `img` and `mark`, and otherwise we default to the original tag.

That's only my opinion though 😅 Was there anything you were specifically concerned about with this approach?

mikachan (commented):
Thanks so much for handling this @jffng! I can see this will fix the issues we've been seeing with translations ✨ 

@creativecoder, if you get a chance it would be great to hear your thoughts on this.

includes/create-theme/theme-locale.php
jffng: I think just needing to add cases for specific tags and if the attributes changes, it may break. 

But the options for adding inline HTML / formatting inside blocks are rather limited, so maybe it's okay.

includes/create-theme/theme-locale.php
```diff
@@ -28,6 +28,72 @@ private static function escape_text_content( $string ) {
 
 		$string = addcslashes( $string, "'" );
```
creativecoder, line 33: Perhaps there's enough going on here this should have it's own utility class?

includes/create-theme/theme-locale.php
```diff
@@ -28,6 +28,72 @@ private static function escape_text_content( $string ) {
...
+							CBT_Theme_Media::add_media_to_local( array( $src ) );
+							$relative_src = CBT_Theme_Media::get_media_folder_path_from_url( $src ) . basename( $src );
+							$tokens[]     = "<img style=\"{$style}\" src=\"' . esc_url( get_stylesheet_directory_uri() ) . '{$relative_src}\" alt=\"{$alt}\">";
+							break;
+						// Handle highlights.
+						case 'mark':
+							$style    = esc_attr( $p->get_attribute( 'style' ) );
+							$class    = esc_attr( $p->get_attribute( 'class' ) );
+							$tokens[] = "<mark style=\"{$style}\" class=\"{$class}\">";
+							break;
```
creativecoder: I think we should be printing out `esc_` functions here, rather than executing them. Typically the advice is to escape as late as possible, which means we should see those functions in the PHP output.

includes/create-theme/theme-locale.php
creativecoder: Do we have to deconstruct the attributes and reassemble them? I'm worried about dropping attributes or other parts of the markup that might be added by plugins or filters.

For example, could we replace `<a href="https://wordpress.org">WordPress</a>` with something like
```php
<a href="<?php echo esc_url( 'https://wordpress.org' ) >?"><?php echo esc_html__( 'WordPress' ) ?></a>
``` 
but maintain any other attributes added to the `<a>` tag?

I'm not really worried about escaping every possible attribute, I think the main thing is to make sure the translated strings are escaped as a form of user input.

includes/create-theme/theme-locale.php
```diff
@@ -28,6 +28,72 @@ private static function escape_text_content( $string ) {
...
+				}
+			} else {
+				// If it's not a tag, just add the text content.
+				$text .= esc_html( $p->get_modifiable_text() );
+			}
+		}
+		// If tokens is not empty, format the string using sprintf.
+		if ( ! empty( $tokens ) ) {
+			// Format the string, replacing the placeholders with the formatted tokens.
+			return "<?php /* Translators: %s are html tags */ echo sprintf( esc_html__( '$text', '" . wp_get_theme()->get( 'TextDomain' ) . "' ), " . implode(
```
creativecoder: `Translators: %s are html tags` doesn't seem like a very helpful comment. From what I understand, the best practices for using multiple placeholders includes

- Use numbered placeholders (`%1$s` rather than `%s`)
- Name the tag in the comment: (e.g. `/* translators: %1$s: start of link (<a>), %2$s: end of link (</a>) */`)

includes/create-theme/theme-locale.php
```diff
@@ -28,6 +28,72 @@ private static function escape_text_content( $string ) {
...
+				', ',
+				array_map(
+					function( $token ) {
+						return "'$token'";
+					},
+					$tokens
+				)
+			) . ' ); ?>';
+		}
```
creativecoder, line 53: If there's a line break (`<br>`), that seems like an indicator to break the content there to create a separate string. I think that will help avoid the situation @madhusudhand [found](https://github.com/WordPress/create-block-theme/pull/719#pullrequestreview-2300097301).

creativecoder (commented):
Thanks for trying this; it looks like a very good start!

includes/create-theme/theme-locale.php
madhusudhand: If the content itself is having a `%s`, it breaks the page because of the improper closings.
Replacing such instances with `%%s` or `%%1$s` will fix things.

tests/test-theme-templates.php
```diff
@@ -262,7 +262,12 @@ public function test_localize_verse() {
...
 		<!-- /wp:verse -->';
 		$new_template      = CBT_Theme_Templates::escape_text_in_template( $template );
-		$this->assertStringContainsString( "<?php esc_html_e('Here is some <strong>verse</strong> to localize', '');?>", $new_template->content );
+
+		$expected_output = '<!-- wp:verse -->
+			<pre class="wp-block-verse"><?php /* Translators: %s are html tags */ echo sprintf( esc_html__( \'Here is some %sverse%s to localize\', \'\' ), \'<strong>\', \'</strong>\' ); ?></pre>
+		<!-- /wp:verse -->';
+
+		$this->assertStringContainsString( $expected_output, $new_template->content );
 	}
```
madhusudhand, line 271: Let's add a test case for this `If the content itself is having a %s, it breaks the page because of the improper closings.`

includes/create-theme/theme-locale.php
jffng: I'm not sure we can or should assume this, see for example the screen recording and use of `<br>` in issue https://github.com/WordPress/create-block-theme/issues/682.

includes/create-theme/theme-locale.php
jffng: I tried splitting this out in [bf8ff55](https://github.com/WordPress/create-block-theme/pull/719/commits/bf8ff55e9629f9bf31317bf18e806f5d086b61ef).

includes/create-theme/theme-locale.php
jffng: > I'm worried about dropping attributes or other parts of the markup that might be added by plugins or filters.

That's a good point. I could not find a way with the HTML Tag Processor to just carry over the whole tag token and all its attributes. But I refactored the approach so that all the attributes should be carried over, rather than processing the attributes based on the type of tag: https://github.com/WordPress/create-block-theme/pull/719/commits/678a8ef22ddecc2e0b4190dd141cfa0c4436cd1e#diff-fc11ad66397ed68725a8fe74f3031f104368f850b5b94370123482ec9e719e1bR54-R70

includes/create-theme/theme-locale.php
jffng: Done here: https://github.com/WordPress/create-block-theme/pull/719/commits/678a8ef22ddecc2e0b4190dd141cfa0c4436cd1e

includes/create-theme/theme-locale.php
jffng: > If the content itself is having a %s, it breaks the page because of the improper closings.

Good catch, I fixed it here: 3f0adfac69e3a240e58257b1bbe40d1db8bef9a4

tests/test-theme-templates.php
jffng: Done here: [4504111](https://github.com/WordPress/create-block-theme/pull/719/commits/4504111b0900b9d377984ab2794a297f94a9659e)

jffng:
I think I addressed all the feedback:

- Changed approach so we don't drop any attributes
- Refactored to its own class.
- Handled case when text contains `%` and added a test for it

Ready for another review.

mikachan (approved):
Thanks so much @jffng! I've tested the latest changes out and I believe everything is working as described. Great job 👏 

Here are my test results, using the Adventurer theme as my test theme, and editing the Blog Home template (which uses a list of headings with links):

| trunk | This PR |
| ----- | ------ |
| image | image |

includes/create-theme/theme-locale.php
```diff
@@ -28,6 +31,24 @@ private static function escape_text_content( $string ) {
...
 		$string = addcslashes( $string, "'" );
 
+		$p = new CBT_Token_Processor( $string );
+		$p->process_tokens();
+		$text             = $p->get_text();
+		$tokens           = $p->get_tokens();
+		$translators_note = $p->get_translators_note();
+
+		if ( ! empty( $tokens ) ) {
+			return "<?php $translators_note echo sprintf( esc_html__( '$text', '" . wp_get_theme()->get( 'TextDomain' ) . "' ), " . implode(
```
creativecoder: I'm not certain, but I suspect the translators comment may need to be on a separate line, immediately above the `esc_html__` function, to be parsed correctly.

includes/create-theme/theme-token-processor.php
```diff
@@ -0,0 +1,127 @@
...
+					$attributes = $this->p->get_attribute_names_with_prefix( '' );
+
+					foreach ( $attributes as $attr_name ) {
+						$attr_value = $this->p->get_attribute( $attr_name );
+						$token     .= $this->process_attribute( $attr_name, $attr_value );
+					}
+
+					$token                  .= '>';
+					$this->tokens[]          = $token;
+					$this->translators_note .= '%' . $this->increment . "\$s is the start of a '" . $token_name . "' HTML element";
```
creativecoder: Do we have any way of differentiating self closing tags? For example "%3$s is the start of a 'br' HTML element" might be a little confusing, since it doesn't (usually) have a closing tag.

creativecoder (commented):
This is working well at solving the problem in my testing! 🎉

Here's the block markup I used:

```html
<!-- wp:paragraph {"align":"center"} -->
<p class="has-text-align-center"><strong>Études</strong> is a pioneering firm that seamlessly merges creativity<br>and functionality to <a href="https://wordpress.org" data-type="link" data-id="https://wordpress.org">redefine <em>architectural</em> excellence</a>.</p>
<!-- /wp:paragraph -->
```

And here's the output in the .php pattern file:

```php
<?php /* Translators: %1$s is the start of a 'strong' HTML element, %2$s is the end of a 'strong' HTML element, %3$s is the start of a 'br' HTML element, %4$s is the start of a 'a' HTML element, %5$s is the start of a 'em' HTML element, %6$s is the end of a 'em' HTML element, %7$s is the end of a 'a' HTML element */  echo sprintf( esc_html__( '%1$sÉtudes%2$s is a pioneering firm that seamlessly merges creativity%3$sand functionality to %4$sredefine %5$sarchitectural%6$s excellence%7$s.', 'twentytwentyfour' ), '<strong>', '</strong>', '<br>', '<a href="' . esc_url( 'https://wordpress.org' ) . '" data-type="link" data-id="https://wordpress.org">', '<em>', '</em>', '</a>' ); ?>
```

A few changes I think we need to make this output correct

- I believe the `/* translators:` comment needs to be on a separate line, immediately above the call to the translation function (but I'm not 100% sure)
- When there are multiple placeholders in a string, the comment should be numbered, e.g. `/* Translators: 1: start of 'strong' HTML element, 2: ...` ([see plugin docs](https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/#variables))
- If there's a way to differentiate between tags that are self closing (like `<br>`), it would be nice to modify the comment (`1: 'br' HTML element, 2: ...`), so it's not confusing when there's no corresponding closing tag

includes/create-theme/theme-locale.php
mikachan: This should be addressed in https://github.com/WordPress/create-block-theme/pull/719/commits/1e317ce80b0ef55f28dccad825264b5410434f9f.

includes/create-theme/theme-token-processor.php
mikachan: Handled in https://github.com/WordPress/create-block-theme/pull/719/commits/6b5fe21f2372f6918cb0e1f6e825b7721d44cfab.

mikachan:
Thanks for all the feedback on this one!

I think I've addressed this feedback:

> I believe the /* translators: comment needs to be on a separate line, immediately above the call to the translation function (but I'm not 100% sure)

Done in https://github.com/WordPress/create-block-theme/pull/719/commits/1e317ce80b0ef55f28dccad825264b5410434f9f.

> When there are multiple placeholders in a string, the comment should be numbered, e.g. /* Translators: 1: start of 'strong' HTML element, 2: ... ([see plugin docs](https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/#variables))

Done in https://github.com/WordPress/create-block-theme/pull/719/commits/c6652860bd350295ad7c85a00033936b132308a1.

> If there's a way to differentiate between tags that are self closing (like <br>), it would be nice to modify the comment (1: 'br' HTML element, 2: ...), so it's not confusing when there's no corresponding closing tag

Done in https://github.com/WordPress/create-block-theme/commit/6b5fe21f2372f6918cb0e1f6e825b7721d44cfab.

Would appreciate any reviews on this so we can hopefully land this soon 🙇

matiasbenedetto (approved):
This is testing fine for me :+1: 
I think it's good enough to solve the translation needs of theme builders.
Let's merge.

annezazu:
Thank you all so much for getting this done!

iamtakashi:
Nice! Thank you so much for the fix.

PR #720, opened 2024-09-10
Fix Image URL Localization URI

### LINKED ISSUE #714: Image URL Localization URI
When using a child theme, `<?php echo esc_url( get_stylesheet_directory_uri() ); ?>` is different.
The image should be the URI when saved, so `<?php echo esc_url( get_template_directory_uri() ); ?>` would be correct.

Reproduction steps
1. Create a pattern that includes an image in the parent theme
2. Save the theme
3. Create an empty child theme of the theme created in 2
4. Activate the child theme.
The image path changes and assets are not displayed.

shimotmk:
I fixed the image URI path as per this issue.

fix #714


matiasbenedetto (commented):
This seems to work fine; it could be merged. However, I found some issues (not strictly related to this change) when creating child themes and bundling images. For example, If I added an image in a child theme and when creating the child theme, the image was bundled in the parent theme,  that would need to be fixed, too.

madhusudhand (commented):
It is failing in the following scenario.

Steps:

1. Active a block theme such as `remote` and create a child theme.
2. Now add an image to the child theme, and save.
3. Now save changes to theme by selecting `Save Template Changes`, `Localize Text`, `Localize Images`.
4. Observe that the template localizes the image with `get_template_directory_uri()`
5. But the image fails to load in the editor.

shimotmk:
@madhusudhand
Thank you for your comment. I was able to reproduce the problem. I was able to solve it by setting a condition with `is_child_theme`.

madhusudhand (approved):
@shimotmk Thanks for addressing the feedback.
Changes looks good to me! 🚀

vcanales (approved):
> @matiasbenedetto For example, If I added an image in a child theme and when creating the child theme, the image was bundled in the parent theme, that would need to be fixed, too.

I think we can address that situation after this fix is merged.

PR #723, opened 2024-09-18
Removed all of the code relating to 'folder management' of a theme.

### LINKED ISSUE #500: Cannot delete a new theme cloned using the subfolder option
## Environment
WordPress: 6.4.2
Create Block Theme: 1.13.8 

## Details

From the Create Block Theme panel of the site editor, enter any text in the subfolder and clone the theme.

The theme will then be activated automatically, but if you try to remove the theme after it has been deactivated, an error will occur.

I tested the following flow.

1. select Create Block in the Create Block Theme panel in the site editor
2. Fill in the required information and click Clone Theme.  

image

3. The cloned theme will be activated automatically. We can see a message "Theme cloned successfully. The editor will now reload."
4. Activate another block theme (disable the cloned theme)
5. Delete the disabled cloned theme from Theme in Appearance

image

6. Error message is displayed and cannot delete. We can see a message "Deletion failed: The requested theme does not exist.".

image

Inside of `wp-content/themes/` is like this. Didn't remove subfolder and cloned theme folder.

screenshot

Of course, this can be handled by accessing the Themes folder and deleting the corresponding folder manually, but this is inconvenient.

### LINKED ISSUE #716: Add some kind of doc/explanation about the 'Theme subfolder' option.
## What?

In the 'Edit theme metadata' modal, there's an option called  'Theme subfolder,' but it's not clear what it does or what to expect after changing it.

image

## Expected

It should include at least some comment/doc explaining the behavior of this not obvious field.
Other fields in this modal have some explanations about their expected content. Example:
image

pbking:
Removes reference to it from the Metadata modal.

Does not leverage it when cloning a theme or creating a new theme.

Themes in folders continue to export as expected.  Themes cloned from themes in folders are cloned as expected, but the `/themes` folder is now targeted as the new theme's destination rather than whatever folder the origin theme was in.

Fixes #716 
Fixes #500

### DIFF
--- a/includes/class-create-block-theme-api.php
+++ b/includes/class-create-block-theme-api.php
@@ -350,13 +350,6 @@ class CBT_Theme_API {
 			CBT_Theme_Utils::replace_screenshot( $theme['screenshot'] );
 		}
 
-		// Relocate the theme to a new folder
-		$response = CBT_Theme_Utils::relocate_theme( $theme['subfolder'] );
-
-		if ( is_wp_error( $response ) ) {
-			return $response;
-		}
-
 		wp_cache_flush();
 
 		return new WP_REST_Response(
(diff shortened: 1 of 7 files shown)


t-hamano (approved):
LGTM! 

I was worried about backward compatibility issues, i.e. what would happen if I had already moved themes into a subfolder, but when I switch to this branch, themes that have been moved into a subfolder continue to work properly.

Since there is a conflict, it may be a good idea to retest after resolving the conflict just to be safe.

vcanales (approved):
Conflicts solved, retested, and it works as advertised.

Thanks @pbking!

PR #726, opened 2024-09-20
Add a main tag wrapper around the query loop

### LINKED ISSUE #725: Move the main tag to a wrapper instead of the query loop
Just a heads up, but since CBT provides a blank template as a starter, the template has the main tag on the query loop. This is going to cause problems when zoom out lands, because zoom out looks for a main tag as a container where all the main sections should be. If the query loop is the main tag, then zoom out will add the inserters inside it. I suggest we wrap the query loop in a group block and add the main tag to that instead

MaggieCabrera:
Closes https://github.com/WordPress/create-block-theme/issues/725

This moves the `main` element to a group wrapping the query loop so it works fine with zoom out mode

### DIFF
--- a/assets/boilerplate/templates/index.html
+++ b/assets/boilerplate/templates/index.html
@@ -1,39 +1,43 @@
 <!-- wp:template-part {"slug":"header","tagName":"header"} /-->
 
-<!-- wp:query {"tagName":"main","layout":{"inherit":true}} -->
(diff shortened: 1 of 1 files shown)


getdave (commented):
This change looks solid from a Zoom Out perspective 👍

mikachan (approved):
LGTM!

PR #729, opened 2024-09-22
Add __nextHasNoMarginBottom to BaseControl-based components

### LINKED ISSUE #728: Add __nextHasNoMarginBottom to BaseControl-based components
As presented in the following issue, BaseControl-based components are planned to have margin-free style as the default behavior.

https://github.com/WordPress/gutenberg/pull/64408

As part of that effort, components that do not have `__nextHasNoMarginBottom` applied will display a deprecation warning, for example the one below:

```
Bottom margin styles for wp.components.TextareaControl is deprecated since version 6.7 and will be removed in version 7.0. Note: Set the `__nextHasNoMarginBottom` prop to true to start opting into the new styles, which will become the default in a future version. 
```

This warning should already be visible on sites where both Gutenberg and CBT are enabled and `SCRIPT_DEBUG` is `true`. As this deprecation is shipping in WordPress 6.7, more sites will see the warning.

Before WordPress 6.7 is released, I think it is necessary to identify components that are causing warnings in CBT and fix them.

t-hamano:
Closes #728

As presented in the following issue, BaseControl-based components are planned to have margin-free style as the default behavior:

https://github.com/WordPress/gutenberg/pull/64408

For this future effort, and to suppress browser warnings, we need to add the `__nextHasNoMarginBottom` prop.

By adding `__nextHasNoMarginBottom`, the bottom margin of the BaseControl-based components will be removed, making the stacked layout a bit cramped. To prevent this, I increased the value of the `spacing` prop of the `VStack` component.

There should be very little visual change caused by this PR.

### DIFF
--- a/src/landing-page/create-modal.js
+++ b/src/landing-page/create-modal.js
@@ -111,6 +111,7 @@ export const CreateThemeModal = ( { onRequestClose, creationType } ) => {
 					) }
 				</Text>
 				<TextControl
+					__nextHasNoMarginBottom
 					label={ __(
 						'Theme name (required)',
 						'create-block-theme'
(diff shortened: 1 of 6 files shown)


src/landing-page/create-modal.js
```diff
@@ -120,15 +121,13 @@ export const CreateThemeModal = ( { onRequestClose, creationType } ) => {
 					onChange={ ( value ) =>
 						setTheme( { ...theme, name: value } )
 					}
-				/>
-
-				<Text variant="muted">
-					{ __(
+					help={ __(
```
t-hamano, line 124: This change is a small improvement. It prevents a large space from appearing between the input field and the help text.

| Before | After |
|--------|--------|
| image| image|

madhusudhand (approved):
Verified the changes and UI layout. It looks good to me! 🚢 

Spacing between the input fields is even and much better now with added VStack 🎉

t-hamano:
@madhusudhand Thanks for the review!

Issue #707, opened 2024-08-23
Modified default font size presets are not added to the resulting theme.json


matiasbenedetto:
## What?
Let's say you created a blank theme and modified the default font sizes to create your theme. If you save the theme, you expect those changes to be saved, but that isn't the case. The changes to the default font sizes are not saved in the theme.json, and the settings have been removed, so those changes are lost.

![image](https://github.com/user-attachments/assets/60a228c6-0d9c-4948-8eb7-19f8d3a05722)

## Expected
If the default font sizes were changed, they should be added to the theme.json file when the user saves, exports, or creates style variations.

## Reproduction steps:
1. Create a blank theme
2. Edit the default font size presets.
3. Save changes to the theme. You will see that the edited font sizes are not present in the resulting theme.json file and the default font sizes were reset to the default.

eirichmond:
I can confirm this behavior and explain why it happens.

By default, `defaultFontSizes` is set to `true`. This means the `defaultFontSizes` are sourced from Core or Global Styles if saved in the database.

When you create a blank theme, no `fontSizes` are defined in the `theme.json` file by default. This differs from other settings like `settings.layout` or `settings.spacing.units`, which are included by default. Additionally, `defaultFontSizes` remains true and is populated from Core.

When a user makes and saves changes via Global Styles, those changes are stored and read from the database, while `defaultFontSizes` continues to be `true`.

If the user saves changes back to the theme, the plugin processes `defaultFontSizes: true`, clears the database, and saves an empty `settings.typography.fontSizes` section back to the `theme.json` file. As a result, when revisiting Global Styles, the font size presets revert to the default values provided by Core.

**Proposed Solution:**

A potentially opinionated but effective solution is to always set `defaultFontSizes: false` and populate `settings.typography.fontSizes` with predefined sizes (e.g., small, medium, large, and x-large). This ensures that even when a blank theme is created, the presets will always be read from the `theme.json` file and saved back there when modified.

Alternatively, after creating a blank theme, you should manually set `defaultFontSizes: false` and define custom sizes for `settings.typography.fontSizes` (e.g., small, medium, large, and x-large) to maintain consistent behavior.

I've submitted a PR for this https://github.com/WordPress/create-block-theme/pull/749

PR #749, opened 2024-11-27
theme json boilerplate fix to ensure fontSizes are saved back to the theme

eirichmond:
related to #707

### DIFF
--- a/includes/create-theme/resolver_additions.php
+++ b/includes/create-theme/resolver_additions.php
@@ -79,6 +79,10 @@ function cbt_augment_resolver_with_utilities() {
 
 			$data = $theme->get_data();
 
+			// move Font size preset settings from 'default' to 'theme' to ensure
+			// any changes made via Global Styles are saved back to the theme
+			$data = static::font_size_preset_changes( $data );
+
 			// Add the schema.
 			if ( empty( $schema ) ) {
 				global $wp_version;
@@ -106,6 +110,25 @@ function cbt_augment_resolver_with_utilities() {
 				: parent::get_user_data();
 		}
 
+		/**
+		 * This checks if the fontSizes has been set and saved to the database.
+		 * If it has then the $data variable is changed and will later be used to save back the theme.
+		 * A method is used from the parent class if it exists to get any user changed settings.
+		 *
+		 * @param array $data
+		 * @return array $data
+		 */
+		public static function font_size_preset_changes( $data ) {
+			$user_data = parent::get_user_data();
+			if ( method_exists( $user_data, 'get_settings' ) ) {
+				$user_data = $user_data->get_settings();
+				if ( isset( $user_data['typography']['fontSizes'] ) && ! empty( $user_data['typography']['fontSizes'] ) ) {
+					$data['settings']['typography']['defaultFontSizes'] = false;
+					$data['settings']['typography']['fontSizes']        = $user_data['typography']['fontSizes']['default'];
+				}
+			}
+			return $data;
+		}
 		/**
 		 * Stringify the array data.
 		 *
--- a/assets/boilerplate/theme.json
+++ b/assets/boilerplate/theme.json
@@ -10,34 +10,11 @@
 			"units": [ "%", "px", "em", "rem", "vh", "vw" ]
 		},
 		"typography": {
-			"defaultFontSizes": false,
 			"fontFamilies": [
 				{
 					"fontFamily": "-apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen-Sans, Ubuntu, Cantarell, 'Helvetica Neue', sans-serif",
 					"name": "System Font",
 					"slug": "system-font"
-				}
-			],
-			"fontSizes": [
-				{
-					"name": "Small",
-					"size": "13px",
-					"slug": "small"
-				},
-				{
-					"name": "Medium",
-					"size": "20px",
-					"slug": "medium"
-				},
-				{
-					"name": "Large",
-					"size": "36px",
-					"slug": "large"
-				},
-				{
-					"name": "X-Large",
-					"size": "42px",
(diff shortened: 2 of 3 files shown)


matiasbenedetto:
> related to https://github.com/WordPress/create-block-theme/issues/707

Hi @eirichmond ! thanks for adding some rationale in https://github.com/WordPress/create-block-theme/issues/707#issuecomment-2503642531 and submitting this PR.

To make the proposed changes easier to review and test it's always good to provide common answers to 'what', 'why' and 'how' the PR is solving a problem and also provide some 'testing instructions' also.

matiasbenedetto (commented):
Thanks for providing a potential solution.  I get your rationale from https://github.com/WordPress/create-block-theme/issues/707#issuecomment-2503642531, but I don't think this change would solve the problem described in #707. 

Why?
`assets/boilerplate/theme.json` is only used to create new blank themes so the change will only impact brand new blank themes, other themes using the default font sizes won't be impacted.  Also this is a way to just disable the list of default font sizes for the new blank themes created but not a way to save the changes made to those default font sizes. Apart from that, I think, blank themes are supposed to be the smallest possible, and adding an array of default font sizes doesn't seem in line with that idea.

eirichmond:
@matiasbenedetto I appreciate the feedback but it's worth noting that #707 specifically refers to creating a blank theme, and this solution would only impact that particular scenario if the boilerplate is only used to create a new blank theme.

> Create a blank theme.
> Edit the default font size presets.
> Save changes to the theme.

You will see that the edited font sizes are not present in the resulting theme.json file, and the default font sizes are reset to the Core defaults.

It was also stated that this approach "is just a way to disable the list of default font sizes for new blank themes created, but not a way to save the changes made to those default font sizes."

This is **not true**. The proposed boilerplate change ensures that:
A) `defaultFontSizes` is disabled.
B) The default font sizes from Core are replicated, making them editable via Global Styles.
C) Changes made via Global Styles and saved to the database are not removed when saving back to the `theme.json` file via the plugin. This directly addresses and fixes the bug mentioned in #707.

By reproducing the steps outlined in the original issue, you’ll find that this approach resolves the problem.

As for other themes, it’s worth noting that if `defaultFontSizes` is set to `true` by default, the `fontSizes` configuration isn’t implemented in the same way as described in #707. This distinction is important, as the issue specifically pertains to blank themes created via the plugin and the associated workflow.

eirichmond:
To further add to the discussion, I’m not sure how this could logically work without first changing the boilerplate to include defaultFontSizes: false if you wanted to save the Global Styles font settings back to the theme. Currently, the process behaves like this after the user clicks **Save to Theme**:

1. Theme is loaded:
The `theme.json` file from the new blank theme (or the current theme) is loaded. In the case of a blank theme boilerplate, `defaultFontSizes: false` is not set, so it defaults to `defaultFontSizes: true`. Additionally, `fontSizes` is not defined in the theme boilerplate.

2. Default settings are loaded:
The default theme settings from Core are loaded, which include `defaultFontSizes: true` and the default `fontSizes`.

3. Default `fontSizes` are applied:
The default `fontSizes` are added to the theme’s `fontSizes` (which is empty at this stage) in what appears to be a recursive process.

4. Database user styles are cleared:
The user’s Global Styles settings from the database are wiped.

As a result:

- The theme.json in the theme ends up with an empty fontSizes array.

- The database user styles for font sizes are cleared.

- The Global Styles font size presets revert to the default Core fontSizes.
    
This makes it clear that without defaultFontSizes: false explicitly set and predefined fontSizes included first, the behavior will always revert to the Core defaults, making it impossible to preserve changes made in Global Styles.

eirichmond:
@matiasbenedetto, perhaps this is a better fix in my last push.

**What this fixes**: When the Typography > Font Size preset is changed via Global Styles and then saved back to the database, the changes are not reflected in the theme when **"Save Changes to Theme"** is used.

**Why this fix**: This improves the theme development workflow. When changes are made via Global Styles and **"Save Changes to Theme"** is used, any font size presets should be respected and applied.

**How it's fixed**: Before the data is stringified, a new function, `font_size_preset_changes`, runs the data through a method of the parent class to check if any font sizes have been added to the database. Rather than saving the merged data with an empty `fontSizes` reference, the data from the database is used before it is cleared and applied to the `theme.json` file. Additionally, `defaultFontSizes` is set to `false` to ensure the font sizes are respected from the theme rather than the core.

includes/create-theme/resolver_additions.php
```diff
@@ -79,6 +79,10 @@ public static function export_theme_data( $content, $extra_theme_data = null ) {
 
 			$data = $theme->get_data();
 
+			// move Font size preset settings from 'default' to 'theme' to ensure
+			// any changes made via Global Styles are saved back to the theme
+			$data = static::font_size_preset_changes( $data );
+
```
eirichmond, line 85: the initial function to start the check and change the $data variable

includes/create-theme/resolver_additions.php
```diff
@@ -106,6 +110,25 @@ public static function get_user_data() {
...
+			if ( method_exists( $user_data, 'get_settings' ) ) {
+				$user_data = $user_data->get_settings();
+				if ( isset( $user_data['typography']['fontSizes'] ) && ! empty( $user_data['typography']['fontSizes'] ) ) {
+					$data['settings']['typography']['defaultFontSizes'] = false;
+					$data['settings']['typography']['fontSizes']        = $user_data['typography']['fontSizes']['default'];
+				}
+			}
+			return $data;
+		}
 		/**
```
eirichmond, line 132: the function that gets and user data and then uses the parent class method if it exists to access data that is private, then checks if fontSizes is set then appends the settings needs to write to the theme.json file.

eirichmond (commented):
Comments added to explain the changes in more detail.

PR #753, opened 2024-12-21
Clean up package.json and update @wordpress/env

t-hamano:
- Delete `@wordpress/element` ~~and `eslint-plugin-unicorn`~~, which are not used at all.
- Update `@wordpress/env` to the latest version. phpMyAdmin is also supported, so I opted in.

## Testing Instructions

- `npm run wp-env start -- --update`
- Access phpMyAdmin of the development instance: http://localhost:9000/
- Access phpMyAdmin of the tests instance: http://localhost:9001/


package.json
```diff
@@ -29,8 +29,7 @@
 		"@emotion/babel-plugin": "^11.12.0",
 		"@wordpress/base-styles": "^5.7.0",
 		"@wordpress/browserslist-config": "^6.7.0",
-		"@wordpress/element": "^6.7.0",
```
madhusudhand, line 32: React components on the admin and sidebar are depending on `@wordpress/element`.

t-hamano: This library is not used during the build process.

wp-scripts simply transpiles JS code into code that the browser can interpret. All APIs and components work through the global `wp` object provided by WordPress.

In fact, CBT should work as before. This is the same as it works correctly even if `@wordpress/components`, `@wordpress/block-editor`, etc. do not exist as dependencies.

madhusudhand: Thanks for the details. Builds work successfully!

madhusudhand (approved):
LGTM!

PR #754, opened 2024-12-24
PHPCS: Remove unused files and adjust release files

t-hamano:
## Remove unused files

`.phpcs.xml.dist` is not used in composer, so delete it. `phpcs.xml.dist` is actually used.

## adjust release files

I noticed that the release file contained a PHPCS configuration file, so I updated the `.distignore`:

source

https://plugins.trac.wordpress.org/browser/create-block-theme/trunk


carolinan (approved):
I would assume that anyone who wants to use the dot prefix and override the phpcs.xml.dist would have their own specific file, and not use the copy from the repo?

And I agree that this file does not need to be released with the plugin.

t-hamano:
@carolinan Thanks for the review!

> I would assume that anyone who wants to use the dot prefix and override the phpcs.xml.dist would have their own specific file, and not use the copy from the repo?

I don't know what this means, but I think there should only be one ruleset.

PR #762, opened 2025-04-08
Code Quality: fix browser deprecation warning

t-hamano:
This PR suppresses browser warning logs that occur in WP 6.8. I made changes based on the field guide from the following three perspectives:

### 1. [Updates to user-interface components in WordPress 6.8 - Default 36px sizes are now deprecated](https://make.wordpress.org/core/2025/03/25/updates-to-user-interface-components-in-wordpress-6-8/#default-36px-sizes-are-now-deprecated)

Add `__next40pxDefaultSize` to the form component that is causing the deprecation warning. This means the height of the component will change from 36px to 40px. Example:

| Before | After |
|--------|--------|
| image | image| 

### 2. [Updates to user-interface components in WordPress 6.8 - Stabilized Navigator](https://make.wordpress.org/core/2025/03/25/updates-to-user-interface-components-in-wordpress-6-8/#stabilized-navigator)

Use `Navigatior.BackButton` instead of `__experimentalNavigatorToParentButton`. For backward compatibility, check whether `Navigatior.BackButton` is undefined, and if it is, use the conventional component.

### 3. [Data: A helpful performance warning for developers in the ‘useSelect’ hook](https://make.wordpress.org/core/2025/03/12/data-a-helpful-performance-warning-for-developers-in-the-useselect-hook/)

Refactored one of the `useSelect` hooks, which is not used properly. The useSelect Hook itself should not be asynchronous.

### DIFF
--- a/src/editor-sidebar/create-panel.js
+++ b/src/editor-sidebar/create-panel.js
@@ -131,6 +131,7 @@ export const CreateThemePanel = ( { createType } ) => {
 			<VStack>
 				<TextControl
 					__nextHasNoMarginBottom
+					__next40pxDefaultSize
 					label={ __( 'Theme name', 'create-block-theme' ) }
 					value={ theme.name }
 					onChange={ ( value ) =>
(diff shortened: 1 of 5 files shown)


pbking (approved):
lgtm

PR #764, opened 2025-04-15
Update wp-env configuration 

pbking:
Update wp-env configuration to use 'development mode' and to store themes in an accessible place for development.

with these environment variables set:
```
"WP_ENVIRONMENT_TYPE": "development",
"WP_DEVELOPMENT_MODE": "all"
```

Things like patterns are not cached and other improvements that make sure the WordPress environment provided by wp-env is the best to develop in.

Additionally the /themes folder is mapped to a local folder so that themes created and saved in the development environment can be easily observed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,5 @@ build/
 
 # phpunit
 *.result.*
+
+dev-env/
--- a/.wp-env.json
+++ b/.wp-env.json
@@ -6,7 +6,12 @@
 	"plugins": [ "." ],
 	"config": {
 		"WP_UPLOAD_MAX_FILESIZE": "128M",
-		"WP_MEMORY_LIMIT": "256M"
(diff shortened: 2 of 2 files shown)


t-hamano (commented):
> Additionally the /themes folder is mapped to a local folder so that themes created and saved in the development environment can be easily observed.

Does this mean that in developing CBT, a custom theme may be necessary? Will the default theme not suffice?

pbking:
A default theme is fine to start with, however observing the themes that CBT creates/exports is important for development and this change allows for those themes to be easily observed (be that default themes, installed, created, etc).

For example, create a new blank theme called "Blankey" in the dev environment, make changes to it and save it and you will see that changes in the new folder.

Previous to this change I have to load the CBT plugin into another environment in order to see the files it's manipulating when saving/creating themes.

t-hamano (approved):

PR #765, opened 2025-04-15
Remove Pattern Saving Option

pbking:
This change removes the "Pattern Management" option from Create Block Theme.

Notes that patterns that are generated to support localization and local assets used in templates remain.  (They are not "editable" however they are updated when changes to a template are made are re-exported.  This is not new behavior.)

Exporting "synced" patterns as "unsynced" patterns is very confusing and breaks the workflow of creating patterns for a theme.  Considering how confusing and damaging it is to themes being edited I propose a completely different mechanism for editing patterns.  In the meantime this should probably be removed.


t-hamano (commented):
Thanks for the PR!

I'm not familiar with the Pattern Saving Option, but this feature seems to be a part of #287. Is there an agreement with the folks involved in #287 about removing this feature?

pbking:
No, there has not yet been discussion about removing this feature.  I don't believe it was a successful addition though so opening this PR was my way to get the conversation started.  I would certainly appreciate input for everyone who has an opinion.

bph:
This is not necessary anymore, as  @scruffian fixed pattern saving with #790. I close this PR. It's also clear that pattern management has a place in this plugin. So if something is not working for you. A new issue would be the best place to start.

PR #771, opened 2025-04-24
added filters for changing paths

Lovor01:
Hello! I added filters for changing localized paths, because I needed that, and others may need that too.
Cheers!


pbking (approved):
Oh, good call.  LGTM.  Thanks!

PR #775, opened 2025-05-18
Bump minimum required WordPress version to 6.7

t-hamano:
Similar to #704

### DIFF
--- a/src/editor-sidebar/screen-header.js
+++ b/src/editor-sidebar/screen-header.js
@@ -17,7 +17,7 @@ import { chevronRight, chevronLeft } from '@wordpress/icons';
 
 const ScreenHeader = ( { title, onBack } ) => {
 	// TODO: Remove the fallback component when the minimum supported WordPress
-	// version was increased to 6.7.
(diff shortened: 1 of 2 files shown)


src/editor-sidebar/screen-header.js
```diff
@@ -17,7 +17,7 @@ import { chevronRight, chevronLeft } from '@wordpress/icons';
 
 const ScreenHeader = ( { title, onBack } ) => {
 	// TODO: Remove the fallback component when the minimum supported WordPress
-	// version was increased to 6.7.
+	// version was increased to 6.8.
 	const BackButton = Navigator?.BackButton || NavigatorToParentButton;
```
t-hamano, line 21: It turns out that `Navigator.BackButton` is not yet exposed in WordPress 6.7. Therefore, we found that the fallback component is necessary as long as CBT supports WordPress 6.7.

In the future, when we bump the minimum required WordPress version to 6.8, we will be able to remove this fallback code.

pbking (approved):

PR #780, opened 2025-09-06
Automatically enqueue style.css when creating a blank theme

### LINKED ISSUE #334: Automatically enqueue style.css when creating a blank theme
Pulling in feedback from the[ FSE Outreach program's Build A Block Theme exploration:](https://make.wordpress.org/test/2023/03/29/fse-program-exploration-build-a-block-theme/#comment-3002)

>There is an empty style sheet added when a blank theme is created with CBT plugin. But adding CSS to it doesn’t work until a functions.php is added to enqueue it. I think that style.css should automatically be enqueued or that a functions.php with the enqueue should be included. I am trying to use only the editor, but think that if the stylesheet is there, it should be usable.

theaminulai:
### What?
Automatically enqueue **style.css** when creating a blank theme with the Create Block Theme (CBT) plugin.

### Why?
Currently, when a blank theme is created, an empty **style.css** file is added, but adding CSS to it does not take effect until a **functions.php** file is added to enqueue the stylesheet. This creates confusion for users who want to work directly in the editor or prefer using **style.css** instead of the Additional CSS area. Automatically enqueuing **style.css** ensures that the stylesheet works out-of-the-box for those who want to use it.

**Fixes / Reference**
Fixes [#334](https://github.com/WordPress/create-block-theme/issues/334)

### How to test:

1. Install and activate the updated Create Block Theme plugin.
2. In the WP Admin, go to Appearance → Create Block Theme → Create Blank Theme.
3. Confirm that a new blank theme is generated.
4. Open the generated theme folder and check:
     - A functions.php file exists.
     - It contains enqueue code for style.css.
5. Add some simple CSS to **style.css**, for example:
```
body {
  background: pink;
}

```
6. Activate the new blank theme.
7. Visit the site’s frontend — the background should reflect your CSS changes (confirming that **style.css** is enqueued).
8. Optional: Inspect the source in the browser dev tools → check that **style.css** is being loaded.

### DIFF
--- a/assets/boilerplate/functions.php
+++ b/assets/boilerplate/functions.php
@@ -1,39 +1,55 @@
 <?php
-/**
- * Enqueue the theme's main stylesheet.
- *
- * This function checks if the `style.css` file exists in the theme directory
- * before attempting to enqueue it. It ensures that missing files do not
- * generate PHP warnings and also applies the theme version for cache busting.
- *
- * Usage: Hooked into 'wp_enqueue_scripts'.
- *
- * @since 1.0.0
- *
- * @see https://developer.wordpress.org/reference/functions/wp_enqueue_style/
- * @see https://developer.wordpress.org/reference/functions/get_stylesheet_directory/
- * @see https://developer.wordpress.org/reference/functions/get_stylesheet_uri/
- * @see https://developer.wordpress.org/reference/functions/wp_get_theme/
- *
- * @return void
- */
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
 
-add_action(
-	'wp_enqueue_scripts',
-	function() {
-		$stylesheet_path = get_stylesheet_directory() . '/style.css';
-
-		if ( file_exists( $stylesheet_path ) ) {
-
-			wp_enqueue_style(
-				'theme-style',
-				get_stylesheet_uri(),
-				array(),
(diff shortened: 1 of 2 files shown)


iqbal-xs:
@theaminuli, @t-hamano PR looks good

t-hamano:
@theaminuli thanks for the PR!

Personally, I'm not sure if loading styles in a particular way is the best way, and I think we should discuss this a bit more before moving forward with this PR.

pbking:
Personally in my themes I usually load my styles from an asset other than the `style.css` file.  (It feels weird keeping the metadata in an asset that I compile from SCSS and doing something like `/styles/index.scss > /assets/css/index.css` is what I do.  

While it would be handy to have that setup for me for my CBT created theme that doesn't feel appropriate.  There's a ton of other ways an opinions on how/where those styles should be loaded.  If you need those loaded I feel like you're taking the created theme beyond "basic" and "universal" which is what I think is where CBT should be landing.  (RE: As opinion-less as possible)

Just my 0.02 since the reality is that I nearly always start with my own starter theme with all the things setup the way I like and just duplicate that instead of creating a blank one.

carolinan:
I think adding CSS in style.css should be discouraged, not encouraged.
For example, this theme does not come with any build tools to add rtl or minifying the styles. And I don't think it should have those tools.
(For background, in October 2025, the performance team needed to add these tools to Twenty Twenty-Five).

PR #790, opened 2025-12-03
Save all patterns

scruffian:
Enables saving all patterns (both synced and unsynced) to the theme, not just synced patterns. Previously, only synced patterns would be exported when using the "Save Patterns" option.

### User-Facing Changes:
  - The "Save Synced Patterns" option has been renamed to "Save Patterns"
  - All patterns created in the Editor (both synced and unsynced) will now be saved to the theme when this option is enabled
  - Unsynced patterns are now included in the export and removal process

 ### Technical Changes:
  - Refactored pattern processing to store raw content first, then wrap in PHP file structure after processing
  - Options are now threaded through to replace_local_pattern_references() to ensure consistent processing
  - Removed the sync status check that previously filtered out unsynced patterns

##  Why?
This allows theme designers to create their own patterns for themes.

### Test Case 1: Save All Pattern Types

  1. Go to Appearance → Editor → Patterns
  2. Create 2 patterns - a synced and an unsynced.
  3. Go to Create Block Theme in the editor
  4. Save Changes to Theme
  5. Enable "Save Patterns" checkbox
  6. Click "Save Changes"
  7. ✅ Expected: Both patterns should be saved to /patterns/ directory as .php files
  8. ✅ Expected: Both patterns should be deleted from the Editor
  9. ✅ Expected: Patterns should now load from the theme instead of the database

###  Test Case 2: Duplicate Pattern Names

  1. Manually create a pattern file: wp-content/themes/your-theme/patterns/test-pattern.php
  2. In the Editor, create a new pattern with the title "Test Pattern" (same slug)
  3. Try to save patterns using Create Block Theme
  4. ✅ Expected: Should show an error: "A pattern with this name already exists: test-pattern"

###  Test Case 3: Localization Options

  1. Create a pattern with text content: "Hello World"
  2. Go to Create Block Theme overwrite screen
  3. Enable both:
    - ✅ "Save Patterns"
    - ✅ "Localize Text"
  4. Save changes
  5. Inspect the pattern file
  6. ✅ Expected: Text should be wrapped in localization functions: <?php esc_html_e('Hello World', 'theme-slug');?>

 ### Test Case 4: Image Localization

  1. Create a pattern with an image from the Media Library
  2. Enable both:
    - ✅ "Save Patterns"
    - ✅ "Localize Images"
  3. Save changes
  4. ✅ Expected: Image should be copied to /assets/ directory
  5. ✅ Expected: Pattern should reference the local image path

 ### Test Case 5: Templates Referencing Patterns

  1. Create a pattern and add it to a template
  2. Modify the template (add a paragraph block)
  3. Save with both "Save Template Changes" and "Save Patterns" enabled
  4. ✅ Expected: Both the template and pattern should be saved
  5. ✅ Expected: Template changes should be preserved
  6. ✅ Expected: Pattern reference should be updated correctly

  ### Test Case 6: UI Label Update

  1. Go to Create Block Theme overwrite screen
  2. ✅ Expected: Checkbox should read "Save Patterns" (not "Save Synced Patterns")
  3. ✅ Expected: Help text should say "All patterns created in the Editor will be moved to the theme..."

### DIFF
--- a/includes/create-theme/theme-patterns.php
+++ b/includes/create-theme/theme-patterns.php
@@ -30,14 +30,7 @@ class CBT_Theme_Patterns {
 		$pattern_category_list = get_the_terms( $pattern->id, 'wp_pattern_category' );
 		$pattern->categories   = ! empty( $pattern_category_list ) ? join( ', ', wp_list_pluck( $pattern_category_list, 'name' ) ) : '';
 		$pattern->sync_status  = get_post_meta( $pattern->id, 'wp_pattern_sync_status', true );
-		// Store the raw content separately for processing
-		$pattern->content = $pattern_post->post_content;
-
-		return $pattern;
-	}
-
-	public static function wrap_pattern_in_php_file( $pattern ) {
-		$pattern_content  = <<<PHP
+		$pattern->content      = <<<PHP
 		<?php
 		/**
 		 * Title: {$pattern->title}
@@ -45,9 +38,9 @@ class CBT_Theme_Patterns {
 		 * Categories: {$pattern->categories}
 		 */
 		?>
-		{$pattern->content}
+		{$pattern_post->post_content}
 		PHP;
-		$pattern->content = $pattern_content;
+
 		return $pattern;
 	}
 
@@ -138,9 +131,6 @@ class CBT_Theme_Patterns {
 				CBT_Theme_Media::add_media_to_local( $pattern->media );
 			}
 		}
-
-		// Wrap the pattern content in the PHP file structure with metadata header
-		$pattern = self::wrap_pattern_in_php_file( $pattern );
 
 		return $pattern;
 	}
--- a/includes/create-theme/theme-patterns.php
+++ b/includes/create-theme/theme-patterns.php
@@ -29,7 +29,6 @@ class CBT_Theme_Patterns {
 		$pattern->slug         = wp_get_theme()->get( 'TextDomain' ) . '/' . $pattern->name;
 		$pattern_category_list = get_the_terms( $pattern->id, 'wp_pattern_category' );
 		$pattern->categories   = ! empty( $pattern_category_list ) ? join( ', ', wp_list_pluck( $pattern_category_list, 'name' ) ) : '';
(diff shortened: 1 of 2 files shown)


includes/create-theme/theme-patterns.php
```diff
@@ -160,41 +160,38 @@ public static function add_patterns_to_theme( $options = null ) {
 				$pattern        = self::prepare_pattern_for_export( $pattern, $options );
 				$pattern_exists = false;
 
-				// Check pattern is synced before adding to theme.
-				if ( 'unsynced' !== $pattern->sync_status ) {
```
mikachan, line 164: `$pattern->sync_status` is still defined in `pattern_from_wp_block` but I don't think it's used anywhere now, so we can probably remove it.

mikachan (commented):
Thanks for working on this!

Most test cases are working well:

- Test Case 1: Save All Pattern Types ✅ 
- Test Case 2: Duplicate Pattern Names ✅ 
- Test Case 3: Localization Options ✅ 
- Test Case 4: Image Localization ✅ 
- Test Case 6: UI Label Update ✅ 

But Test Case 5: Templates Referencing Patterns seems to break the template file. When I insert a pattern into a template and then save to the theme with localization, the template .html file is updated with `<!-- wp:pattern {"slug":"pixl/page"} /-->`, but the new template pattern isn't created (pixl/page in this case), which breaks the template in the editor as it can't find the template pattern. The new pattern referenced in the template is added to the theme with the correct contents.

includes/create-theme/theme-patterns.php
scruffian: Done in 199aedf4e5b0d84444cdd98447f6f0440eaa72f3

mikachan (approved):
I've reset my local env and I'm no longer seeing the issue with saving a pattern to a template. Test Case 5 is working well for me now ✅ 

I think we should bring this in! 🚢

JiveDig:
This is great! I'm curious what the expectation is when using a synced pattern with overrides? I do this all the time but haven't tested with the latest update yet.

PR #791, opened 2025-12-05
Add full and wide width support for blank theme.

torounit:
Add full and wide width support for blank theme.

This is achieved by setting the `align` property to `full` for blocks like the query block and post-template block. It uses the same method as “Twenty Twenty-Five”. (https://github.com/WordPress/twentytwentyfive/blob/trunk/patterns/template-query-loop.php)

|Before|After|
|-|-|
|before|after|

### DIFF
--- a/assets/boilerplate/templates/index.html
+++ b/assets/boilerplate/templates/index.html
@@ -2,30 +2,30 @@
 
 <!-- wp:group {"tagName":"main","layout":{"type":"constrained"}} -->
 <main class="wp-block-group">
-	<!-- wp:query {"layout":{"inherit":true}} -->
-	<div class="wp-block-query">
-		<!-- wp:post-template -->
-		<!-- wp:group -->
-		<div class="wp-block-group">
+	<!-- wp:query {"align":"full","layout":{"type":"default"}} -->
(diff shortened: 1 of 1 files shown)


carolinan:
Hi
Is there an issue for this? I'm not sure what changing the default solves and what the purpose is?

The template in the pull request does not result in a layout matching the "After" screenshot.
I don't agree that having the text edge to edge like in the screenshot is the better option, it is often considered more difficult to read the text.

If the parent container (group, query, post template) has `"layout":{"type":"default"}`, if "Inner blocks use content width" is toggled off, then setting align full on the child in the code will not do anything.
Align full and align wide are available when the layout type is constrained, when "Inner blocks use content width" is toggled on.

If what you wanted to achieve was to keep the content centered, at the expected content with, but allow for example a full width featured image, then use the current template and turn off "Inner blocks use content width" for the main group, the query, and the post template, and turn it on for the group that is **inside** the post template.

bph:
The site editor can take care of things as @carolinan suggested or my cloning the Twenty-Twenty-five. I think this can be closed. One can also create the theme with the right alignment settings and make it the starter theme for an agency team or a theme shop. The blank theme should be as plain as possible.

PR #793, opened 2025-12-08
Update "Tested up to" version to 6.9

mikachan:
This updates the "Tested up to" version to WordPress 6.9.

I'll release a new version of the plugin after this has been merged.


t-hamano (approved):
LGTM! 

After this PR is merged, we should probably rebase #792 on top of the trunk branch, right? Otherwise, the new version will be deployed with the "Tested up to" field still at 6.8.

PR #801, opened 2026-01-24
Polish sidebar using Card component

t-hamano:
The sidebar content is essentially wrapped in a `PanelBody` component, which causes a double border at the top:

double-border

In this PR, I replaced all `PanelBody` components with `Card` components, which is the same implementation as the Global Styles sidebar.

### How to Test

There are no functional changes, just make sure the layout is correct on all screens.

### Screenshots

<details><summary>Screenshots</summary>

### Root

| Before | After |
|--------|--------|
| image | image |

### Help screen

| Before | After |
|--------|--------|
| image | image |

### Create Variation page

| Before | After |
|--------|--------|
| image | image |

</details>


mikachan (approved):
This works as described, thanks @t-hamano 🙌

PR #802, opened 2026-01-24
Use wp-env for PHP unit test

t-hamano:
I noticed that PHP unit tests are failing in recent pull requests.

https://github.com/WordPress/create-block-theme/actions/runs/21311844655/job/61349179401

image

However, I'm not sure why we're using svn repo for the WordPress environment in the first place. I think using `wp-env` is much simpler.

This PR also configures a matrix of **five PHP versions** and **two WordPress versions** to ensure compatibility across multiple environments, which is similar to [the PHP unit test configuration in Gutenberg](https://github.com/WordPress/gutenberg/blob/b1870dd57d4c922e928b46ffb76c219336417acf/.github/workflows/unit-test.yml#L187-L189).

## How to test

Run the following script locally:

- `test:unit:php`: This command starts wp-env before processing PHP unit tests.
- `test:unit:php:base`: Only process PHP unit tests.

Verify that unit test failures are detected correctly. For example, change any string in [this file](https://github.com/WordPress/create-block-theme/blob/trunk/tests/CbtThemeLocale/escapeAttribute.php) and run `test:unit:php:base`:

image


.github/workflows/pr-checks.yml
```diff
@@ -66,23 +102,24 @@ jobs:
...
-                  mysql-version: '8.0'
+                  php-version: '${{ matrix.php }}'
+                  ini-file: development
+                  coverage: none
+
+            - name: Docker debug information
+              run: docker -v
 
-            - name: Run JavaScript unit tests
-              run: npm run test:unit
```
mikachan, line 79: It looks like we've removed the JS unit tests from here. Was that intentional? Although we don't currently have any JS unit test files, anyway 😅

.github/workflows/pr-checks.yml
```diff
@@ -66,23 +102,24 @@ jobs:
...
+
+            - name: Log running Docker containers
+              run: docker ps -a
 
             - name: Run PHP tests
-              run: |
-                  mysql -uroot -h127.0.0.1 -e 'SELECT version()' \
-                  && ./bin/install-wp-tests.sh --wp-version=trunk --recreate-db wordpress_test root ''  > /dev/null \
-                  && composer run-script test
+              run: npm run test:unit:php
```
mikachan: I think calling `test:unit:php` here will run `wp-env start` again. Would we be better using `test:unit:php:base`?

```suggestion
              run: npm run test:unit:php:base
```

mikachan (approved):
These two commands work well for me locally:

- `test:unit:php`
- `test:unit:php:base`

I've left two small inline comments, but they're not blockers. Thanks!

.github/workflows/pr-checks.yml
t-hamano: Yes, that's intentional. We can restore this when we need JS unit tests.

.github/workflows/pr-checks.yml
t-hamano: Nice catch! `test:unit:php:base` should work. Let's give it a try.

PR #803, opened 2026-01-24
Bump minimum required WordPress version to 6.8

t-hamano:
Similar to #775

Confirm that all features work correctly as before in WP 6.8.


mikachan (approved):
Changes look good and test well for me, thanks!

PR #804, opened 2026-01-24
Performance: Consolidate redundant APIs into getCurrentTheme

t-hamano:
The theme.json modal and the metadata modal reference their own endpoints to retrieve data that the `getCurrentTheme` selector can't retrieve, but this is redundant:

- theme.json modal: `getCurrentTheme` + `fetchThemeJson`
- metadata modal:  `getCurrentTheme` + `fetchReadmeData`

It would make more sense to include the additional data in the `getCurrentTheme` response.

Another reason for submitting this PR is to enable future updates to theme templates and metadata without a browser reload.

If all data is integrated into the default WordPress store, `invalidateResolution` will allow the selector resolution cache to be refreshed without a browser reload.


src/editor-sidebar/json-editor-modal.js
```diff
@@ -12,32 +12,30 @@ import { useState, useEffect } from '@wordpress/element';
...
+	const handleSave = () => {};
 
 	return (
 		<Modal
 			size="large"
 			title={ sprintf(
 				// translators: %s: theme name.
 				__( 'theme.json for %s', 'create-block-theme' ),
-				themeName
+				themeJsonData.name.raw
```
mikachan: Should we guard against `themeJsonData` being `null` on initial render?

```suggestion
				themeJsonData?.name?.raw ?? ''
```

src/editor-sidebar/metadata-editor-modal.js
```diff
@@ -70,37 +70,21 @@ export const ThemeMetadataEditorModal = ( { onRequestClose } ) => {
...
+			uri: themeData.theme_uri.raw,
+			version: themeData.version,
+			requires_wp: themeData.requires_wp,
+			author: themeData.author.raw,
+			author_uri: themeData.author_uri.raw,
+			tags_custom: themeData.tags.rendered,
+			screenshot: themeData.screenshot,
+			recommended_plugins: themeData.readme.recommended_plugins || '',
+			font_credits: themeData.readme.fonts || '',
+			image_credits: themeData.readme.images || '',
```
mikachan: Should we be more defensive with `themeData.readme` here? Something like this:

```suggestion
			recommended_plugins: themeData.readme?.recommended_plugins || '',
			font_credits: themeData.readme?.fonts || '',
			image_credits: themeData.readme?.images || '',
```

mikachan (commented):
Thanks for working on this, it's looking great! I've left some minor comments.

src/editor-sidebar/json-editor-modal.js
t-hamano: Yes, let's do that just in case.

mikachan (approved):
I think this is good to bring in, thanks @t-hamano!

PR #811, opened 2026-02-25
Add basic e2e tests

mikachan:
This adds basic e2e tests to this repo. The tests cover the two main plugin entry points:

- Site Editor Sidebar (specs/editor-sidebar.spec.js): Verifies the Site Editor loads and the "Create Block Theme" toolbar button is present.
- Admin Landing Page (specs/landing-page.spec.js): Verifies the page loads, the Export button is visible, the "Create Blank Theme" modal opens with a name field, and the full create-blank-theme flow completes end-to-end.

Also adds an e2e job to the CI workflow that spins up wp-env, runs the e2e tests, and uploads the artifacts on failure. Adds test:e2e and test:e2e:debug npm scripts and gitignores the artifacts/ directory.


specs/landing-page.spec.js
```diff
@@ -0,0 +1,91 @@
...
+	} );
+
+	test.afterAll( async ( { requestUtils } ) => {
+		if ( originalThemeSlug ) {
+			await requestUtils.activateTheme( originalThemeSlug );
+		}
+		try {
+			execSync(
+				`npx wp-env run cli wp theme delete ${ E2E_THEME_SLUG }`,
+				{ stdio: 'ignore' }
```
jeryj, line 42: I haven't seen a need for this before. Could you explain why this is needed? Is there a precedence for this in Gutenberg already?

specs/landing-page.spec.js
```diff
@@ -0,0 +1,91 @@
...
+			.getByRole( 'button', { name: /Create a new Blank Theme/i } )
+			.click();
+		await page.getByLabel( /Theme name/i ).fill( 'E2E Test Theme' );
+		await page
+			.getByRole( 'button', {
+				name: 'Create and Activate Blank Theme',
+			} )
+			.click();
+
+		await page.waitForURL( /site-editor/ );
```
jeryj: Could we assert some kind of success? How do we know if it worked correctly besides going to the site-editor?

specs/editor-sidebar.spec.js
```diff
@@ -0,0 +1,16 @@
...
+
+test.describe( 'Create Block Theme — Site Editor Sidebar', () => {
+	test( 'plugin sidebar button is present in the editor toolbar', async ( {
+		admin,
+		page,
+	} ) => {
+		await admin.visitSiteEditor( { canvas: 'edit' } );
+		await expect(
+			page.getByRole( 'button', { name: 'Create Block Theme' } )
+		).toBeVisible();
```
jeryj, line 14: Could we test that it works? What does it do when you click it?

specs/landing-page.spec.js
```diff
@@ -0,0 +1,91 @@
...
+	} );
+
+	test( 'export button is visible', async ( { admin, page } ) => {
+		await admin.visitAdminPage(
+			'themes.php',
+			'page=create-block-theme-landing'
+		);
+		await expect(
+			page.getByRole( 'button', { name: /Export/ } )
+		).toBeVisible();
```
jeryj: Can we assert that it works when clicked?

specs/landing-page.spec.js
```diff
@@ -0,0 +1,91 @@
...
+			'themes.php',
+			'page=create-block-theme-landing'
+		);
+		await page
+			.getByRole( 'button', { name: /Create a new Blank Theme/i } )
+			.click();
+		await expect( page.getByLabel( /Theme name/i ) ).toBeVisible();
+	} );
+
+	test( 'create blank theme end-to-end', async ( { admin, page } ) => {
```
jeryj: When I run this test file on its own, this "'create blank theme end-to-end'" test fails: `npm run test:e2e specs/landing-page.spec.js `

```
 1) [chromium] › specs/landing-page.spec.js:70:2 › Create Block Theme — Admin Landing Page › create blank theme end-to-end 

    Test timeout of 100000ms exceeded.

    Error: page.waitForURL: Target page, context or browser has been closed
    =========================== logs ===========================
    waiting for navigation until "load"
    ============================================================

      87 | 			.click();
      88 |
    > 89 | 		await page.waitForURL( /site-editor/ );
         | 		           ^
      90 | 	} );
      91 | } );
      92 |
```

jeryj (commented):
I tried testing this with a fresh install:
- `npm install`
- `npm run build`
- `npm run test:e2e`

And all the e2e tests passed. Then I tried running some of the tests in isolation and got a failure on "Create Block Theme — Admin Landing Page › create blank theme end-to-end". After the failure started, I couldn't get the test to pass anymore. I think it's a test cleanup thing, because it's failing due to the "Theme already exists" message.

npx playwright failure with theme already exists message in modal

Also, the timeout for the test is really long (100000ms) so it takes a minute and 40 seconds for the test to fail. I'm not sure what Gutenbergs is, but I feel sure it's less than that. I imagine 10 seconds is probably plenty enough for most actions?

Thanks for adding e2e tests to this! 🙌🏻

specs/landing-page.spec.js
mikachan: There's no `deleteTheme()` method available in `@wordpress/e2e-test-utils-playwright`, and the REST API doesn't support theme deletion (`/wp/v2/themes` is read-only). Because we're testing this plugin creating a theme, I think WP-CLI is the best way to clean them up. 

`wp-env` is already running at this point, so `npx wp-env run cli` should be OK to use here.

specs/landing-page.spec.js
mikachan: Yes, makes sense! I've added a test for the theme being activated in https://github.com/WordPress/create-block-theme/pull/811/commits/f76d4d9b60bc855031c431502857e7c13e22480e.

specs/editor-sidebar.spec.js
mikachan: Yes, added a test for opening the sidebar in https://github.com/WordPress/create-block-theme/pull/811/commits/4665e2edf85ab7029c3ae2128bae45e253bd33fe.

specs/landing-page.spec.js
mikachan: Added a test to ensure that the export was downloaded successfully: https://github.com/WordPress/create-block-theme/pull/811/commits/040ed222186bf6db8b862c62748665e186624419.

specs/landing-page.spec.js
mikachan: I think I've managed to fix this, and when I try `npm run test:e2e specs/landing-page.spec.js` locally it now passes. It seems like waiting for the full Site Editor to load will take a while, so I've changed it to just check if the navigation has been successful, which should be good enough for this test for now.

mikachan:
Thanks so much, @jeryj!

> Then I tried running some of the tests in isolation and got a failure on "Create Block Theme — Admin Landing Page › create blank theme end-to-end". After the failure started, I couldn't get the test to pass anymore. I think it's a test cleanup thing, because it's failing due to the "Theme already exists" message.

I think this should be fixed now, I'm able to run both test suites independently. Seems like we need to delete the test theme in the `beforeAll` and the `afterAll`.

> Also, the timeout for the test is really long (100000ms) so it takes a minute and 40 seconds for the test to fail. I'm not sure what Gutenbergs is, but I feel sure it's less than that. I imagine 10 seconds is probably plenty enough for most actions?

We're just using the default timeout settings from `@wordpress/scripts` here, which I think is what Gutenberg does too. Maybe we could add some overrides here to make them faster.

jeryj (commented):
I think there's still something wrong with the cleanup of the theme deletion. I ran these tests:

1. 🟢 `npm run test:e2e specs/landing-page.spec.js`
2. 🟢 `npm run test:e2e`
3. 🟢 `npm run test:e2e specs/editor-sidebar.spec.js`
4. 🔴 `npm run test:e2e specs/landing-page.spec.js `

The failure was the same as earlier - "theme already exists".

jeryj:
> We're just using the default timeout settings from @wordpress/scripts here

Oh - that's fine then. I looked a little more. I think it's because it's failing while waiting for the site redirection, which has a much longer default. If before the site-editor redirect check, you add a check for the success message, then it fails with a 5000ms timeout instead.

specs/landing-page.spec.js
```diff
@@ -0,0 +1,110 @@
...
+
+		await page
+			.getByRole( 'button', { name: /Create a new Blank Theme/i } )
+			.click();
+		await page.getByLabel( /Theme name/i ).fill( 'E2E Test Theme' );
+		await page
+			.getByRole( 'button', {
+				name: 'Create and Activate Blank Theme',
+			} )
+			.click();
```
jeryj, line 104:
```suggestion
		const createThemeButton = page.getByRole( 'button', {
			name: 'Create and Activate Blank Theme',
		} );

		await createThemeButton.click();

		await expect( createThemeButton ).toBeHidden();

		await expect(
			page.getByText(
				'Theme created successfully. The editor will now load.'
			)
		).toBeVisible();
```

jeryj: This makes the failure 5000ms instead of 100000ms

mikachan: Thank you!

mikachan:
> I think there's still something wrong with the cleanup of the theme deletion. I ran these tests:

I can't seem to replicate this. If I run these exact tests locally, then I don't see any failures:

1. npm run test:e2e specs/landing-page.spec.js
2. npm run test:e2e
3. npm run test:e2e specs/editor-sidebar.spec.js
4. npm run test:e2e specs/landing-page.spec.js 

I'll keep trying to reproduce, but I'm thinking it might be worth merging this as it is and fixing this in a follow-up if that's OK.

specs/landing-page.spec.js
mikachan: Reducing this timeout seems to cause this test to fail. I think the `execSync` calls used to delete the theme are quite slow, so I've changed this to only run when the theme exists to hopefully speed things up a bit: https://github.com/WordPress/create-block-theme/pull/811/commits/2071d1f3f890d3fa820d4f3db60f212ab114080b.

scruffian (approved):
```Running 6 tests using 1 worker

  ✓  1 [chromium] › specs/editor-sidebar.spec.js:7:2 › Create Block Theme — Site Editor Sidebar › plugin sidebar button is present in the editor toolbar (1.0s)
  ✓  2 [chromium] › specs/editor-sidebar.spec.js:17:2 › Create Block Theme — Site Editor Sidebar › clicking the toolbar button opens the plugin sidebar (1.1s)
  ✓  3 [chromium] › specs/landing-page.spec.js:47:2 › Create Block Theme — Admin Landing Page › page loads and shows main heading (350ms)
  ✓  4 [chromium] › specs/landing-page.spec.js:57:2 › Create Block Theme — Admin Landing Page › clicking export downloads a zip file (1.1s)
  ✓  5 [chromium] › specs/landing-page.spec.js:69:2 › Create Block Theme — Admin Landing Page › create blank theme modal opens with name field (415ms)
  ✓  6 [chromium] › specs/landing-page.spec.js:83:2 › Create Block Theme — Admin Landing Page › create blank theme end-to-end (1.0s)

  6 passed (8.9s)
  ```

Looks good, thank you.

PR #817, opened 2026-03-16
Sanitize pattern titles and categories

MaggieCabrera:
Sanitize pattern title and category values before writing them into PHP docblock headers in generated pattern files, preventing comment escape via */ sequences.

Screenshot 2026-03-16 at 11 47 26

### DIFF
--- a/includes/create-theme/theme-patterns.php
+++ b/includes/create-theme/theme-patterns.php
@@ -4,10 +4,11 @@ class CBT_Theme_Patterns {
 	public static function pattern_from_template( $template, $new_slug = null ) {
 		$theme_slug      = $new_slug ? $new_slug : wp_get_theme()->get( 'TextDomain' );
 		$pattern_slug    = $theme_slug . '/' . $template->slug;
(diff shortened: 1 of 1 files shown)


includes/create-theme/theme-patterns.php
```diff
@@ -4,10 +4,11 @@ class CBT_Theme_Patterns {
 	public static function pattern_from_template( $template, $new_slug = null ) {
 		$theme_slug      = $new_slug ? $new_slug : wp_get_theme()->get( 'TextDomain' );
 		$pattern_slug    = $theme_slug . '/' . $template->slug;
```
mikachan: Should we use the sanitised version of `$template_slug` here, instead of `$template->slug`?

MaggieCabrera: Good catch! Done in dddc06a97658129a560654f876afaf98f5991349

mikachan (approved):
This is working well for me in testing. Here is an exported pattern title using `*/`:

Image

🚢

PR #824, opened 2026-04-24
Add setting to expand typography controls in the block inspector

### LINKED ISSUE #823: Setting to show all typography (and other) controls by default while authoring a theme
When building a theme with Create Block Theme, adjusting typography in the Site Editor or block inspector takes two steps per control: open the ToolsPanel ellipsis menu, enable the control, and then set the value. For theme authors iterating on typography across blocks and elements, this adds up.

## Background from the Gutenberg repo
The ToolsPanel pattern — controls hidden behind an ellipsis menu, shown by default only when a block's [`__experimentalDefaultControls`](https://github.com/WordPress/gutenberg/pull/34064) opts them in — was introduced deliberately. The design intent, expressed across several discussions, was to reduce sidebar clutter and surface only the most common controls for each block, keeping less-used ones one click away.
Requests to persist or expand which controls are shown have come up repeatedly:

- WordPress/gutenberg#41376 asked for the ellipsis-menu selection to be remembered as a user preference. It was closed in favor of #41544.
- WordPress/gutenberg#41544 narrows the ask to persisting a selection once made. Still open.
- WordPress/gutenberg#34257 captured early UX concerns about the panel's behavior.

The hesitations in those threads are consistent: a global "show everything" toggle risks undoing the progressive-disclosure design, and per-user persistence raises questions about scope (per block type? per session? synced across devices?). Progress has been slow as a result.

## Why this fits Create Block Theme
The use case here is narrower than anything raised upstream. It's not about changing defaults for end users or persisting preferences across the editor in general. It's for a theme author, actively building a theme, who needs every typography control visible while iterating — and then goes back to the default experience when the plugin is off.
That narrow scope sidesteps the upstream concerns:

- No impact on end users of the theme
- No impact on the editor outside theme-authoring work
- Nothing written to theme.json, block.json, or any exported file
- Reversible by toggling the setting or deactivating the plugin

## Proposal
Add a setting in Create Block Theme — e.g. "Show all typography controls by default" — that expands every typography control in the editor while the plugin is active. Implementation would likely be a filter on block type args overriding `__experimentalDefaultControls` for typography, registered only while the setting is on.

## Feedback welcome
All feedback is welcome — whether on the use case, the scope, the implementation approach, or whether this belongs in Create Block Theme at all. Some specific questions:

- Does the narrow, plugin-only scope feel right, or should this be broader (e.g. covering dimensions and border too)?
- Is "all on" the right default, or should it be a checklist of which typography controls to show?
- Would a visible indicator in the editor be helpful so theme authors know the expanded view is coming from the plugin?

_Claude helped with research and structure_

bph:
## Summary

Adds an opt-in **Editor preferences** panel in the Create Block Theme sidebar with a single toggle that expands every typography control in the block inspector by default, instead of hiding them behind the ToolsPanel ellipsis menu. Intended for theme authors iterating on typography.

Fixes #823.

## Why

Adjusting typography in the Site Editor takes two steps per control today: open the ellipsis menu, enable the control, then set the value. For theme authors iterating across many blocks and elements, this adds up. The issue has more context (#823), including the related upstream discussions in [gutenberg#41376](https://github.com/WordPress/gutenberg/issues/41376) and [gutenberg#41544.](https://github.com/WordPress/gutenberg/issues/41544)

The narrow scope here — plugin-only, reversible by toggling off or deactivating CBT.

**New files**
- `src/editor-enhancements/expand-typography-controls.js` — mapping table, pure `expandTypographyDefaults()`, and a `blocks.registerBlockType` filter that populates `__experimentalDefaultControls` based on the typography supports the block already declares.
- `src/editor-sidebar/editor-preferences-panel.js` — new panel with a single `CheckboxControl`. Toggle writes the preference, fires `createInfoNotice('Applying…')`, and reloads the editor.
- `src/test/expand-typography-controls.test.js` — 13 unit tests for the pure mapping function (null/undefined, each support key, `fontAppearance` merge, falsy filtering, full shape, key preservation) plus one mapping-table sanity check.

**Modified**
- `src/plugin-sidebar.js` — side-effect import of the filter at the top (so it registers before block registration), `cog` icon, menu item under `/editor-preferences`, new `NavigatorScreen`.
- `readme.txt` — one bullet under Description.

## Design decisions

- **Typography only for v1.** Defer spacing / border / dimensions to a follow-up based on feedback.
- **Reload on toggle.** `blocks.registerBlockType` only fires at registration time. Auto-reload matches the pattern already used.
- **CBT sidebar, not the core Preferences modal.** The preference scope is `create-block-theme`, same store as existing save-panel preferences. It really is only helpful while authoring a theme.
- **Site Editor only.** CBT's sidebar bundle is gated to `$pagenow === 'site-editor.php'` in `includes/class-create-block-theme-editor-tools.php:19`, so the filter doesn't load in the post editor. The help text notes this. Broader coverage (splitting the filter into its own unconditionally-enqueued script) is a plausible v2. 

## Mapping source

Verified against `gutenberg/packages/block-editor/src/components/global-styles/typography-panel.js` on `trunk`. `fontWeight` and `fontStyle` collapse into a single `fontAppearance` control. If the `__experimental*` keys rename in core, only `expand-typography-controls.js` needs to change.

## Open questions for reviewers

From issue #823:
1. **Typography-only v1, or broader?** (spacing, dimensions, border)
2. **Reload vs live apply.** We chose reload for the reasons above; open to reconsidering if reviewers feel strongly.
3. **Visible in-editor indicator** that expansion is active — helpful, or noise?
4. **Site Editor only.** Should this also work in the post editor? Would require splitting the filter module out of the \`site-editor.php\`-gated sidebar bundle.

## Test plan
- [x] Site Editor manual: toggle OFF baseline for Paragraph, Heading, Group, Cover, Quote
- [x] Site Editor manual: toggle ON → reload → all declared typography controls visible on the above
- [x] Third-party block: Jetpack Contact Form's Label block — all typography controls expanded, no crash, no invented controls
- [x] Toggle OFF → reload → controls revert
- [x] Deactivate CBT → default Gutenberg behavior returns
- [x] Preference persists across browser reloads and sessions

A Claude-assisted PR. 

## Here are a few visusal: 

Video:  

https://github.com/user-attachments/assets/3d1bd283-6f51-4e52-b31f-9d080de615ae

Screenshot 2026-04-24 at 11 17 09

Screenshot 2026-04-28 at 16 02 26

Screenshot 2026-04-24 at 11 17 48

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -20,7 +20,7 @@ This plugin allows you to:
 - Create a new style variation
 - Export a theme
 - Save user changed templates and styles to the active theme
-- Expand all typography controls in the block inspector while authoring a theme (opt-in, under Editor preferences)
+- Expand the typography controls in the block inspector while authoring a theme (opt-in, under Editor preferences)
 
 All newly created themes or style variations will include changes made within the WordPress Editor.
 
--- a/readme.txt
+++ b/readme.txt
@@ -20,6 +20,7 @@ This plugin allows you to:
 - Create a new style variation
 - Export a theme
 - Save user changed templates and styles to the active theme
+- Expand all typography controls in the block inspector while authoring a theme (opt-in, under Editor preferences)
 
 All newly created themes or style variations will include changes made within the WordPress Editor.
 
--- a/src/editor-enhancements/expand-typography-controls.js
+++ b/src/editor-enhancements/expand-typography-controls.js
@@ -1,6 +1,7 @@
 /**
  * WordPress dependencies
  */
+// eslint-disable-next-line import/no-unresolved -- Provided as an external at build time via @wordpress/dependency-extraction-webpack-plugin.
 import { addFilter } from '@wordpress/hooks';
 import { select } from '@wordpress/data';
 import { store as preferencesStore } from '@wordpress/preferences';
--- a/src/editor-enhancements/expand-typography-controls.js
+++ b/src/editor-enhancements/expand-typography-controls.js
@@ -41,7 +41,10 @@ export function expandTypographyDefaults( typographySupport ) {
 	}
 	return {
 		...typographySupport,
-		__experimentalDefaultControls: defaults,
+		__experimentalDefaultControls: {
+			...( typographySupport.__experimentalDefaultControls ?? {} ),
+			...defaults,
+		},
 	};
 }
 
--- a/src/editor-enhancements/expand-typography-controls.js
+++ b/src/editor-enhancements/expand-typography-controls.js
@@ -1,0 +1,68 @@
+/**
+ * WordPress dependencies
+ */
+import { addFilter } from '@wordpress/hooks';
+import { select } from '@wordpress/data';
+import { store as preferencesStore } from '@wordpress/preferences';
+
+// Mapping verified against
+// gutenberg/packages/block-editor/src/components/global-styles/typography-panel.js
+// fontWeight and fontStyle collapse into the single "fontAppearance" control.
+export const TYPOGRAPHY_SUPPORT_TO_CONTROL = {
+	fontSize: 'fontSize',
+	lineHeight: 'lineHeight',
+	textAlign: 'textAlign',
+	textColumns: 'textColumns',
+	textIndent: 'textIndent',
+	__experimentalFontFamily: 'fontFamily',
+	__experimentalLetterSpacing: 'letterSpacing',
+	__experimentalTextDecoration: 'textDecoration',
+	__experimentalTextTransform: 'textTransform',
+	__experimentalWritingMode: 'writingMode',
+	__experimentalFontWeight: 'fontAppearance',
+	__experimentalFontStyle: 'fontAppearance',
+};
+
+export const PREFERENCE_SCOPE = 'create-block-theme';
+export const PREFERENCE_KEY = 'expandAllTypographyControls';
+
+export function expandTypographyDefaults( typographySupport ) {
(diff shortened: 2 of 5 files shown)


bph:
Switched the panel's outer wrapper from <PanelBody> to <Card size="small" isBorderless> + <CardBody> — the same pattern used by save-panel.js and     
  create-panel.js. This gives the checkbox the proper horizontal padding so it lines up with the "Editor preferences" header instead of sitting flush against the left edge.  

### Before 
Screenshot 2026-04-24 at 11 17 20

### After
Screenshot 2026-04-28 at 16 02 26

bph:
Also updated the handling of the reload to use prior method of alert box display on the top. 
Screenshot 2026-04-28 at 16 39 51

bph:
Note on the "Fit text" gap: only the Heading and Paragraph blocks declare `typography.fitText` support, and `FitTextControl` is registered as its own `ToolsPanelItem` outside `TypographyPanel` — it does not read `__experimentalDefaultControls`, so the existing mapping table cannot reach it. To keep the copy accurate I narrowed the label/help to refer to the typography panel rather than promising "all" controls (commit 7140ed3). Reaching Fit text would require a separate `editor.BlockEdit` filter that wraps Heading/Paragraph to inject `isShownByDefault`, which felt out of proportion for two blocks and one toggle — happy to add it as a follow-up if reviewers prefer.

mikachan (approved):
This looks good to me! Code changes look good. I think we can bring this in so we can test it out 🙌

Issue #825, opened 2026-04-29
i18n: Additional blocks not made translation-ready on theme save


bph:
Following the pattern established by #735, #737, #742, #773, #744, #756, and the HTML-comment attribute mechanism added in #788, several core blocks with user-facing string content still aren't picked up by the "Localize Text" option on theme save.

This is a continuation of the band-aid approach acknowledged in #776 — the architectural fix is tracked there. This issue is just the punch list of remaining blocks that fit the existing mechanism (`get_localizable_block_attributes` and `get_text_replacement_patterns_for_html` in `includes/create-theme/theme-locale.php`).

### JSON-attribute coverage gaps (small additions to `get_localizable_block_attributes`)

| Block | Attribute | Notes |
|---|---|---|
| `core/navigation-link` | `label` | Visible menu item text. Only helps inline navigation; menus stored as `wp_navigation` posts are out of scope here. |
| `core/navigation-submenu` | `label` | Same as above. |
| `core/home-link` | `label` | Default "Home". |
| `core/social-link` | `label` | Custom label override per icon. |
| `core/categories` | `label` | Dropdown label when `displayAsDropdown` is true. |

### Larger gap (needs a different mechanism)

- `core/file` — `downloadButtonText` and `fileName` are `rich-text`-sourced attributes stored in inner HTML, not in the block-comment JSON, so the attribute path doesn't reach them. This needs handling similar to the figcaption/alt cases in `get_text_replacement_patterns_for_html`.

### Out of scope here

- The architectural rewrite tracked in #776.
- Any block whose translatable text comes from dynamic data (post titles, taxonomy terms, etc.) — those translate via core, not via `theme.json`/templates.

cc @mikachan @scruffian @t-hamano for visibility (active on #776).

bph:
PR #826 covers the five JSON-attribute additions (`navigation-link`, `navigation-submenu`, `home-link`, `social-link`, `categories`).

Leaving this issue open after that merges to track the remaining `core/file` work — `downloadButtonText` and `fileName` are `rich-text`-sourced (stored in inner HTML, not in the block-comment JSON), so they need handling in `get_text_replacement_patterns_for_html` rather than the JSON-attribute path.

Issue #776, opened 2025-05-18
Consider the ideal approach for text localization


t-hamano:
When you save changes, if you check "Localize Text", the template content will be saved as a pattern and translation functions (`esc_html_e()`) will be applied to some of the text:

![Image](https://github.com/user-attachments/assets/965fd4e4-120a-486d-a012-ab1172d4161c)

The text that gets replaced is determined by a hard-coded list of [regular expressions](https://github.com/WordPress/create-block-theme/blob/b9e0228e39fb84c2e8be5511c1c4850348311ec6/includes/create-theme/theme-locale.php#L94-L127) and [blocks](https://github.com/WordPress/create-block-theme/blob/b9e0228e39fb84c2e8be5511c1c4850348311ec6/includes/create-theme/theme-locale.php#L159-L170).

As a result, there have been reported issues with anything that doesn't fit these not being localized:

- #737
- #742
- #773
- #744
- #756

Maintaining this hard-coded data is impractical.

Perhaps a more generic approach is needed, although this has probably been considered in the past.

iamtakashi:
Thanks for looking into it, @t-hamano! I hope this gets sorted soon. It's a pain to do this manually in every theme 😅

iamtakashi:
Hi @t-hamano, I'm curious about the plan for this. I'm all for a new approach, but at the same time, we want to help the people who work on themes every day because missing internationalisation for those blocks comes up very often.

t-hamano:
Unfortunately, I don't have enough time to solve this problem right now, and I don't have a concrete answer as to what the ideal approach would be to solve this problem.

iamtakashi:
Okay, should we then reopen all the issues that were closed due to this initiative?

t-hamano:
No, it would be better to centralize the discussion here rather than opening multiple issues with the same underlying problem.

iamtakashi:
I want the issue to be fixed sooner rather than later because the oldest issue in the group has already been left for eight months. Those strings are either left uninternationalised or the themers are fixing them manually in every single theme. If this initiative is going to take a long time, some "band-aid" fixes, such as those for other blocks, would be helpful for themers. Or will this be looked at soon?

t-hamano:
> If this initiative is going to take a long time, some "band-aid" fixes, such as those for other blocks, would be helpful for themers.

Yes, I think it's a good idea as a temporary fix.

iamtakashi:
[...]

PR #826, opened 2026-04-29
i18n: Localize label attribute of additional blocks

bph:
## Summary

Extends the "Localize Text" feature so that the `label` attribute of five more core blocks is wrapped in `esc_attr_e()` on theme save:

- `core/navigation-link`
- `core/navigation-submenu`
- `core/home-link`
- `core/social-link`
- `core/categories`

Each is added as a fall-through case in `CBT_Theme_Locale::get_localizable_block_attributes()` (`includes/create-theme/theme-locale.php`) alongside the existing `label`-only blocks. No new mechanism, no behavioural change for blocks already covered.

This continues the band-aid approach acknowledged in #776 — the architectural fix is tracked there.

Refs #825 (one part of it; `core/file` `downloadButtonText` / `fileName` will be a separate PR since those need rich-text/HTML handling).

### Caveat for navigation blocks

Navigation menus stored as `wp_navigation` posts live outside templates and are not affected by this change. The fix only helps when a Navigation block contains inline `navigation-link` / `navigation-submenu` / `home-link` children in the template markup.

## Test plan

- [x] `npm run lint:php` — clean
- [x] `npm run test:unit:php` — verified by CI (PHP 7.4 / 8.0 / 8.1 / 8.2 / 8.3, current and previous WP major).
- [x] Manual smoke in WordPress Studio: ran the production pipeline (`parse_blocks` → `escape_text_content_of_blocks` → `serialize_blocks` → `escape_block_attribute_strings`) on hand-crafted markup for each of the five new blocks. All five produced the expected `<?php esc_attr_e('<label>', '<theme-slug>');?>` wrapping.

### DIFF
--- a/includes/create-theme/theme-locale.php
+++ b/includes/create-theme/theme-locale.php
@@ -146,6 +146,11 @@ class CBT_Theme_Locale {
 			case 'core/comments-pagination-previous':
 			case 'core/comments-pagination-next':
 			case 'core/post-navigation-link':
+			case 'core/navigation-link':
+			case 'core/navigation-submenu':
+			case 'core/home-link':
+			case 'core/social-link':
+			case 'core/categories':
 				return array( 'label' );
 			case 'core/post-excerpt':
 				return array( 'moreText' );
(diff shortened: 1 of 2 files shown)


mikachan (approved):
Thank you, changes LGTM!

PR #827, opened 2026-04-30
i18n: Localise fileName and downloadButtonText of the File block

bph:
## Summary

- Adds `core/file` to `get_text_replacement_patterns_for_html()` and the dispatch switch in `escape_text_content_of_blocks()` so the File block's two rich-text-sourced attributes — `fileName` and `downloadButtonText` — are wrapped with `esc_html_e()` on theme save.
- The block.json declares `"selector": "a:not([download])"` for `fileName` and `"selector": "a[download]"` for `downloadButtonText`. Two regexes mirror that: a negative-lookahead pattern for the file-name link, and a `\sdownload[\s>]` pattern for the download button.
- Refs #825 — second of the two pieces called out there. JSON-attribute additions shipped in #826; this is the rich-text path.

## Why a separate path

`fileName` and `downloadButtonText` are `"source": "rich-text"`, so the values live as inner-HTML text inside the two `<a>` tags rather than in the block-comment JSON. `escape_block_attribute_strings()` (the JSON path) can't see them, so they need handling alongside `figcaption`, `alt`, and `<summary>`.

## Test plan

- [x] `npm run lint:php` clean
- [x] `npm run test:unit:php` — 5 new tests pass (`file`, `file without download button`, `file with custom button text`, `file with formatted file name`, idempotency); 99/99 overall
- [x] `npm run build` succeeds
- [x] Studio smoke through `parse_blocks → escape_text_content_of_blocks → serialize_blocks → escape_block_attribute_strings` confirms both attributes wrapped with the active theme's text domain
- [x] Reviewer to spot-check on a theme with a real File block via the Save Theme flow

## Notes

- **Behaviour change**: themes saved with "Localize Text" enabled will now have File blocks rewritten — they were untouched before. Desired, but worth flagging.
- The regex assumes the current core HTML shape (boolean `download` attribute on the second anchor). If a future Gutenberg release restructures the block (e.g. swaps `<a download>` for `<button>`), the download-button regex stops matching. Same fragility as every other entry in `get_text_replacement_patterns_for_html()` — accepted for the band-aid approach tracked in #776.

### DIFF
--- a/includes/create-theme/theme-locale.php
+++ b/includes/create-theme/theme-locale.php
@@ -125,6 +125,13 @@ class CBT_Theme_Locale {
 				return array( '/(alt=")(.*?)(")/' );
 			case 'core/details':
 				return array( '/(<summary[^>]*>)(.*?)(<\/summary>)/' );
+			case 'core/file':
+				return array(
+					// File-name link: <a> without the boolean `download` attribute.
+					'/(<a(?:(?!\sdownload[\s>])[^>])*?>)(.*?)(<\/a>)/',
+					// Download button: <a ... download ...>.
+					'/(<a[^>]*\sdownload[^>]*?>)(.*?)(<\/a>)/',
+				);
 			default:
 				return null;
 		}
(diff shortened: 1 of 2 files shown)


mikachan (approved):
This looks good to me, thanks @bph! It works as described when adding a File block to a template via the Save Theme flow ✨

Issue #830, opened 2026-04-30
Feature: Add wp cbt save CLI command


bph:
Part of #828 (which fulfills #688). Depends on the `CBT_Theme_Save` service refactor tracked in #828.

## Goal

Expose the `CBT_Theme_Save` service as a WP-CLI subcommand so developers can persist Editor changes to disk from the terminal.

## Scope

**New file:** `includes/cli/class-cbt-cli-command.php`

**New class:** `CBT_CLI_Command` extending `WP_CLI_Command`, with a `save` method.

**Registration** (in `includes/class-create-block-theme.php`):

```php
if ( defined( 'WP_CLI' ) && WP_CLI ) {
    require_once plugin_dir_path( __DIR__ ) . 'includes/cli/class-cbt-cli-command.php';
    WP_CLI::add_command( 'cbt', 'CBT_CLI_Command' );
}
```

## CLI surface

```
wp cbt save [--fonts] [--templates] [--styles] [--patterns]
            [--all] [--only-modified-templates] [--dry-run]
```

- Flags are kebab-case publicly; the command translates internally to the `saveFonts` / `saveTemplates` / etc. keys the service expects
- `--all` is a convenience for "save everything"
- `--only-modified-templates` maps to `processOnlySavedTemplates`
- `--dry-run` reports what would change without writing

## Required behavior

- Guard with `wp_is_block_theme()` — `WP_CLI::error()` if not, mirroring the plugin UI gate documented in AGENTS.md
- `WP_CLI::error()` if no save flag is passed (no silent no-op success)
- All user-facing strings i18n'd via `__( ..., 'create-block-theme' )`
- PHPDoc with shortdesc, full `## OPTIONS`, and at least two `## EXAMPLES`
- Translate `WP_Error` from the service into `WP_CLI::error()`

## Tests

PHPUnit tests against the command class (the service itself is covered in its own issue):

- Flag normalization (kebab-case → service options)
- Guard fires on non-block themes
- Empty invocation produces an error

## Out of scope

- `wp cbt export` — separate issue tracked in #828
- Behat features — not in the plugin's infrastructure today
- Any other CBT operation as a CLI command

## Acceptance

- Manual smoke test: install plugin in a Studio site, run each flag combination against a block theme
- `wp help cbt save` displays the documented options and examples
- `lint:php`, `test:php`, `test:unit` all green

KarunyaChavan:
[...]

Issue #828, opened 2026-04-30
[Tracking] WP-CLI support: phased implementation


bph:
Tracking issue for adding WP-CLI support to Create Block Theme, originally requested in #688.

## Background

Issue #688 proposes WP-CLI commands so developers can persist Editor changes to disk from the terminal — useful for git workflows and automation. @bacoords prototyped a `wp cbt save` command on a separate branch (https://github.com/bacoords/create-block-theme/tree/add/cli-export-command) but did not open a PR; the prototype acknowledges duplicated logic from the REST API.

This tracking issue proposes a four-PR sequence that ships the CLI as a thin layer over a properly extracted service layer, so REST and CLI share a single implementation.

## Approach

For each of `save` and `export`:

1. **Refactor first** — extract the REST handler's orchestration into a service class (`CBT_Theme_Save`, `CBT_Theme_Export`) under `includes/create-theme/`. REST endpoint becomes a thin wrapper. Add PHPUnit coverage. No behavior change at the wire boundary.
2. **Add the CLI command** — register on the same service. Kebab-case flags, `wp_is_block_theme()` guard, `## EXAMPLES`, i18n.

Doing the refactor first means the test coverage we need anyway lands in a small, reviewable PR before any new surface area is added.

Variation, clone, child-theme, blank-theme, and reset commands are deliberately out of scope for v1 and can be added in follow-ups once the pattern is established.

## Sub-issues

- [ ] #829 — Extract `CBT_Theme_Save` service from `rest_save_theme`
- [ ] #830 — Add `wp cbt save` CLI command
- [ ] #831 — Extract `CBT_Theme_Export` service from `rest_export_theme`
- [ ] #832 — Add `wp cbt export` CLI command

## Out of scope (potential follow-ups)

- Variation / clone / blank / child theme creation as CLI commands
- Behat feature tests (the plugin has no Behat infrastructure today)
- Promoting silent filesystem failures to `WP_Error` (separate concern from the refactor)

## References

- Original feature request: #688
- Prototype branch: https://github.com/bacoords/create-block-theme/tree/add/cli-export-command
- WP-CLI Commands Cookbook: https://make.wordpress.org/cli/handbook/guides/commands-cookbook/

Open to discussion on scope, sequencing, and naming before any PRs are opened.

PR #833, opened 2026-05-05
Refactor: Extract CBT_Theme_Save service from rest_save_theme

### LINKED ISSUE #829: Refactor: Extract CBT_Theme_Save service from rest_save_theme
Part of #828 (which fulfills #688).

## Goal

Move the orchestration logic out of `rest_save_theme()` (in `includes/class-create-block-theme-api.php`) into a new service class. No behavior change at the REST boundary.

## Scope

**New file:** `includes/create-theme/theme-save.php`

**New class:** `CBT_Theme_Save` with one public static method:

```php
public static function run( array $options ): true|WP_Error;
```

**Behavior:** identical to the current `rest_save_theme()` orchestration — handle `saveFonts`, `saveTemplates` + `processOnlySavedTemplates`, `saveStyle`, `savePatterns`. Honor `is_child_theme()` for scope selection. Invalidate the theme cache once at the end. Propagate `WP_Error` from `CBT_Theme_Patterns::add_patterns_to_theme`.

**REST endpoint update:** `rest_save_theme()` becomes a thin wrapper — call `CBT_Theme_Save::run()`, return its `WP_Error` directly or wrap success in the existing `WP_REST_Response` shape (`status`, `message`). The URL, permissions, and response shape do not change.

## Bug fixes that fall out of the refactor

- The current code dereferences `$options['processOnlySavedTemplates']` without `isset()` — fix by normalizing all flags through a single `! empty()` check at the top of `run()`.
- Strict `true ===` flag comparisons are replaced with truthy checks, accepting `true`, `1`, `"true"` consistently.

## Tests (PHPUnit, in this PR)

- Each flag in isolation persists the right thing
- `saveTemplates` without `processOnlySavedTemplates` does not warn
- Scope selection for `'user'` / `'current'` / `'all'` across child-theme on/off
- `WP_Error` from patterns step propagates out of `run()`
- Empty options → no-op success
- Cache invalidation fires exactly once

## Out of scope

- Any change to the utility classes (`CBT_Theme_Fonts`, `CBT_Theme_Templates`, etc.)
- Any change to the REST URL, permissions, or response shape
- Promoting silent filesystem failures to `WP_Error`
- Any CLI work — that's a follow-up issue tracked in #828

## Acceptance

- All existing REST tests pass unchanged
- New PHPUnit tests cover the cases above
- `lint:php`, `test:php`, `test:unit` all green

bph:
## Summary

Lift the orchestration logic out of `rest_save_theme()` into a new
`CBT_Theme_Save::run()` static service, so the upcoming `wp cbt save`
CLI command (#830) can share the same code path. The REST endpoint
becomes a thin wrapper.

This is the first PR in the WP-CLI track tracked in #828.

## Changes

- **New:** `includes/create-theme/theme-save.php` —
  `CBT_Theme_Save::run( array $options ): true|WP_Error`. Verbatim
  port of the four save branches (fonts, templates, styles, patterns)
  + theme cache invalidation.
- **Modified:** `includes/class-create-block-theme-api.php` —
  `rest_save_theme()` shrunk from 50 lines to 14: call `run()`,
  return `WP_Error` directly, otherwise wrap success in the existing
  `WP_REST_Response` shape.
- **New:** `tests/test-theme-save.php` — 13 PHPUnit cases.

## Bug fixes that fall out of the refactor

The orchestration logic had two latent issues that get resolved by
the new `normalize_flags()` helper, which routes every flag through a
single `! empty()` check:

1. `processOnlySavedTemplates` was previously dereferenced without an
   `isset()` guard, raising an `Undefined index` notice when
   `saveTemplates` was passed without it. The current Site Editor UI
   always sends both flags, so the notice was only reachable from
   non-JS callers — but the upcoming CLI is exactly that.
2. Flags were compared with strict `true ===`, rejecting `1`,
   `"true"`, `"1"`. This is fine for a typed JS client and brittle
   for CLI and third-party REST callers.

## Scope guarantees

No wire-level behavior change at the REST boundary:

- URL, HTTP method, permission callback unchanged.
- Response shape unchanged (`{ status: 'SUCCESS', message: 'Theme Saved.' }`).
- `WP_Error` from the patterns step is still returned as-is — the
  REST framework converts it to a JSON error response with the same
  status code as before.

The looser truthy semantics from the bug fix are isolated to the
`/save` endpoint. Sibling spots in the same file
(`rest_create_variation`, `rest_reset_theme`) keep their existing
strict-`===` checks; those endpoints already had `isset()` guards
and never produced a notice. If we want consistency across the file,
that's worth a separate small PR.

## Test plan

PHPUnit (added in this PR):

- [x] `test_run_with_empty_options_returns_true`
- [x] `test_run_without_options_does_not_invoke_any_save_step`
- [x] `test_save_templates_without_process_only_flag_does_not_warn` —
      regression test for the `processOnlySavedTemplates` notice.
      Matcher narrowed to the key name so unrelated downstream notices
      do not fail it.
- [x] `test_truthy_save_fonts_values_trigger_persist` (data provider:
      `true`, `1`, `"1"`, `"true"`)
- [x] `test_falsy_save_fonts_values_skip_persist` (data provider:
      `false`, `0`, `""`, `null`)
- [x] `test_missing_save_fonts_key_skips_persist`
- [x] `test_run_propagates_wp_error_from_patterns_step`

Manual smoke (Site Editor → Save Changes panel) covering the four
branches and the `WP_Error` path:

- [x] Save Fonts only — activated font moves into theme.
- [x] Save Templates with **Process Only Modified** on — only
      modified templates written.
- [x] Save Templates with **Process Only Modified** off — all
      templates written, no `Undefined index` notice in the PHP
      error log.
- [x] Save Style Changes only — `theme.json` updated, user-side
      global styles cleared.
- [x] Save Patterns with a name conflict — UI shows the
      `pattern_already_exists` snackbar error from `WP_Error`.

Cache invalidation (`wp_get_theme()->cache_delete()`) is not asserted
in PHPUnit; observing it would require mocking statics, which isn't
worth a new test framework just for one assertion.

## Out of scope

- Any change to the underlying utility classes (`CBT_Theme_Fonts`,
  `CBT_Theme_Templates`, `CBT_Theme_JSON`, `CBT_Theme_Patterns`,
  `CBT_Theme_Styles`).
- The CLI command itself (#830).
- Extracting the other REST endpoints (variation/reset) — same
  pattern is there but the WP-CLI track only needs save and export.
- Promoting silent filesystem failures to `WP_Error` — separate
  concern, intentionally not broadened here.

## Pre-existing bug noticed during testing (separate)

The strict error handler in the regression test surfaced an undefined
property access at `theme-templates.php:233` (`$template->media`
read without a property check). Not introduced by this PR; worth a
small follow-up issue.

Refs #828
Closes #829


bph:
@Scruffian would appreciate a review here when you have a moment — this is the first refactor in the WP-CLI track (#828) and unblocks #830. No wire-level behavior change; the REST endpoint keeps the same URL, method, permission callback, and response shape.

Just pushed [dbc140a](https://github.com/WordPress/create-block-theme/commit/dbc140a) addressing the Copilot review comment about the docblock — the previous wording implied a REST framework sanitize layer that doesn't actually exist for this route (`/save` has no `args` with `sanitize_callback`), so the docblock now states plainly that callers are responsible for sanitizing non-flag options.

Issue #838, opened 2026-05-06
Foundation + Color tab fully wired


bph:
Part of #836. First feature PR after the endpoint lands. Wires the modal's Color tab end-to-end so users can persist palette / gradient / duotone / default-toggle changes to `theme.json`. Other tabs continue rendering but their Update is a no-op until subsequent PRs.

Depends on #837.

### Files

- `src/resolvers.js` — add `postUpdateThemeSettings(payload)` mirroring `postUpdateThemeMetadata`.
- `src/editor-sidebar/edit-theme-settings-modal.js` — lift Color-tab state to modal level, wire `handleUpdateClick` to call the resolver, add loading / success-notice / error-notice / refresh.
- `src/plugin-styles.scss` — any save-state styling adjustments (loading indicator on the footer button).

### Behavior

- Update button enabled when there are pending changes; disabled + spinner during save.
- Success: surface `createSuccessNotice`, refresh the theme entity record (panels' `useEffect` resets local state from new server data automatically). Modal stays open by default.
- Error: surface `createErrorNotice` with the `WP_Error` message.

### State lifting

Modal owns:
- `colorSettings` (defaults flags + custom flags + link)
- `palette` / `gradients` / `duotone` arrays

Passed to `ColorSettingsPanel` / `PalettePanel` / `GradientPanel` / `DuotonePanel` as `value` + `onChange`. Tabs other than Color keep their existing local state for now.

## Acceptance criteria

- [x] Edit any Color sub-panel value, click Update — value persists to theme.json on disk.
- [ ] Reload editor — change is reflected.
- [ ] Failure (e.g. 500) surfaces an error notice; modal stays open.
- [ ] Update button shows loading state during save and is disabled when there are no pending changes.
- [ ] Other tabs (Dimensions / Typography / Shadows / Templates) still render and do not error.
- [ ] e2e smoke test for the Color save path (if e2e infra is in place per #811).

## Out of scope

- Other tabs' persistence (subsequent PRs, see #836).
- WP-CLI command.
- Documentation in `readme.txt` — fold into this PR if user-visible documentation is desired before further tabs land; otherwise the final per-tab PR.

## References

- Tracking issue: #836
- Endpoint sub-issue: #837
- Pattern mirrored for save UX: `metadata-editor-modal.js`.

Issue #836, opened 2026-05-06
[Tracking] Edit Theme Settings: phased implementation


bph:
Tracking issue for adding an Edit Theme Settings modal to Create Block Theme.

## Why this is important
Building a block theme in the editor should feel seamless — design in the browser, save back to clean theme files. CBT delivers most of that, but the settings layer (palette, fonts, sizes, shadows, units, custom templates) still requires editing theme.json by hand. Edit Theme Settings closes that gap inside the same modal pattern as Edit Theme Metadata.

CBT is not a niche developer tool:

- It's the recommended tool in the [theme developer learning pathway](https://learn.wordpress.org/) on Learn WordPress.
- Theme design teams use it to build themes in the editor.
- Agencies use it to sync client customizations back to version control.
- It's the centerpiece of the Editor + GitHub + Playground workflow being demonstrated at WordCamps and Meetups.
- Independent theme builders like [Anders Norén](https://www.andersnoren.se/) have added it to their workflow.

## Protoype

@henriqueiamarino has built a prototype of the modal with five tabs (Color, Dimensions, Typography, Shadows, Templates) on the [`edit-theme-settings`](https://github.com/henriqueiamarino/create-block-theme/tree/edit-theme-settings) branch. The tabs render and reflect data from the active theme's `theme.json`, but the Update button is not wired up yet — the prototype is a UI demonstration of the proposed feature shape, not a working save flow.

**Try it in Playground:** [open the prototype](https://playground.wordpress.net/?blueprint-url=https://raw.githubusercontent.com/henriqueiamarino/create-block-theme/edit-theme-settings/.wordpress-org/blueprints/preview.json) (Twenty Twenty-Four + the prototype plugin pre-installed; "Edit Theme Settings" sits under the CBT sidebar with the cog icon).

## Background

The existing "Edit Theme Metadata" modal lets users persist `style.css` / README / screenshot changes from the editor. "Edit Theme Settings" extends that pattern to `theme.json` — letting theme authors curate the theme's color palette, font families, spacing presets, custom templates, and so on without leaving the Site Editor.

The remaining work is server-side persistence (REST endpoint + service), state lifting in the modal, save UX (loading / notices / refresh), and splitting the feature across reviewable PRs.

## Approach

Foundation first, then features:

[...]

PR #843, opened 2026-05-06
Add /theme-settings REST endpoint + CBT_Theme_Settings_Save service

### LINKED ISSUE #837: REST endpoint + CBT_Theme_Settings_Save service
Part of #836. Server-side foundation for the Edit Theme Settings feature. UI is unaffected by this PR — the prototype's Update button remains a no-op until PR 1 wires the resolver.

## Scope

Add a REST endpoint and service that persist a partial-`theme.json` payload to the active theme's `theme.json` on disk. No JavaScript changes.

### Files

- `includes/class-create-block-theme-api.php` — register `POST /create-block-theme/v1/theme-settings`, add `rest_save_theme_settings` handler.
- `includes/create-theme/theme-settings-save.php` — **new** `CBT_Theme_Settings_Save` service class with `run( array $payload )`.
- `includes/class-create-block-theme-loader.php` — load the new service file.
- `tests/phpunit/tests/test-theme-settings-save.php` — **new** PHPUnit suite.

### Endpoint contract

- Method: `POST /create-block-theme/v1/theme-settings`
- Capability: `edit_theme_options`
- Request body (JSON; only present keys are written; missing keys leave theme.json untouched):

```json
{
  "settings": {
    "color": { "defaultPalette": false, "palette": [], "gradients": [], "duotone": [] },
    "spacing": { "spacingSizes": [], "units": ["px","rem"] },
    "typography": { "fontFamilies": [], "fontSizes": [] },
    "layout": {}
  },
  "templateParts": [],
  "customTemplates": [],
  "removedShadowDefaults": ["natural", "sharp"]
}
```

- Response: `{ "status": "SUCCESS", "theme_json": <merged theme.json array> }`. Errors return `WP_Error`.

### Service responsibilities

1. Validate payload shape (top-level allowlist; type checks for known sub-keys). Reject unknown top-level keys.
2. Sanitize strings (slugs, names, CSS values via `sanitize_text_field` and dedicated helpers). Use `wp_validate_boolean()` for flags.
3. Read current theme.json via `CBT_Theme_JSON_Resolver::get_theme_file_contents()`.
4. Deep-merge `settings.*` (replace at leaf); array-replace for `customTemplates` and `templateParts`. Create missing parent keys cleanly (e.g., theme without prior `settings.color` accepts a new palette).
5. Reify `removedShadowDefaults` into `settings.shadow.defaultPresets: false` + kept core shadows re-registered as `settings.shadow.shadow: [...]`.
6. Write via `CBT_Theme_JSON_Resolver::write_theme_file_contents()`.
7. Return the merged theme.json.

## Acceptance criteria

- [x] Endpoint registered, capability-gated, returns merged theme.json on success.
- [x] Partial payload (only `settings.color.palette`) leaves other keys untouched.
- [x] Empty palette array clears the palette (distinct from missing key).
- [x] Theme without prior `settings.color` accepts a new palette without errors.
- [x] `removedShadowDefaults: ["natural"]` produces a theme.json with `settings.shadow.defaultPresets: false` and the other 4 core shadows re-registered.
- [x] Shadow reification is idempotent — running the same payload twice produces the same theme.json.
- [x] Malformed payload returns `WP_Error` with a 400-class status.
- [x] PHPUnit covers each acceptance criterion.

## Out of scope

- JS resolver, modal state lifting, save UX (PR 1 — tracked separately).
- Wiping user customizations on save — explicitly *not* done in v1 (see decision in #836).
- WP-CLI command (separate effort).

## References

- Tracking issue: #836
- Pattern mirrored: `CBT_Theme_Save::run()` in `includes/create-theme/theme-save.php`.
- Writer used: `CBT_Theme_JSON_Resolver::write_theme_file_contents()` in `includes/create-theme/resolver_additions.php`.

bph:
Server-side foundation for the upcoming **Edit Theme Settings** modal. UI is unaffected by this PR — the prototype's Update button remains a no-op until the resolver is wired up in a follow-up.

Closes #837. Part of #836.

## What this adds

A new REST endpoint that accepts a partial-`theme.json` payload from the modal, deep-merges it into the active theme's `theme.json`, and writes the result.

- **Route:** `POST /create-block-theme/v1/theme-settings`
- **Capability:** `edit_theme_options`
- **Service:** `CBT_Theme_Settings_Save` in `includes/create-theme/theme-settings-save.php`. Pure helpers (`validate`, `sanitize`, `merge`, `reify_shadow_removals`) are public for unit testing.

### Payload contract

```json
{
  "settings": { "color": { }, "spacing": { }, "typography": { }, "layout": { } },
  "customTemplates": [],
  "templateParts": [],
  "removedShadowDefaults": ["natural", "sharp"]
}
```

- Only present keys are written. Missing keys leave `theme.json` untouched.
- `settings.*` is deep-merged (leaves replace, lists replace, missing parent keys are created).
- `customTemplates` / `templateParts` are full-list replace (the modal owns the canonical list per the prototype's UI).
- `removedShadowDefaults` is the only **operational** key. It's reified server-side into `settings.shadow.defaultPresets: false` plus the kept WP core shadow defaults re-registered under `settings.shadow.presets`. Centralizes WP-core-defaults knowledge in PHP and keeps the client-side resolver dumb.

### Response

```json
{ "status": "SUCCESS", "theme_json": { } }
```

Returning the merged theme.json lets the client refresh its theme record without a separate fetch.

## Decisions worth flagging

- **No user-customization wipe.** Unlike `CBT_Theme_Save` (whose purpose is promoting user changes into theme files), this endpoint edits the theme's offering. Existing user styles can continue to override what's seen in the editor; users can clear them via Site Editor → Styles → Reset. Open to revisit if testing shows confusion. See #836 for the full reasoning.
- **Shadow reification on the server.** The alternative (client expands `removedShadowDefaults` before sending) would push WP-core-defaults knowledge into JS. Server-side keeps the contract simpler and mirrors the plugin's PHP-heavy style.
- **Validation is shape-only.** Top-level keys are an allowlist; sub-keys under `settings` are not enumerated against the full theme.json schema. Keeps the endpoint forward-compatible with new theme.json keys without requiring a service bump.

## Tests

16 new PHPUnit tests in `tests/test-theme-settings-save.php` covering:

- Validation: known/unknown keys, type mismatches per top-level field
- Sanitization: tags stripped from string leaves, booleans/numbers preserved, camelCase keys preserved
- Merge: partial leaves siblings untouched, empty array clears, missing parent keys created, list replace, operational key dropped
- Shadow reification: `defaultPresets: false` + kept defaults re-registered, idempotent across repeat calls, user customs preserved

Run with `npm run test:unit:php:base`. Full suite (110 tests) passes; lint clean.

## Test plan

- [x] PHPUnit `Test_CBT_Theme_Settings_Save` (16 tests) green
- [x] Full PHP test suite (110 tests) green
- [x] `composer run-script lint` clean
- [x] Smoke test from a REST client (Postman / cURL) against the endpoint with a small payload
- [x] Confirm endpoint round-trips on a theme without a prior `settings.color` key

## Out of scope

- JS resolver, modal state lifting, save UX (next PR — see #838)
- Wiping user customizations (decided against for v1)
- WP-CLI command (parallels #828)

### DIFF
--- a/includes/create-theme/resolver_additions.php
+++ b/includes/create-theme/resolver_additions.php
@@ -137,11 +137,14 @@ function cbt_augment_resolver_with_utilities() {
 			$theme_json = wp_json_encode( $theme_json_data, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE );
 			$target     = static::get_file_path_from_theme( 'theme.json' );
 
-			// Atomic write: write to a sibling temp file, then rename into place.
-			// If the write or rename fails partway, the original theme.json is
-			// untouched. `file_put_contents` would truncate-then-write, leaving
-			// a corrupt file on interruption.
-			$tmp = $target . '.tmp';
+			// Atomic write with a request-unique temp file: write to a
+			// per-request sibling temp file, then rename into place. A
+			// per-request name prevents two concurrent saves from clobbering
+			// each other's staging payload (a shared `theme.json.tmp` is unsafe
+			// — request A could rename request B's truncated contents, or one
+			// could unlink the other's temp file mid-write). Each request
+			// cleans up only its own temp file on failure.
+			$tmp = $target . '.' . uniqid( '', true ) . '.tmp';
 			// Suppress warnings so a permission/disk error returns false cleanly
 			// rather than emitting a PHP warning that may be promoted to an
 			// exception. Callers must check the boolean return value.
--- a/includes/create-theme/resolver_additions.php
+++ b/includes/create-theme/resolver_additions.php
@@ -135,8 +135,13 @@ function cbt_augment_resolver_with_utilities() {
 
 		public static function write_theme_file_contents( $theme_json_data ) {
 			$theme_json = wp_json_encode( $theme_json_data, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE );
-			file_put_contents( static::get_file_path_from_theme( 'theme.json' ), $theme_json );
+			// Suppress warnings so a permission/disk error returns false cleanly
+			// rather than emitting a PHP warning that may be promoted to an
+			// exception. Callers must check the boolean return value.
+			// phpcs:ignore WordPress.PHP.NoSilencedErrors.Discouraged
+			$bytes = @file_put_contents( static::get_file_path_from_theme( 'theme.json' ), $theme_json );
 			static::clean_cached_data();
+			return false !== $bytes;
 		}
 
 		public static function write_user_settings( $user_settings ) {
--- a/includes/create-theme/resolver_additions.php
+++ b/includes/create-theme/resolver_additions.php
@@ -135,13 +135,34 @@ function cbt_augment_resolver_with_utilities() {
 
 		public static function write_theme_file_contents( $theme_json_data ) {
 			$theme_json = wp_json_encode( $theme_json_data, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE );
+			$target     = static::get_file_path_from_theme( 'theme.json' );
+
+			// Atomic write: write to a sibling temp file, then rename into place.
+			// If the write or rename fails partway, the original theme.json is
+			// untouched. `file_put_contents` would truncate-then-write, leaving
+			// a corrupt file on interruption.
+			$tmp = $target . '.tmp';
 			// Suppress warnings so a permission/disk error returns false cleanly
 			// rather than emitting a PHP warning that may be promoted to an
 			// exception. Callers must check the boolean return value.
 			// phpcs:ignore WordPress.PHP.NoSilencedErrors.Discouraged
-			$bytes = @file_put_contents( static::get_file_path_from_theme( 'theme.json' ), $theme_json );
+			$bytes = @file_put_contents( $tmp, $theme_json );
+			if ( false === $bytes ) {
+				return false;
+			}
+
+			// phpcs:ignore WordPress.PHP.NoSilencedErrors.Discouraged
+			if ( ! @rename( $tmp, $target ) ) {
+				// phpcs:ignore WordPress.PHP.NoSilencedErrors.Discouraged
+				@unlink( $tmp );
+				return false;
+			}
+
 			static::clean_cached_data();
-			return false !== $bytes;
+			// Bust the WP-level theme cache too — `clean_cached_data()` only
+			// clears the JSON resolver's caches, not `wp_get_theme()`.
+			wp_get_theme()->cache_delete();
+			return true;
 		}
 
 		public static function write_user_settings( $user_settings ) {
--- a/includes/create-theme/resolver_additions.php
+++ b/includes/create-theme/resolver_additions.php
@@ -135,7 +135,14 @@ function cbt_augment_resolver_with_utilities() {
 
 		public static function write_theme_file_contents( $theme_json_data ) {
 			$theme_json = wp_json_encode( $theme_json_data, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE );
-			$target     = static::get_file_path_from_theme( 'theme.json' );
+			// `wp_json_encode` returns false if the data contains an
+			// unencodable value (resources, NAN/INF after wrapping, etc.).
+			// Bail before opening a temp file so theme.json is never
+			// truncated or replaced with empty content.
+			if ( false === $theme_json ) {
+				return false;
+			}
+			$target = static::get_file_path_from_theme( 'theme.json' );
 
 			// Atomic write with a request-unique temp file: write to a
 			// per-request sibling temp file, then rename into place. A
--- a/includes/create-theme/theme-settings-save.php
+++ b/includes/create-theme/theme-settings-save.php
@@ -4,15 +4,21 @@
  *
  * Persists Edit Theme Settings modal payloads to the active theme's theme.json.
  * Accepts a partial-`theme.json` payload (only the keys the user edited),
- * deep-merges it into the existing file, and writes the result. Reifies the
- * `removedShadowDefaults` operational key into the standard
- * `settings.shadow.defaultPresets` + `settings.shadow.presets` shape.
+ * deep-merges it into the existing file using JSON Merge Patch semantics
+ * (RFC 7396), and writes the result. Reifies the `removedShadowDefaults`
+ * operational key into the standard `settings.shadow.defaultPresets` +
+ * `settings.shadow.presets` shape.
  *
  * Callers must enforce capability checks (`edit_theme_options`) before
  * invoking `run()`. The service performs payload-shape validation and
  * sanitization but does not authenticate.
  *
+ * Known limitation: concurrent edits are last-write-wins. Two clients with
+ * the modal open simultaneously will silently overwrite each other. ETag /
+ * If-Match handling can be added if this becomes a problem in practice.
+ *
  * @package Create_Block_Theme
+ * @see https://datatracker.ietf.org/doc/html/rfc7396 JSON Merge Patch
  */
 class CBT_Theme_Settings_Save {
 
@@ -26,6 +32,13 @@ class CBT_Theme_Settings_Save {
 	/**
 	 * Keys whose string values are user-facing labels and benefit from
 	 * `sanitize_text_field()` (HTML stripping, whitespace normalization).
+	 *
+	 * This is an allowlist. Adding a new label-class field to theme.json
+	 * upstream means adding it here — otherwise it falls through to
+	 * `wp_kses_no_null` and HTML will not be stripped. Trade-off accepted:
+	 * the inverse rule ("strip HTML by default, allowlist CSS-value fields")
+	 * silently corrupts CSS for new fields, which is worse than missing a
+	 * label sanitization update.
 	 */
 	const TEXT_FIELD_KEYS = array( 'name', 'title', 'label', 'description' );
 
@@ -259,11 +272,28 @@ class CBT_Theme_Settings_Save {
 	}
 
 	/**
-	 * Deep-merge $payload into $current. At every level, associative-array
-	 * values are merged recursively; lists, scalars, and empty arrays replace
-	 * the existing value. Missing parent keys are created.
+	 * Deep-merge $payload into $current using JSON Merge Patch (RFC 7396)
+	 * semantics, with one PHP-imposed accommodation.
 	 *
-	 * The operational key `removedShadowDefaults` is dropped here — it's
+	 * Rules:
+	 *
+	 * - **`null` deletes the key.** Sending `{"settings":{"color":{"custom":null}}}`
+	 *   removes `settings.color.custom` from the result. Deleting a missing
+	 *   key is a no-op.
+	 * - **Empty object (`{}`) is a no-op.** Per RFC 7396, an empty object means
+	 *   "no change at this key." PHP cannot distinguish a JSON `{}` from a JSON
+	 *   `[]` after `json_decode(..., true)` (both become empty PHP arrays), so
+	 *   we use a heuristic: if the existing value at this key is a list, an
+	 *   empty payload value clears that list; otherwise it is treated as a
+	 *   no-op. This means clients that intend to clear a list MUST already
+	 *   know the field is list-typed (which is how the modal is built).
+	 * - **Associative-object values are merged recursively.** Leaves replace.
+	 * - **List values replace wholesale** — RFC 7396 does not support
+	 *   per-element list patching. To remove one palette entry, send the full
+	 *   new palette.
+	 * - **Missing parent keys are created** when assigning into them.
+	 *
+	 * The operational key `removedShadowDefaults` is skipped here — it's
 	 * handled separately by `reify_shadow_removals()`.
 	 *
 	 * @param array $current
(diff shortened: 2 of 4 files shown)


bph:
## The smoke test via cURL

### Add a color to the theme palette
```
❯ curl -sS -u "$APP_USER:$APP_PASS" \
  -X POST "$WP_URL/wp-json/create-block-theme/v1/theme-settings" \
  -H "Content-Type: application/json" \
  -d '{
    "settings": {
      "color": {
        "palette": [
          { "slug": "brand", "name": "Brand", "color": "#0066ff" }
        ]
      }
    }
  }' | jq .status
"SUCCESS"
```

Screenshot 2026-05-07 at 10 09 44

### Remove default shadow preset "Natural" "Sharp"

```
❯ curl -sS -u "$APP_USER:$APP_PASS" \
  -X POST "$WP_URL/wp-json/create-block-theme/v1/theme-settings" \
  -H "Content-Type: application/json" \
  -d '{
    "removedShadowDefaults": ["natural", "sharp"]
  }' | jq '.theme_json.settings.shadow'
```
**Results**
```
{
  "defaultPresets": false,
  "presets": [
    {
      "name": "Deep",
      "slug": "deep",
      "shadow": "12px 12px 50px rgba(0, 0, 0, 0.4)"
    },
    {
      "name": "Outlined",
      "slug": "outlined",
      "shadow": "6px 6px 0px -3px rgb(255, 255, 255), 6px 6px rgb(0, 0, 0)"
    },
    {
      "name": "Crisp",
      "slug": "crisp",
      "shadow": "6px 6px 0px rgb(0, 0, 0)"
    }
  ]
}
```

Screenshot 2026-05-07 at 10 41 28

### Random Error Message
```
❯ curl -sS -u "$APP_USER:$APP_PASS" \
  -X POST "$WP_URL/wp-json/create-block-theme/v1/theme-settings" \
  -H "Content-Type: application/json" \
  -d '{ "totallyMadeUp": true }' | jq .
```
**Result:**
```
{
  "code": "cbt_invalid_payload",
  "message": "Unknown top-level key: totallyMadeUp",
  "data": {
    "status": 400
  }
}
```

bph:
## Review feedback addressed

Pushed `edf4431` with the three high-value fixes:

- **Context-aware `sanitize()`** — `sanitize_text_field` is now applied only to label-class fields (`name`/`title`/`label`/`description`); slug fields go through `sanitize_key`; everything else (CSS values like `shadow`, `color`, `gradient`) through `wp_kses_no_null`, which preserves whitespace, commas, and parens. Fixes silent corruption of multi-segment shadows like `6px 6px 0px -3px rgb(255, 255, 255), 6px 6px rgb(0, 0, 0)`.
- **Slug validation in `validate()`** — `removedShadowDefaults` slugs and the `name` field on `customTemplates`/`templateParts` must now survive `sanitize_key()` unchanged and be non-empty. Rejects malformed input early.
- **Write-failure handling** — `run()` now returns `WP_Error('cbt_write_failed', status: 500)` when the theme.json write fails. `write_theme_file_contents()` returns `bool` to support this. Previously a write failure was silently swallowed and the response said SUCCESS.

7 new tests (CSS round-trips for shadow/gradient/postTypes, slug rejections, write-failure integration). 23/23 in the new suite, 117/117 overall, lint clean.

## Soft-skipped items (with rationale)

A few suggestions weren't applied in this PR — calling them out here for visibility:

- **No `args` schema on the route.** Existing CBT endpoints (`/save`, `/update`, `/clone`, `/create-variation`, etc.) all register without `args` schemas. Adding it just to this one would create inconsistency. Suggest a follow-up that retrofits all endpoints — happy to open it.
- **No `defined('ABSPATH') || exit` guard.** None of the other files under `includes/create-theme/` have this guard either (`theme-save.php`, `theme-json.php`, `theme-utils.php`, etc.). Same consistency argument — fold into a plugin-wide cleanup if desired.
- **No `class_exists` guard around the class declaration.** Same — none of the sibling classes use one.
- **`status: "SUCCESS"` response shape.** Mirrors the existing convention in `rest_save_theme`, `rest_update_theme`, `rest_create_blank_theme`, etc. Diverging here would be the bigger inconsistency.
- **No payload size cap.** WP server-level `post_max_size` and PHP's `memory_limit` already gate this. Belt-and-braces could be added but feels out of scope for the foundation PR.

These all feel like they belong in a separate "tighten the CBT REST API conventions" PR rather than mixed into the foundation work for Edit Theme Settings.

bph:
## Round 2 review feedback addressed

Pushed \`2e6b5ba\`:

- **Entry-level structural validation** — \`customTemplates\` entries now require non-empty \`name\` + \`title\`. \`templateParts\` entries require non-empty \`name\` + \`area\`. \`name\` continues to require slug format.
- 4 new tests covering the rejection paths (missing title, missing area, missing name, empty required key) plus a positive case for a complete templatePart entry.

All-tests count: 28 in this suite (was 23), 122 overall, lint clean.

## Cross-cutting items moved to a tracking issue

Opened #844 to track the items that apply uniformly to all CBT endpoints rather than just this one:

- \`args\` schemas on REST routes
- \`defined('ABSPATH') || exit\` guards
- \`class_exists\` guards
- Response shape audit (\`status: "SUCCESS"\` consistency)
- \`WP_REST_Request\` type hints
- Optional payload size caps
- Cross-endpoint entry validation

Each is small and reviewable on its own; piecemeal-fixing them only on new files would create the inconsistency the reviewer is trying to prevent. #844 is the right venue.

The architectural suggestion to split \`run()\` into \`prepare()\` + \`persist()\` for testability is debatable — the public helpers (\`validate\`, \`sanitize\`, \`merge\`, \`reify_shadow_removals\`) already give the same testability benefit without changing the public API or forcing PR 1's resolver to call two methods. Happy to revisit if a clear pattern emerges across the plugin.

scruffian:
Some thoughts on this:

The merge almost implements JSON Merge Patch (RFC 7396), but PHP can’t distinguish {} from [] after json_decode — so the documented “empty arrays replace” rule means a frontend that sends {"settings": {}} (intending “no change”) wipes all settings. 

There's also no way to delete — to drop a single palette entry you have to send the whole palette. Workable, but it forces every modal interaction to be read-modify-write.

One option is to adopt merge-patch semantics (use null to delete, treat {} as “no-op”) and document it. An alternative is to invert the model and require the frontend to send full subtrees (no merge at all). The current middle ground is a bit confusing.

includes/create-theme/resolver_additions.php
```diff
@@ -135,8 +135,13 @@ public static function get_theme_file_contents() {
 
 		public static function write_theme_file_contents( $theme_json_data ) {
 			$theme_json = wp_json_encode( $theme_json_data, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE );
-			file_put_contents( static::get_file_path_from_theme( 'theme.json' ), $theme_json );
+			// Suppress warnings so a permission/disk error returns false cleanly
+			// rather than emitting a PHP warning that may be promoted to an
+			// exception. Callers must check the boolean return value.
+			// phpcs:ignore WordPress.PHP.NoSilencedErrors.Discouraged
+			$bytes = @file_put_contents( static::get_file_path_from_theme( 'theme.json' ), $theme_json );
```
scruffian: If this fails then we'll break everything. We might want to think about writing the file first with a temporary name, and then renaming.

includes/create-theme/resolver_additions.php
```diff
@@ -135,8 +135,13 @@ public static function get_theme_file_contents() {
...
 			static::clean_cached_data();
```
scruffian, line 171: `rest_save_theme` also calls `wp_get_theme()->cache_delete()` after writing, should we do that here too?

includes/create-theme/resolver_additions.php
scruffian: Two admin tabs with the modal open will silently overwrite each other — last write wins, no warning. Not sure how big of a problem that is in reality.

includes/create-theme/theme-settings-save.php
```diff
@@ -0,0 +1,394 @@
...
+		'customTemplates',
+		'templateParts',
+		'removedShadowDefaults',
+	);
+
+	/**
+	 * Keys whose string values are user-facing labels and benefit from
+	 * `sanitize_text_field()` (HTML stripping, whitespace normalization).
+	 */
+	const TEXT_FIELD_KEYS = array( 'name', 'title', 'label', 'description' );
```
scruffian, line 49: This could become a maintenance burden. If theme.json adds more key (say caption), it silently falls through to wp_kses_no_null and HTML is no longer stripped.

bph:
@scruffian Thanks Ben for the thorough review. 

Claude and I  pushed `980c89d` with the fixes. Walking through each comment:

## Merge semantics — adopted RFC 7396

Thanks for pointing out the partial merge-patch semantics but PHP's `json_decode(..., true)` indistinguishability between `{}` and `[]` was the footgun. Updated `merge()` to:

- **`null` deletes the key** — `{"settings":{"color":{"custom":null}}}` removes `settings.color.custom`. Deleting a missing key is a no-op.
- **Empty object is a no-op** — `{"settings":{}}` no longer wipes everything. Heuristic since PHP can't tell `{}` from `[]`: if the existing value at the key is a list, empty payload clears; otherwise no-op. This preserves the "send empty list to clear" semantics for `palette: []`, `customTemplates: []`, etc., while closing the wipe-everything case for `settings: {}`.

Per-element list patching (your "drop a single palette entry" concern) isn't part of RFC 7396.  Arrays are replaced wholesale under the spec. To remove one palette entry the client still has to send the full new palette. I hope that's the right read-modify-write boundary for now. 

Cited RFC 7396 in the class docblock and added 6 new tests covering the new semantics.

## Atomic write

Fixed. `write_theme_file_contents` now writes to `theme.json.tmp` and `rename()`s into place. The original file is untouched if either step fails.

## `wp_get_theme()->cache_delete()`

Added. Called after a successful rename. `clean_cached_data()` only clears the JSON resolver's caches; this also busts WP's theme cache to match `rest_save_theme`.

## Concurrent edits

Initially documented as a known limitation. **Followed up in `dd93c0e`** — see Round 4 below.

## TEXT_FIELD_KEYS maintenance

Real trade-off. Added a comment on the constant spelling out *why* this is an allowlist: the inverse rule ("strip HTML by default, allowlist CSS-value fields") silently corrupts CSS for any new CSS-like field that gets added to theme.json, which is worse than missing a label sanitization update. The maintenance cost is "add a slug to this list when theme.json adds a new label field" — visible in `git blame`, low-frequency.

---

## Round 4 — pushed `dd93c0e`

Three follow-up review comments addressed together since they share code paths.

### Shared `theme.json.tmp` was unsafe under concurrent saves

The atomic write was right in shape but wrong in detail — a fixed staging path means two concurrent requests can collide on the temp file and one can rename the other's truncated payload as the new theme.json. Fix: `write_theme_file_contents()` now uses a per-request `theme.json.<uniqid>.tmp` filename. Each request cleans up only its own temp file. No more shared-staging-path race.

### Concurrent read-modify-write silently dropped disjoint edits

The previous "last write wins" caveat for concurrent edits wasn't acceptable for a settings save — disjoint edits should compose. Fix: `run()` now holds an exclusive `flock()` on a sibling `theme.json.lock` around the read-merge-write block. Request B blocks until A finishes, then re-reads A's saved state and merges its own payload on top. Disjoint edits (A touches palette, B touches typography) now both persist.

Conflicting same-key edits remain last-write-wins — that's the genuine domain of If-Match/ETag, which would require a contract change (client tracks theme.json hash, server returns 409 on mismatch). Candidate for a follow-up once the modal lands and we see whether that case bites in practice. On distributed hosts where `flock()` isn't honored (some NFS configs) the serialization degrades to best-effort, documented in the class docblock.

If flock acquisition fails for any reason (rare), the endpoint surfaces `WP_Error('cbt_lock_failed', status 503)` so the client can retry rather than seeing silent success.

### Payload shape validation gap (`array_is_list()`)

Right call. The internal `is_list()` helper was used inside `merge()` to route the merge but wasn't gating validation, so a non-empty JSON list passed where an object is required (e.g., `settings: [{"foo":1}]`) would slip through and land a malformed shape in theme.json. Fix: `validate()` now explicitly rejects:

- JSON lists in the `settings` object position
- JSON objects in the `customTemplates` / `templateParts` / `removedShadowDefaults` list positions

Empty arrays remain accepted for either (PHP can't tell them apart; `merge()` handles them correctly).

`merge()` also defensively normalizes lists via `array_values()` so any sparse-keyed array can't slip through and serialize as a JSON object.

---

133/133 tests pass, lint clean.

scruffian:
Thanks, looking good. A couple more things:

1. We don't check if `wp_json_encode` fails - if it does then we still write the empty contents to the theme.json, breaking it.
2. The theme.json.lock file lives in the theme directory and ships in exports (theme-settings-save.php:83). I think we should move it elsewhere so it doesn't send up shipping with the theme.

bph:
@scruffian Thanks Ben — both real catches. Pushed `1cc73d5`:

- **`wp_json_encode` failure**: writer now bails before opening a temp file if encoding returns `false`. Surfaces as `cbt_write_failed`.
- **Lockfile location**: moved to `get_temp_dir() . 'cbt-theme-settings-<md5(stylesheet)>.lock'` so theme exports don't bundle it.
- New test: writer leaves theme.json byte-for-byte unchanged on encoding failure.

134/134 tests pass, lint clean.

bph:
After additional review passes — pushed `b49eb1b`:

- **400 on non-array bodies**: handler now checks `is_array()` on `get_json_params()` and returns `cbt_invalid_payload` (400) for empty, null, and scalar bodies. Previously the service's `array $payload` type hint would fatal as a 500.
- **`removedShadowDefaults: null` is now a no-op**: replaced `array_key_exists` with `isset` so a null value at the operational key is treated as absent rather than reaching `reify_shadow_removals(..., null)` and TypeError-ing.
- Doc tweaks: corrected `validate()` docblock (extraction happens in `merge()`, not `validate()`) and added a note in `merge()` explaining that wholesale list-replace for `customTemplates`/`templateParts` is intentional under the modal-owns-canonical-list design.

4 new tests (empty body, scalar body, null body, null `removedShadowDefaults`). 44 in this suite, 138 overall, lint clean.

scruffian (approved):
Looks good, thanks for the updates.

bph:
Thank you @scruffian for your feedback and helping this PR over the finish line.

PR #850, opened 2026-06-16
Require super-admin for theme-modifying REST routes on multisite

mikachan:
On a multisite, themes are network-shared. This PR tightens the permission check on the 9 filesystem-mutating REST routes in `CBT_Theme_API` so that only super-admins can call them when the site is multisite. The read-only `/font-families` route is unchanged.

Also restores `DISALLOW_FILE_EDIT` / `DISALLOW_FILE_MODS` honouring that the plugin had prior to v2.1.2. Both constants now gate theme-file modifications, matching the behaviour of `wp-admin/theme-editor.php`.

To test, ensure the following tests pass:

```bash
  npm run test:unit:php -- --filter Test_Create_Block_Theme_Api

### DIFF
--- a/includes/class-create-block-theme-api.php
+++ b/includes/class-create-block-theme-api.php
@@ -508,48 +508,22 @@ class CBT_Theme_API {
 	/**
 	 * Permission check for filesystem-mutating REST routes.
 	 *
-	 * On multisite, themes are network-shared. Require super-admin to prevent
-	 * sub-site administrators from crossing the tenant boundary into the
-	 * shared `wp-content/themes/` directory.
+	 * Combines two WordPress Core primitives:
 	 *
-	 * Honors the WP hardening constants `DISALLOW_FILE_EDIT` and
-	 * `DISALLOW_FILE_MODS` via `file_mods_allowed()`, matching the behaviour
-	 * the plugin had pre-v2.1.2 and aligning with WP Core's theme-editor.
+	 *  - `current_user_can( 'edit_themes' )` — Core's canonical theme-file
+	 *    capability. Held by Administrators on single-site, super-admins on
+	 *    multisite (NOT sub-site admins), and automatically denied when
+	 *    `DISALLOW_FILE_EDIT` is defined. This single check covers the
+	 *    multisite tenant boundary and the `DISALLOW_FILE_EDIT` hardening
+	 *    that the plugin honoured pre-v2.1.2.
+	 *  - `wp_is_file_mod_allowed( 'create_block_theme_modify_theme' )` —
+	 *    Core's canonical file-modification gate. Handles `DISALLOW_FILE_MODS`
+	 *    and the `file_mod_allowed` filter (used by hosts / security plugins).
 	 *
-	 * @return bool
+	 * @return bool True when both checks pass.
 	 */
 	private function can_modify_theme() {
-		if ( ! $this->file_mods_allowed() ) {
-			return false;
-		}
-		if ( is_multisite() ) {
-			return is_super_admin();
-		}
-		return current_user_can( 'edit_theme_options' );
-	}
-
-	/**
-	 * Whether theme-file modifications are permitted by site configuration.
-	 *
-	 * Delegates the `DISALLOW_FILE_MODS` check (and the canonical
-	 * `file_mod_allowed` filter that hosts and security plugins use to disable
-	 * file modifications globally) to WordPress Core's `wp_is_file_mod_allowed()`.
-	 * `DISALLOW_FILE_EDIT` is checked explicitly because it is a separate
-	 * constant that Core's helper does NOT cover.
-	 *
-	 * The `cbt_file_mods_allowed` filter remains as a test seam — it can only
-	 * further restrict, never re-enable, the policy decided by core / the
-	 * constants.
-	 *
-	 * @return bool True when file modifications are allowed by site configuration.
-	 */
-	private function file_mods_allowed() {
-		if ( defined( 'DISALLOW_FILE_EDIT' ) && DISALLOW_FILE_EDIT ) {
-			$allowed = false;
-		} else {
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: includes/class-create-block-theme-api.php [hunk 1]
```diff
@@ -504,4 +504,45 @@ private function sanitize_theme_data( $theme ) {
...
+	 *
+	 * @return bool True if both DISALLOW_FILE_EDIT and DISALLOW_FILE_MODS are absent / false.
+	 */
+	private function file_mods_allowed() {
+		$allowed = ! ( defined( 'DISALLOW_FILE_EDIT' ) && DISALLOW_FILE_EDIT );
+		if ( $allowed ) {
+			$allowed = ! ( defined( 'DISALLOW_FILE_MODS' ) && DISALLOW_FILE_MODS );
+		}
+		return apply_filters( 'cbt_file_mods_allowed', $allowed );
+	}
```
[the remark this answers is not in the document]
scruffian: This looks like a legit issue.

mikachan:
Looks like these were all good suggestions - I've applied them all.

includes/class-create-block-theme-api.php
```diff
@@ -504,4 +504,41 @@ private function sanitize_theme_data( $theme ) {
...
+	 *
+	 * @return bool True when file modifications are allowed by configuration (DISALLOW_FILE_EDIT / DISALLOW_FILE_MODS) and not disabled by the cbt_file_mods_allowed filter.
+	 */
+	private function file_mods_allowed() {
+		$const_allowed = ! ( defined( 'DISALLOW_FILE_EDIT' ) && DISALLOW_FILE_EDIT );
+		if ( $const_allowed ) {
+			$const_allowed = ! ( defined( 'DISALLOW_FILE_MODS' ) && DISALLOW_FILE_MODS );
+		}
+		$filtered = (bool) apply_filters( 'cbt_file_mods_allowed', $const_allowed );
+		return $const_allowed && $filtered;
```
scruffian: This bypasses WordPress’s standard file_mod_allowed policy. A host or site disabling file changes via Core’s file_mod_allowed filter would still allow these theme-file REST writes. Core exposes this through wp_is_file_mod_allowed().

mikachan: Ah thanks, good catch. I think this should be fixed in https://github.com/WordPress/create-block-theme/pull/850/commits/0c6174ae55893dcaaa0093ea8655b88bd2c39f13. I updated it so `file_mods_allowed()` now delegates the `DISALLOW_FILE_MODS` check to `wp_is_file_mod_allowed( 'create_block_theme_modify_theme' )`.

includes/class-create-block-theme-api.php
```diff
@@ -142,7 +142,7 @@ public function register_rest_routes() {
 				'methods'             => WP_REST_Server::EDITABLE,
 				'callback'            => array( $this, 'rest_reset_theme' ),
 				'permission_callback' => function () {
-					return current_user_can( 'edit_theme_options' );
+					return $this->can_modify_theme();
```
scruffian: `reset-theme` actually doesn't modify any theme files, it only drops user customizations from the database, so technically these permissions are unnecessarily restrictive. However I think it's good to have consistency here and keep the permissions level the same.

includes/class-create-block-theme-api.php
```diff
@@ -504,4 +504,52 @@ private function sanitize_theme_data( $theme ) {
...
+	 * further restrict, never re-enable, the policy decided by core / the
+	 * constants.
+	 *
+	 * @return bool True when file modifications are allowed by site configuration.
+	 */
+	private function file_mods_allowed() {
+		if ( defined( 'DISALLOW_FILE_EDIT' ) && DISALLOW_FILE_EDIT ) {
+			$allowed = false;
+		} else {
+			$allowed = wp_is_file_mod_allowed( 'create_block_theme_modify_theme' );
```
scruffian: This catches `DISALLOW_FILE_EDIT` above and global `file_mod_allowed` filters, but it misses sites that disable core theme-file editing with a context-specific `file_mod_allowed` filter. Core’s `edit_themes` capability checks the `capability_edit_themes` context, so those sites would block the theme editor while these REST routes could still write to `wp-content/themes`. If this is meant to mirror Core’s theme-file policy, can we reuse `current_user_can( 'edit_themes' )` or the core `capability_edit_themes` context instead?

mikachan: Thanks! `current_user_can( 'edit_themes' )` makes sense here I think. Addressed in https://github.com/WordPress/create-block-theme/pull/850/commits/342f90f4baf602f95a3eed04e1b843cba3d487c0.

tests/test-class-create-block-theme-api.php
```diff
@@ -0,0 +1,179 @@
...
+		wp_set_current_user( $admin );
+
+		$request  = new WP_REST_Request( 'POST', '/create-block-theme/v1/save' );
+		$response = rest_get_server()->dispatch( $request );
+
+		$this->assertSame( 403, $response->get_status() );
+		$this->assertSame( 'rest_forbidden', $response->get_data()['code'] );
+	}
+
+	public function test_admin_landing_menu_not_registered_when_cannot_modify_theme() {
```
scruffian: Do we need to switch to a block theme for this test to be accurate -otherwise create_admin_menu and create_block_theme_sidebar_enqueue will just return false anyway.

mikachan: Yes! Thank you for spotting. This should be fixed in https://github.com/WordPress/create-block-theme/pull/850/commits/755413ee16999be149b548d015067f26ff5712da - the tests will now activate a blank block theme before asserting.

scruffian (approved):
Looking good now! Thanks for all the iterations 🚢

PR #852, opened 2026-06-16
Validate downloaded theme assets against extension and MIME allowlists

mikachan:
Add a per-sink extension allowlist and post-download MIME validation to the four places the plugin calls `download_url()` and writes the response to disk: media downloads from template/pattern content (`add_media_to_local`), font downloads from user global styles (`copy_font_assets_to_theme`), and the zip-export equivalents of both (`add_media_to_zip`, `add_activated_fonts_to_zip`). Also rejects multi-extension polyglot URLs (`evil.php.jpg`).

To test, ensure the following tests pass:

```bash
  npm run test:unit:php:base -- --filter "Test_Create_Block_Theme_(Media|Fonts|Zip)"

### CODE REVIEW

File: includes/create-theme/theme-media.php [hunk 1]
```diff
@@ -19,6 +32,121 @@ public static function get_media_folder_path_from_url( $url ) {
...
+	 *
+	 * @param string $tmp_file Local path to the downloaded file.
+	 * @param string $url      The originating URL (used to hint the basename).
+	 * @return bool True if the file's detected type is in the allowlist.
+	 */
+	public static function is_allowed_media_file( $tmp_file, $url ) {
+		if ( ! is_string( $tmp_file ) || ! file_exists( $tmp_file ) ) {
+			return false;
+		}
+		$allowed = array(
```
scruffian: This list is different from one in `is_allowed_media_url`, which allows svg, wmv avi, whereas this list only accepts image/svg+xml and video/x-msvideo. Default WordPress does not include SVG in allowed mimes, maps wmv to video/x-ms-wmv, and maps avi to video/avi, so those URLs will pass preflight and then fail post-download.

includes/create-theme/theme-zip.php
```diff
@@ -51,6 +51,12 @@ public static function add_activated_fonts_to_zip( $zip, $theme_json_string ) {
 				$font_dir         = wp_get_font_dir();
 				$font_face['src'] = (array) $font_face['src'];
```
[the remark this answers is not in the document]
mikachan: Should be fixed with https://github.com/WordPress/create-block-theme/pull/852/commits/5b22c72f49ceab892720c5f64b94bde2589d193e.

includes/create-theme/theme-media.php
mikachan: Good spot, thanks! I've fixed this in https://github.com/WordPress/create-block-theme/pull/852/commits/68f574cd851e2293dd9ca7b79681d5c2e941c1f2, by switching the media post-download check to magic-byte verification (the same as what we did for fonts). This removes the dependency on Core's mime registry, so SVG/WMV/AVI now work end-to-end.

scruffian:
I found 3 issues:

1.  The temp file is now deleted before ZipArchive has read it.
2. URL rewriting still uses the raw URL basename, so query-string slashes can produce broken or dangerous-looking asset references even though the download sink now uses the parsed path. For example http://example.com/cat.jpg?/evil.php now becomes cat.jpg for validation/download, but exported template markup points at /assets/images/evil.php.
3. is_allowed_media_file() does not verify that the downloaded bytes match the URL extension. e.g. a .jpg URL with SVG/MP4/WebM bytes passes and gets saved as .jpg.
'
I added unit tests to demonstrate the issues (but not fixes)

mikachan:
Thanks! These should all be fixed now.

1. `add_media_to_zip()` now reads the downloaded bytes into memory and unlinks the tmp file before adding via `addFromStringToTheme()`.
2. `make_relative_media_url()` now uses `basename(wp_parse_url($url, PHP_URL_PATH))` instead of raw `basename($url)`. The example `cat.jpg?/evil.php` now produces `/assets/images/cat.jpg` as expected.
3. `is_allowed_media_file()` now switches on the URL extension and requires the bytes to match that specific format. A .jpg URL with SVG/MP4/anything-else bytes is rejected so we never persist a file whose extension lies about its content.

includes/create-theme/theme-zip.php
```diff
@@ -243,11 +262,27 @@ static function add_media_to_zip( $zip, $media ) {
 
 			// If there was an error downloading the file, skip it.
 			// TODO: Implement a warning if the file is missing
-			if ( ! is_wp_error( $download_file ) ) {
-				$content_array  = file( $download_file );
-				$file_as_string = implode( '', $content_array );
-				$zip->addFromStringToTheme( $folder_path . basename( $url ), $file_as_string );
+			if ( is_wp_error( $download_file ) ) {
+				continue;
```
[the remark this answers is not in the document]
mikachan: Fixed in https://github.com/WordPress/create-block-theme/pull/852/commits/5b550bebdb893e804f1308edbabb18ddecc53267.

includes/create-theme/theme-media.php
```diff
@@ -133,15 +344,24 @@ public static function add_media_to_local( $media ) {
...
 
 			// TODO: implement a warning if the file is missing
-			if ( ! is_wp_error( $download_file ) ) {
-				$media_path = get_stylesheet_directory() . DIRECTORY_SEPARATOR . self::get_media_folder_path_from_url( $url );
-				if ( ! is_dir( $media_path ) ) {
-					wp_mkdir_p( $media_path );
-				}
-				rename( $download_file, $media_path . basename( $url ) );
+			if ( is_wp_error( $download_file ) ) {
+				continue;
```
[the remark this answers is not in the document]
mikachan: Fixed in https://github.com/WordPress/create-block-theme/pull/852/commits/5b550bebdb893e804f1308edbabb18ddecc53267.

includes/create-theme/theme-fonts.php
```diff
@@ -112,6 +112,133 @@ public static function make_filename_from_fontface( $font_face, $src, $src_index
...
+		if ( 12 === strlen( $head ) ) {
+			$version_le_bytes = array(
+				"\x00\x00\x01\x00", // 0x00010000
+				"\x01\x00\x02\x00", // 0x00020001
+				"\x02\x00\x02\x00", // 0x00020002
+			);
+			if ( in_array( substr( $head, 8, 4 ), $version_le_bytes, true ) ) {
+				return true;
+			}
+		}
```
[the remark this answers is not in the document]
mikachan: Fixed in https://github.com/WordPress/create-block-theme/pull/852/commits/a592c6a85643608e832969649838ba37f944ad11.

scruffian:
`is_allowed_media_file` has an `mpeg` branch, but `is_allowed_media_url` and the folder mapping only allow `mpg`. A template containing a `.mpeg` video would get rewritten to a local asset URL, then skipped before download, leaving the exported theme pointing at a missing file.

includes/create-theme/theme-zip.php
```diff
@@ -62,7 +67,15 @@ public static function add_activated_fonts_to_zip( $zip, $theme_json_string ) {
...
+						$tmp_file = download_url( $font_src );
+						if ( is_wp_error( $tmp_file ) ) {
+							continue;
+						}
+						// Post-download MIME allowlist.
+						if ( ! CBT_Theme_Fonts::is_allowed_font_file( $tmp_file, $font_src ) ) {
+							@unlink( $tmp_file );
+							continue;
+						}
 						$zip->addFileToTheme( $tmp_file, $font_face_path );
```
scruffian: I think we have the same problem as in `add_media_to_zip`, where `unlink` is being called before the sources are added to the zip, so we need the same fix here too I think.

mikachan:
> is_allowed_media_file has an mpeg branch, but is_allowed_media_url and the folder mapping only allow mpg.

Good spot, thank you. This should be fixed in https://github.com/WordPress/create-block-theme/pull/852/commits/375663ca48a95b9185f30529a3250d1c44660d2b.

includes/create-theme/theme-zip.php
mikachan: Thanks! Applied the same fix here in https://github.com/WordPress/create-block-theme/pull/852/commits/ef205f7e759b1852ea67a4b41fe2566d0341627a.

includes/create-theme/theme-media.php
```diff
@@ -19,6 +34,213 @@ public static function get_media_folder_path_from_url( $url ) {
...
+
+			case 'webp':
+				return strlen( $head ) >= 12
+					&& 'RIFF' === substr( $head, 0, 4 )
+					&& 'WEBP' === substr( $head, 8, 4 );
+
+			case 'avif':
+				// AVIF is ISO BMFF (like MP4) but with `avif` or `avis` as
+				// the major brand at offset 8.
+				return strlen( $head ) >= 12
```
scruffian: I think there other FileTypeBox brands - do we want to consider them as well?

mikachan: Yeah, I think covering more brands makes sense, thanks! Addressed in https://github.com/WordPress/create-block-theme/pull/852/commits/1e931c40118d9ad5ab961882a698386a24f8e894.

includes/create-theme/theme-zip.php
scruffian, line 75: In `copy_font_assets_to_theme` we check `is_allowed_font_file`. Should we do the same thing here so that non-font files aren't copied?

includes/create-theme/theme-media.php
```diff
@@ -117,43 +363,79 @@ public static function make_relative_media_url( $absolute_url ) {
...
+		}
+		$media_to_localize = array_unique( (array) $media_to_localize );
+
 		// Replace the absolute URLs with relative URLs in the templates
 		foreach ( $template->media as $media_url ) {
+			if ( ! in_array( $media_url, $media_to_localize, true ) ) {
+				continue;
+			}
 			$local_media_url   = CBT_Theme_Media::make_relative_media_url( $media_url );
 			$template->content = str_replace( $media_url, $local_media_url, $template->content );
```
scruffian: Because this operates on the whole template, a validated URL can still rewrite a different rejected URL when it is a prefix, for example photo.png also matching inside photo.png.php. In that case the rejected URL is localized even though it was not copied or validated, leaving the export pointing at a missing local asset and bypassing the new validated-media guard. Can we use the HTML tag processor  to replace only parsed URL attribute/block values that exactly match the validated URL?

includes/create-theme/theme-zip.php
mikachan: Yep I think so, addressed in https://github.com/WordPress/create-block-theme/pull/852/commits/ffea01d3d782fd5912899ba5d3cda0a7a1c6e832.

includes/create-theme/theme-media.php
mikachan: Yeah, makes sense, thank you. Fixed in https://github.com/WordPress/create-block-theme/pull/852/commits/86397ea95690b8055515fe6d77573b51646770ea. `make_template_images_local()` now walks the parsed block tree instead of running `str_replace` on the whole serialised template.

scruffian (approved):
🚢 
Thank you!

includes/create-theme/theme-fonts.php
```diff
@@ -112,6 +112,140 @@ public static function make_filename_from_fontface( $font_face, $src, $src_index
...
+				return 'wOF2' === $first_four;
+
+			case 'woff':
+				return 'wOFF' === $first_four;
+
+			case 'otf':
+				return 'OTTO' === $first_four;
+
+			case 'ttf':
+				return "\x00\x01\x00\x00" === $first_four || 'true' === $first_four;
```
dmsnell, line 234: the WHATWG spec does not recognize `true`, which means browsers also should not be recognizing it. as a “legacy” value there may be even more reason not to support it.

https://mimesniff.spec.whatwg.org/#matching-a-font-type-pattern

dmsnell (commented):
it’s worth looking into the `WP_Mime_Sniffer` class I have worked with in WordPress/wordpress-develop#10640. that implements the [WHATWG MIME Sniffing](https://mimesniff.spec.whatwg.org/) Living Standard, which should mirror what browsers are doing, and matching browser behavior is pretty important for being able to safely and security agree on what’s what.

PR #854, opened 2026-06-22
i18n: Handle backslashes in localized template text

MaggieCabrera:
## Summary

Updates the Localize Text export path so template text with backslashes and quotes is preserved correctly when it is wrapped for translation.

This covers localized text in normal block content, inline formatted content, and block-comment attributes. The change keeps the exported theme markup valid for strings that include a backslash before a quote, and adds regression coverage for those cases.

## Test plan

- [x] `npm run test:unit:php:base -- --filter CBT_Theme_Locale`
- [x] `vendor/bin/phpcs --standard=phpcs.xml.dist includes/create-theme/theme-locale.php tests/CbtThemeLocale/base.php tests/CbtThemeLocale/escapeTextContent.php tests/CbtThemeLocale/escapeAttribute.php tests/CbtThemeLocale/escapeBlockAttributes.php`

### DIFF
--- a/includes/create-theme/theme-locale.php
+++ b/includes/create-theme/theme-locale.php
@@ -8,6 +8,16 @@ require_once __DIR__ . '/theme-token-processor.php';
 class CBT_Theme_Locale {
 
 	/**
+	 * Escape a string that will be embedded in generated PHP single-quoted strings.
+	 *
+	 * @param string $string The string to escape.
+	 * @return string The escaped string.
+	 */
+	private static function escape_php_single_quoted_string( $string ) {
+		return addcslashes( (string) $string, "\\'" );
+	}
+
+	/**
 	 * Escape text for localization.
 	 *
 	 * @param string $string The string to escape.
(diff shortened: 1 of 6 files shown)


includes/create-theme/theme-locale.php
```diff
@@ -302,14 +321,15 @@ function ( $matches ) {
...
-						$attrs[ $attr_name ] = self::escape_attribute( $attrs[ $attr_name ] );
+						$attrs[ $attr_name ] = $next_placeholder( self::escape_attribute( $attrs[ $attr_name ] ) );
 						$modified            = true;
 					}
 				}
 
 				// If we modified any attributes, re-encode to JSON.
 				if ( $modified ) {
 					$new_attrs_json = wp_json_encode( $attrs, JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES );
+					$new_attrs_json = strtr( $new_attrs_json, $placeholders );
```
scruffian: This is an edge case, but this strstr replaces `__CBT_LOCALIZED_ATTRIBUTE_` across the content. Instead we could save the placeholders in an array and only replace them. I've added a commit to do this.

scruffian:
I commited a change to add a block-attribute-specific localization path. It tokenises unsafe characters into `chr(...)` placeholders, preserves translator context, and JSON-encodes the generated PHP string when rebuilding block attributes.

The previous escaping handled PHP single-quoted strings, but localised block attributes also have to survive JSON encoding in serialised block markup. Values containing backslashes, quotes, carriage returns, new lines, tabs, or `%`  could otherwise produce unsafe or invalid generated PHP/JSON.

I also added tests to cover the injection payload, double quotes, control characters, percent placeholders, and JSON decode validity.

MaggieCabrera (commented):
This is testing well for me but I can't approve, can you, @scruffian please?

scruffian (approved):
LGTM

PR #855, opened 2026-06-23
Keep SVG URLs remote during export

MaggieCabrera:
## Summary

Updates media export so SVG URLs found in templates and patterns are left at their original URL instead of being copied into the generated theme assets. Raster images and supported video files continue to be localized as before.

This aligns the plugin with WordPress core’s handling of SVGs: themes can still include their own SVG assets, but SVG files from content/media URLs are not treated like ordinary localizable image uploads by default.

## Test plan

- [x] `npm run test:unit:php:base -- --filter 'Test_Create_Block_Theme_Media|Test_Create_Block_Theme_Zip'`
- [x] `vendor/bin/phpcs --standard=phpcs.xml.dist includes/create-theme/theme-media.php tests/test-theme-media.php tests/test-theme-zip.php`

### DIFF
--- a/includes/create-theme/theme-media.php
+++ b/includes/create-theme/theme-media.php
@@ -108,7 +108,6 @@ class CBT_Theme_Media {
 			'jpeg',
 			'png',
 			'gif',
-			'svg',
 			'webp',
 			'avif',
 			// videos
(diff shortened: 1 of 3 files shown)


scruffian (approved):
LGTM.

One question - Is it worth detecting whether URL is a network accessible global domain rather than a local one?

MaggieCabrera:
> LGTM.
> 
> One question - Is it worth detecting whether URL is a network accessible global domain rather than a local one?

I don't think it's worth the complexity, there's a lot of dev environments (docker, playground, etc) we would have to consider.

